PR #9, opened 2025-12-11
fix: Add sysfs fallback for RDMA detection on InfiniBand interfaces

anson627:
The rdmamap library v1.1.0 has a bug where it compares InfiniBand
    interface hardware addresses against the node GUID instead of the
    port GUID. This causes InfiniBand (IPoIB) interfaces to incorrectly
    report dra.net/rdma=false even when they have RDMA capability.

    Related upstream issue: https://github.com/Mellanox/rdmamap/issues/15

    Root cause:
    - IPoIB interfaces use port GUID (derived from GID)
    - rdmamap library compares against node GUID
    - The two GUIDs differ in byte 5 (0xfd vs 0xfe)
    
    Changes:
    - Add hasRDMADeviceInSysfs() helper that checks sysfs directly
    - Update discoverRDMADevices() to fall back to sysfs check when
      rdmamap library returns false
    - This workaround detects RDMA for both Ethernet (RoCE/iWARP) and
      InfiniBand interfaces correctly
    
    The fix maintains backward compatibility by trying rdmamap first,
    then falling back to sysfs only when needed.
    
    Fixes: dra.net/rdma incorrectly showing false for InfiniBand devices

### DIFF
--- a/pkg/inventory/db.go
+++ b/pkg/inventory/db.go
@@ -398,7 +398,8 @@ func (db *DB) discoverRDMADevices(devices []resourceapi.Device) []resourceapi.De
 
 			// Fallback to sysfs check if rdmamap fails. This is particularly
 			// needed for InfiniBand interfaces where rdmamap has a bug comparing
-			// against node GUID instead of port GUID.
+			// against node GUID instead of port GUID:
+			// https://github.com/Mellanox/rdmamap/issues/15
 			if !isRDMA {
 				isRDMA = hasRDMADeviceInSysfs(*ifName)
 			}


gauravkghildiyal:
Thanks for contributing @anson627 !

Am I understanding this correctly that once the upstream fix is made in https://github.com/Mellanox/rdmamap/issues/15, we won't need a separate fix within DRANET? If yes, and assuming we can wait, we can try to get this upstreamed and then bump up the dependency with the fixed version

MikeZappa87:
@anson627 I dont see that rdma issue has moved. Is this PR still required?

anson627:
@MikeZappa87 I may need some help to get review from nvidia on https://github.com/Mellanox/rdmamap/pull/16

anson627:
> @gauravkghildiyal this seems stuck in a rock and a hardplace. @anson627 if this PR merges and then the PR you have in the other repo merges, nothing breaks correct?

yes, the current logic is fallback, I've tested with GB200 nodes on Azure

pkg/inventory/db.go
```diff
@@ -393,7 +393,15 @@ func (db *DB) discoverRDMADevices(devices []resourceapi.Device) []resourceapi.De
 	for i := range devices {
 		isRDMA := false
 		if ifName := devices[i].Attributes[apis.AttrInterfaceName].StringValue; ifName != nil && *ifName != "" {
+			// Try rdmamap library first
 			isRDMA = rdmamap.IsRDmaDeviceForNetdevice(*ifName)
+
+			// Fallback to sysfs check if rdmamap fails. This is particularly
+			// needed for InfiniBand interfaces where rdmamap has a bug comparing
+			// against node GUID instead of port GUID.
```
gauravkghildiyal: nit: Can you please add a link to the upstream issue here for easy future reference. Thanks!

anson627: Done

PR #18, opened 2025-12-15
add check to fix duplicate IPv6 routes

aman0408:
Fix google/dranet#274

Before adding a new route, we first check for an existing route. If route exists, skip adding a new route since it is likely a kernel route.

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -501,9 +501,16 @@ func getRouteInfo(nlHandle nlwrap.Handle, ifName string, link netlink.Link) ([]a
 			continue
 		}
 		// Discard IPv6 link-local routes, but allow IPv4 link-local.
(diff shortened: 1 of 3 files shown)


gauravkghildiyal (commented):
Thanks @aman0408!

I'm wondering that rather then checking for duplicates, can we skip copying routes where `proto=kernel` here? https://github.com/kubernetes-sigs/dranet/blob/a80be5ee904c43af534daa7fcba02324536efcc6/pkg/driver/dra_hooks.go#L475-L479 

As per https://man7.org/linux/man-pages/man8/ip-route.8.html:

_"kernel - the route was installed by the kernel during autoconfiguration."_

so the kernel is expected to create this route even within the network namespace as well and we should not manually copy it. Can we try verifying the same?

gauravkghildiyal:
Thanks!

/lgtm
/approve

Issue #19, opened 2026-01-05
DRA Driver fails to start after restart


tamilmani1989:
I have noticed DRA daemonset consistently fails after restarting it. Either restarting node or restarting kubelet helps to fix it but its not recovering by itself without restarts.

I0105 21:06:02.087493  877319 draplugin.go:597] "Starting"
I0105 21:06:02.087717  877319 nonblockinggrpcserver.go:90] "GRPC server started" logger="dra" endpoint="/var/lib/kubelet/plugins/dra.net/dra.sock"
I0105 21:06:02.088023  877319 nonblockinggrpcserver.go:90] "GRPC server started" logger="registrar" endpoint="/var/lib/kubelet/plugins_registry/dra.net-reg.sock"
F0105 21:06:32.088382  877319 app.go:169] driver failed to start: context deadline exceeded

PR #22, opened 2026-01-05
chore: Add rolling update support for DRA driver

tamilmani1989:
DRA plugins can be restarted with rolling update mode and this actually helped fixing dra driver restart issue https://github.com/kubernetes-sigs/dranet/issues/19. 

To support this, driver must pass pod UID to kubeletplugin.RollingUpdate. Rolling update is not enabled by default and to enable, user has to set `--pod-uid` as argument when invoking dra.

### DIFF
--- a/cmd/dranet/app.go
+++ b/cmd/dranet/app.go
@@ -57,6 +57,7 @@ var (
 	minPollInterval  time.Duration
 	maxPollInterval  time.Duration
 	pollBurst        int
+	podUID           string
 
 	ready atomic.Bool
 )
@@ -69,6 +70,7 @@ func init() {
 	flag.DurationVar(&minPollInterval, "inventory-min-poll-interval", 2*time.Second, "The minimum interval between two consecutive polls of the inventory.")
 	flag.DurationVar(&maxPollInterval, "inventory-max-poll-interval", 1*time.Minute, "The maximum interval between two consecutive polls of the inventory.")
 	flag.IntVar(&pollBurst, "inventory-poll-burst", 5, "The number of polls that can be run in a burst.")
+	flag.StringVar(&podUID, "pod-uid", "", "The UID of the pod running this plugin. This field is required for enabling rolling update support.")
 
 	flag.Usage = func() {
 		fmt.Fprint(os.Stderr, "Usage: dranet [options]\n\n")
@@ -164,6 +166,10 @@ func main() {
 		inventory.WithMaxPollInterval(maxPollInterval),
 	)
 	opts = append(opts, driver.WithInventory(db))
+	if podUID != "" {
+		opts = append(opts, driver.WithPodUID(podUID))
+		klog.Infof("Rolling updates enabled with pod UID: %s", podUID)
+	}
 	dranet, err := driver.Start(ctx, driverName, clientset, nodeName, opts...)
 	if err != nil {
 		klog.Fatalf("driver failed to start: %v", err)
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -31,6 +31,7 @@ import (
 	"github.com/google/dranet/internal/nlwrap"
 
 	resourceapi "k8s.io/api/resource/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/dynamic-resource-allocation/kubeletplugin"
@@ -80,6 +81,13 @@ func WithInventory(db inventoryDB) Option {
 	}
 }
 
+// WithPodUID sets the pod UID for rolling updates support.
+func WithPodUID(uid string) Option {
+	return func(o *NetworkDriver) {
+		o.podUID = uid
+	}
+}
+
 type NetworkDriver struct {
 	driverName string
 	nodeName   string
@@ -94,6 +102,7 @@ type NetworkDriver struct {
 	// Cache the rdma shared mode state
 	rdmaSharedMode bool
 	podConfigStore *PodConfigStore
+	podUID         string
 }
 
 type Option func(*NetworkDriver)
@@ -132,11 +141,17 @@ func Start(ctx context.Context, driverName string, kubeClient kubernetes.Interfa
 		kubeletplugin.NodeName(nodeName),
 		kubeletplugin.KubeClient(kubeClient),
 	}
+	if plugin.podUID != "" {
+		kubeletOpts = append(kubeletOpts, kubeletplugin.RollingUpdate(types.UID(plugin.podUID)))
+		klog.Infof("Enabling rolling update mode with pod UID: %s", plugin.podUID)
+	}
 	d, err := kubeletplugin.Start(ctx, plugin, kubeletOpts...)
 	if err != nil {
 		return nil, fmt.Errorf("start kubelet plugin: %w", err)
 	}
 	plugin.draPlugin = d
+
+	klog.Infof("Waiting for kubelet to register plugin %s", driverName)
 	err = wait.PollUntilContextTimeout(ctx, 1*time.Second, 30*time.Second, true, func(context.Context) (bool, error) {
 		status := plugin.draPlugin.RegistrationStatus()
 		if status == nil {
--- a/install.yaml
+++ b/install.yaml
@@ -130,6 +130,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: POD_UID
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.uid
         resources:
           requests:
             cpu: "100m"


pkg/driver/driver.go
```diff
@@ -127,16 +136,32 @@ func Start(ctx context.Context, driverName string, kubeClient kubernetes.Interfa
 		return nil, fmt.Errorf("failed to create plugin path %s: %v", driverPluginPath, err)
 	}
 
+	// Clean up stale sockets from previous instances to prevent registration failures
+	draSocketPath := filepath.Join(driverPluginPath, "dra.sock")
+	if err := os.Remove(draSocketPath); err != nil && !os.IsNotExist(err) {
+		klog.Warningf("failed to remove stale DRA socket %s: %v", draSocketPath, err)
+	}
+	regSocketPath := filepath.Join(kubeletPluginRegistryPath, driverName+"-reg.sock")
```
MikeZappa87: Not sure if we validate the driver name or plugin path, however is it possible to do a path traversal here and delete something unintentionally? Like ../../../etc/something?

tamilmani1989: actually this is not required. i noticed sock file getting cleaned up after dra driver restart. removing it for now and scoping this PR only to add rolling update support

cmd/dranet/app.go
```diff
@@ -69,6 +70,7 @@ func init() {
 	flag.DurationVar(&minPollInterval, "inventory-min-poll-interval", 2*time.Second, "The minimum interval between two consecutive polls of the inventory.")
 	flag.DurationVar(&maxPollInterval, "inventory-max-poll-interval", 1*time.Minute, "The maximum interval between two consecutive polls of the inventory.")
 	flag.IntVar(&pollBurst, "inventory-poll-burst", 5, "The number of polls that can be run in a burst.")
+	flag.StringVar(&podUID, "pod-uid", "", "The UID of the pod running this plugin. This field is required for enabling rolling update support.")
```
michaelasp, line 73:
```suggestion
	flag.StringVar(&podUID, "pod-uid", "", "The UID of the pod running this plugin. ")
```

Nit: This can go out of date quickly, for docs let's just put what the required args are, other uses and changes can happen which would make the second sentence obsolete.

cmd/dranet/app.go
```diff
@@ -164,6 +166,10 @@ func main() {
 		inventory.WithMaxPollInterval(maxPollInterval),
 	)
 	opts = append(opts, driver.WithInventory(db))
+	if podUID != "" {
```
michaelasp, line 169: Should we make this required? What's the advantage of not requiring the pod UID to be an env var here? I'd rather reduce the matrix of "modes" that DRANet can run in.

pkg/driver/driver.go
```diff
@@ -132,11 +141,17 @@ func Start(ctx context.Context, driverName string, kubeClient kubernetes.Interfa
 		kubeletplugin.NodeName(nodeName),
 		kubeletplugin.KubeClient(kubeClient),
 	}
+	if plugin.podUID != "" {
```
michaelasp, line 144: Again here I'd rather we just enable this always, do we have examples from other drivers that enable it optionally?

michaelasp (changes requested):

cmd/dranet/app.go
MikeZappa87: This probably makes sense. Isn't the current behavior broken if the pod is rescheduled? I also don't think anything is cleaning up the previous socket file and adding this field as required gives the kubelet a clue to what socket file to delete since it can detect if a pod uid is no longer running. 

https://github.com/kubernetes/kubernetes/blob/a1d638ea114f34497b0bf1f13f072dcb8c355bb0/staging/src/k8s.io/dynamic-resource-allocation/kubeletplugin/draplugin.go#L354

gauravkghildiyal (commented):
Thanks for contributing @tamilmani1989!

There's a potential edge case that I see with enabling rolling updates today which has to do with [this](https://github.com/kubernetes/kubernetes/blob/c36cb4913d480fadb046251b751936aa9bac27eb/staging/src/k8s.io/dynamic-resource-allocation/kubeletplugin/draplugin.go#L327-L331) comment:

> // RollingUpdate can be used to enable support for running two plugin instances
> // in parallel while a newer instance replaces the older 
// ...
> // There is no guarantee which of the two instances are used by kubelet.
> // For example, it can happen that a claim gets prepared by one instance
> // and then needs to be unprepared by the other. Kubelet then may fall back
> // to the first one again for some other operation. In practice this means
> // that each instance must be entirely stateless across method calls.

In particular, as things exist today, DRANET is not stateless between the `NodePrepareResources()` call made by the kubelet and the `RunPodSandbox()` call made by the container runtime (when DRANET is acting as a NRI plugin) use a shared in-process state. We expect both of these calls to be received by the SAME DRANET daemon process on the node. 

My suggestion would be that if rolling update is only needed to fix the deadlock from https://github.com/kubernetes-sigs/dranet/issues/19, then that issue has it's own separate root cause and we should not try to enable rolling update until it's absolutely needed. (At this point, I think it's acceptable for the ~seconds worth of downtime that may exist when the daemon pod is being replaced with a newer one)

If on the other hand, we do see immediate value with rolling updates, then we should try and verify how our "stateful" implementation behaves during this edge case, and if required try to think of  a fix.

MikeZappa87:
> Thanks for the contributing @tamilmani1989!
> 
> There's a potential edge case that I see with enabling rolling updates today which has to do with [this](https://github.com/kubernetes/kubernetes/blob/c36cb4913d480fadb046251b751936aa9bac27eb/staging/src/k8s.io/dynamic-resource-allocation/kubeletplugin/draplugin.go#L327-L331) comment:
> 
> > // RollingUpdate can be used to enable support for running two plugin instances
> > // in parallel while a newer instance replaces the older
> > // ...
> > // There is no guarantee which of the two instances are used by kubelet.
> > // For example, it can happen that a claim gets prepared by one instance
> > // and then needs to be unprepared by the other. Kubelet then may fall back
> > // to the first one again for some other operation. In practice this means
> > // that each instance must be entirely stateless across method calls.
> 
> In particular, as things exist today, DRANET is not stateless between the `NodePrepareResources()` call made by the kubelet and the `RunPodSandbox()` call made by the container runtime (when DRANET is acting as a NRI plugin). We expect both of these calls to be received by the SAME DRANET daemon process on the node.
> 
> My suggestion would be that if rolling update is only needed to fix the deadlock from #19, then that issue has it's own separate root cause and we should not try to enable rolling update until it's absolutely needed. (At this point, I think it's acceptable for the ~seconds worth of downtime that may exist when the daemon pod is being replaced with a newer one)
> 
> If on the other hand, we do see immediate value with rolling updates, then we should try and verify how our "stateful" implementation behaves during this edge case, and if required try to think of a fix.

We are in alignment with my comment as well. https://github.com/kubernetes-sigs/dranet/pull/22#discussion_r2676998606

I believe TM wanted to address the issue in #19 in a different PR

gauravkghildiyal:
>  One follow-up just for my understanding: I assume the state that dranet maintains is persistent. if so, where this state is stored? is that local file or something else?

Unfortunately, right now it's not persistent, and just an in-process state. 

Specifically, I'm talking about this, for instance https://github.com/kubernetes-sigs/dranet/blob/c2b4e5e4fb4a5710c353442a5919426270d47d07/pkg/driver/driver.go#L96 The state not being persistent does lead to its own gotchas. I'd like see us eventually make this persistent, but so far things have been manageable without.

MikeZappa87:
> > Unfortunately, right now it's not persistent, and just an in-process state.
> 
> ah got it but this would be a problem if DRANet get restarted in middle between nodeprepareresource and nri call, right?

The prepare resources RPC would fail and exit the SyncPod function and this would trigger kubelet to retry. 

https://github.com/kubernetes/kubernetes/blob/6cbb58349d147e99c11f8d6657abfc81fef7e269/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1531

tamilmani1989:
> > > Unfortunately, right now it's not persistent, and just an in-process state.
> > 
> > 
> > ah got it but this would be a problem if DRANet get restarted in middle between nodeprepareresource and nri call, right?
> 
> The prepare resources RPC would fail and exit the SyncPod function and this would trigger kubelet to retry. Since this fails the RunPodSandbox RPC is never sent from the kubelet to the container runtime (NRI would not be invoked in this case).
> 
> https://github.com/kubernetes/kubernetes/blob/6cbb58349d147e99c11f8d6657abfc81fef7e269/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1531

that's not my ask. What if prepareresourceclaim [link](https://github.com/kubernetes-sigs/dranet/blob/c2b4e5e4fb4a5710c353442a5919426270d47d07/pkg/driver/dra_hooks.go#L139) succeeds and dranet get restared before nri runpodsandbox call.. on restart it loses podConfigStore state right and nri call would fail?

MikeZappa87:
> > > > Unfortunately, right now it's not persistent, and just an in-process state.
> > > 
> > > 
> > > ah got it but this would be a problem if DRANet get restarted in middle between nodeprepareresource and nri call, right?
> > 
> > 
> > The prepare resources RPC would fail and exit the SyncPod function and this would trigger kubelet to retry. Since this fails the RunPodSandbox RPC is never sent from the kubelet to the container runtime (NRI would not be invoked in this case).
> > https://github.com/kubernetes/kubernetes/blob/6cbb58349d147e99c11f8d6657abfc81fef7e269/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1531
> 
> that's not my ask. What if prepareresourceclaim [link](https://github.com/kubernetes-sigs/dranet/blob/c2b4e5e4fb4a5710c353442a5919426270d47d07/pkg/driver/dra_hooks.go#L139) succeeds and dranet get restared before nri runpodsandbox call.. on restart it loses podConfigStore state right and nri call would fail?
> 
> if on restart, kubelet calls prepareresourceclaim all pods again, then it should be ok.. if not dra net will lose internal state

@gauravkghildiyal isnt the state populated by the kubelet part of the SyncPod function? (https://github.com/kubernetes/kubernetes/blob/1c894014ebe25e0b042efa91698284f527493d90/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1531)

tamilmani1989:
im closing this PR for now as discussed since dranet maintains a state in process and rolling update cannot be supported

PR #35, opened 2026-01-16
Dependency update improvements

stmcginnis:
Adds configuration to use dependabot to get updates for Go, Docker, and GitHub Action dependencies. Groups k8s.io updates since those should usually be taken as a set. Adds labels to make sure tests are run automatically.

Also pins third party actions to a specific SHA.


michaelasp:
/lgtm
/approve

Thanks for improving the dependency management!

Issue #58, opened 2026-01-28
Add support for cri-o


kanlkan:
The current DRANET installation manifest (install.yaml) assumes containerd as the container runtime and includes an init container in the DaemonSet.

https://github.com/kubernetes-sigs/dranet/blob/main/install.yaml#L96-L120

I’d like to add support for CRI-O by adding conditional logic to the init container command. I think it would be a good idea to detect the container runtime by reading status.nodeInfo.containerRuntimeVersion.

If you agree, I’m happy to submit a PR with the patch.

I also have one discussion point: the current init container implementation modifies the container runtime configuration to enable NRI. This may be unexpected for cluster administrators. In recent versions of containerd and CRI-O, NRI is enabled by default, so administrators may have intentionally disabled it. What do you think about this behavior?

kanlkan:
@gauravkghildiyal 

I've opened a PR for this issue. Could you please take a look?

PR #59, opened 2026-02-02
Add support for CRI-O

kanlkan:
#58 

I've added CRI-O support to install.yaml.
I've also added a note that DRANET runs with `privileged=true`.

### DIFF
--- a/install.yaml
+++ b/install.yaml
@@ -92,32 +92,6 @@ spec:
       - operator: Exists
         effect: NoSchedule
       serviceAccountName: dranet
-      hostPID: true
-      initContainers:
-      - name: enable-nri
-        image: busybox:stable
-        volumeMounts:
-        - mountPath: /etc
-          name: etc
-        securityContext:
-          privileged: true
-        command:
-        - /bin/sh
-        - -c
-        - |
-          set -o errexit
-          set -o pipefail
-          set -o nounset
-          set -x
-          if grep -q "io.containerd.nri.v1.nri" /etc/containerd/config.toml
-          then
-             echo "containerd config contains NRI reference already; taking no action"
-          else
-             echo "containerd config does not mention NRI, thus enabling it";
-             printf '%s\n' "[plugins.\"io.containerd.nri.v1.nri\"]" "  disable = false" "  disable_connections = false" "  plugin_config_path = \"/etc/nri/conf.d\"" "  plugin_path = \"/opt/nri/plugins\"" "  plugin_registration_timeout = \"5s\"" "  plugin_request_timeout = \"5s\"" "  socket_path = \"/var/run/nri/nri.sock\"" >> /etc/containerd/config.toml
-             echo "restarting containerd"
-             nsenter -t 1 -m -u -i -n -p -- systemctl restart containerd
-          fi
       containers:
       - name: dranet
         args:


install.yaml
```diff
@@ -109,15 +116,61 @@ spec:
...
-          then
-             echo "containerd config contains NRI reference already; taking no action"
-          else
-             echo "containerd config does not mention NRI, thus enabling it";
-             printf '%s\n' "[plugins.\"io.containerd.nri.v1.nri\"]" "  disable = false" "  disable_connections = false" "  plugin_config_path = \"/etc/nri/conf.d\"" "  plugin_path = \"/opt/nri/plugins\"" "  plugin_registration_timeout = \"5s\"" "  plugin_request_timeout = \"5s\"" "  socket_path = \"/var/run/nri/nri.sock\"" >> /etc/containerd/config.toml
-             echo "restarting containerd"
-             nsenter -t 1 -m -u -i -n -p -- systemctl restart containerd
-          fi
+
+          CONTAINER_RUNTIME_VERSION="$(kubectl get node "${NODE_NAME}" -o jsonpath='{.status.nodeInfo.containerRuntimeVersion}')"
```
michaelasp: Is this possible to get locally? Not a huge deal if it can't be.

install.yaml
```diff
@@ -109,15 +116,61 @@ spec:
...
+              then
+                echo "containerd config contains NRI reference already; taking no action"
+              else
+                echo "containerd config does not mention NRI, thus enabling it";
+                printf '%s\n' "[plugins.\"io.containerd.nri.v1.nri\"]" "  disable = false" "  disable_connections = false" "  plugin_config_path = \"/etc/nri/conf.d\"" "  plugin_path = \"/opt/nri/plugins\"" "  plugin_registration_timeout = \"5s\"" "  plugin_request_timeout = \"5s\"" "  socket_path = \"/var/run/nri/nri.sock\"" >> /etc/containerd/config.toml
+                echo "restarting containerd"
+                nsenter -t 1 -m -u -i -n -p -- systemctl restart containerd
+              fi
+              ;;
+            cri-o://*|crio://*)
```
michaelasp: Not sure how feasible it is, but this is entirely untested currently. Is there a way to run kind or something with CRIO? Otherwise if this ever fails in the future then we won't know.

aojea: see https://github.com/kubernetes-sigs/dranet/pull/59#issuecomment-3844080259

install.yaml
aojea: see https://github.com/kubernetes-sigs/dranet/pull/59#issuecomment-3844080259

MikeZappa87:
Could we simplify? Instead of modifying the container runtimes configuration, can we just detect if the NRI socket is present or not? If its not present, we just have the readiness check fail and output an error that it cannot reach the NRI socket? I know this is less than ideal, however it alleviates out of band updates to the configuration.

michaelasp:
> Could we simplify? Instead of modifying the container runtimes configuration, can we just detect if the NRI socket is present or not? If its not present, we just have the readiness check fail and output an error that it cannot reach the NRI socket? I know this is less than ideal, however it alleviates out of band updates to the configuration.

I guess the question in that case is whether we want DRANet to be in charge of enabling NRI if it isn't on already. If we delegate that to the administrator then we can go with this minimal approach and just check if the socket exists already.

aojea:
as I said before, this is legacy thing when I started, so just remove the init container then, no need to check for NRI enable, dranet will do that check and fail if it is not enable

install.yaml
```diff
@@ -95,12 +96,18 @@ spec:
...
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
         volumeMounts:
         - mountPath: /etc
           name: etc
         securityContext:
           privileged: true
+          runAsUser: 0
```
aojea: I think we can just remove the `hostPID` option and the entire initContainer based on the discussion

kanlkan: I’ve made the changes. I verified the behavior by assigning a 1Gbps Intel NIC with a static IP to a Pod while DRANET was deployed without `hostPID: true`.

Issue #42, opened 2026-01-16
DRAnet support symetric routing withing a Pod with multiple namespaces


tamilmani1989:
On deletion of pod, dra doesn't restore the routes and rules on host namespace which existed before moving the nic to pod network namespace. During pod creation, prepareresourceclaim will be invoked and dranet gets all routes and rules associated with that interface from host: https://github.com/kubernetes-sigs/dranet/blob/9b77771f1a7dd036ddd20c2aaaf3e35fbf580a33/pkg/driver/dra_hooks.go#L297 and then sets up `podCfg.NetworkInterfaceConfigInPod` object. When nri runpodsanbox invoked, it moves the nic to pod network namespace and programs those rules and routes in that namespace. On pod deletion, nic is moved back to host namespace but routes and rules are not restored back.

MikeZappa87:
https://github.com/kubernetes-sigs/dranet/blob/9b77771f1a7dd036ddd20c2aaaf3e35fbf580a33/pkg/driver/nri_hooks.go#L313

This is the code that moves the interface from the pod network namespace to the root network namespace. It doesn't apply routes or rules after its moved back to the root network namespace.

tamilmani1989:
> @tamilmani1989 and @MikeZappa87 -- I suppose that if you had some use cases in mind that cannot make do without this restoration, we can re-evaluate the current stance? Thoughts?

@gauravkghildiyal @aojea The requirement is that when secondary nic used as ingress, packet comes to nic but response leaves through other nic, this results in asymmetric routing which leads to packet loss.. if we have appropriate rules/routes configured inside pod namespace then we can make reply to go via same nic.. im fine with any alternate idea too as long this usecase is met.. 

one other option is to allow cloud provider to plumb their own custom rules or routes during prepareresourceclaim or during nri runpodsandbox call

tamilmani1989:
> but adding these routes/rules is what gives us symmetric routing, which is currently missing
>> Let me start over , this is the line that confuses me, I need to understand the use cases that requires symmetric routing and for what, how users will use this symmetric routing? in which scenarios?

[...]

PR #75, opened 2026-02-20
refactor code to be cloud provider agnostic

aojea:
we literally leaked GCE to the abstraction, before we move on decouple GCEism and make an interface that can be implemented by other cloud providers.

We need to use a key to identify the device in the cloud provider, so we create a common type that allow to identify the interface by mac, pciaddress or name

### DIFF
--- a/pkg/cloudprovider/cloud.go
+++ b/pkg/cloudprovider/cloud.go
@@ -16,7 +16,10 @@ limitations under the License.
 
 package cloudprovider
 
-import resourceapi "k8s.io/api/resource/v1"
+import (
+	"github.com/google/dranet/pkg/apis"
+	resourceapi "k8s.io/api/resource/v1"
+)
 
 // DeviceIdentifiers contains locally discovered hardware identifiers
 // that a cloud provider can use to match against its metadata.
@@ -31,4 +34,8 @@ type CloudInstance interface {
 	// GetDeviceAttributes takes multiple identifiers, allowing the provider
 	// to choose the best way to match the local device to cloud metadata.
 	GetDeviceAttributes(id DeviceIdentifiers) map[resourceapi.QualifiedName]resourceapi.DeviceAttribute
+
+	// GetDeviceConfig allows a cloud provider to return an infrastructure-specific
+	// network configuration for a given device.
+	GetDeviceConfig(id DeviceIdentifiers) *apis.NetworkConfig
 }
--- a/pkg/inventory/cloud.go
+++ b/pkg/inventory/cloud.go
@@ -49,8 +49,13 @@ func getInstanceProperties(ctx context.Context) cloudprovider.CloudInstance {
 
 // getProviderAttributes retrieves cloud provider-specific attributes for a network interface
 func getProviderAttributes(device *resourceapi.Device, instance cloudprovider.CloudInstance) map[resourceapi.QualifiedName]resourceapi.DeviceAttribute {
-	if instance == nil || device == nil {
-		klog.Warningf("instance metadata or device is nil (instance: %v, device: %v), cannot get provider attributes.")
+	if instance == nil {
+		klog.Warningf("instance metadata is nil, cannot get provider attributes.")
+		return nil
+	}
+
+	if device == nil {
+		klog.Warningf("device is nil, cannot get provider attributes.")
 		return nil
 	}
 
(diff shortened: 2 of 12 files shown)


tamilmani1989:
> /assign @gauravkghildiyal @michaelasp
> 
> this will allow other cloud providers to implement the interface @tamilmani1989

nice. I kind of suggested same here: https://github.com/kubernetes-sigs/dranet/issues/42#issuecomment-3937481676

aojea:
> > /assign @gauravkghildiyal @michaelasp
> > this will allow other cloud providers to implement the interface @tamilmani1989
> 
> nice. I kind of suggested same here: [#42 (comment)](https://github.com/kubernetes-sigs/dranet/issues/42#issuecomment-3937481676)

then it looks this is going in the right direction, please validate this model works for you, I'm not familiar with other cloud providers

michaelasp:
Lgtm from a glance, we should definitely move the GCE implementation to its own go interface that just implements the generic cloudprovider one. Not too sure about what other cloud providers use, I'll review the implementation in the meantime but if @tamilmani1989 can validate if this model works well I am fine with this approach.

pkg/cloudprovider/cloud.go
```diff
@@ -16,22 +16,22 @@ limitations under the License.
...
-type NetworkInterface struct {
-	IPv4      string   `json:"ip,omitempty"`
-	IPv6      []string `json:"ipv6,omitempty"`
-	Mac       string   `json:"mac,omitempty"`
-	MTU       int      `json:"mtu,omitempty"`
-	Network   string   `json:"network,omitempty"`
-	IPAliases []string `json:"ipAliases,omitempty"`
+// CloudInstance defines the generic interface for all cloud providers.
+type CloudInstance interface {
+	GetProvider() CloudProvider
```
tamilmani1989: why do we need this? i didn't see its used anywhere

tamilmani1989 (commented):
mostly lgtm. left one comment

pkg/cloudprovider/cloud.go
aojea: maybe is too much future looking, but for cloud provider specific code this allow to identify at runtime which cloud provider is the instance belonging too, just in case we end with some specific logic per cloud provider ... or the absence of it ... I can drop it since as you say is not used anywhere

tamilmani1989: yes we can add it when required. Approving though

pkg/inventory/cloud_test.go
```diff
@@ -42,6 +42,10 @@ func (m *mockCloudInstance) GetDeviceAttributes(id cloudprovider.DeviceIdentifie
 	return nil
 }
 
+func (m *mockCloudInstance) GetDeviceConfig(id cloudprovider.DeviceIdentifiers) *apis.NetworkConfig {
+	return nil
```
michaelasp: Should we allow the mockCloudInstance to define this?

pkg/inventory/db.go
```diff
@@ -78,6 +78,10 @@ type DB struct {
 	// The deviceStore is periodically updated by the Run method.
 	deviceStore map[string]resourceapi.Device
```
gauravkghildiyal: nit: Please remove this new line. It usually serves as an indicator that all these fields are protected by the mutex `mu`.

pkg/inventory/cloud.go
```diff
@@ -48,18 +48,24 @@ func getInstanceProperties(ctx context.Context) *cloudprovider.CloudInstance {
 }
 
 // getProviderAttributes retrieves cloud provider-specific attributes for a network interface
-func getProviderAttributes(mac string, instance *cloudprovider.CloudInstance) map[resourceapi.QualifiedName]resourceapi.DeviceAttribute {
-	if instance == nil {
-		klog.Warningf("instance metadata is nil, cannot get provider attributes.")
+func getProviderAttributes(device *resourceapi.Device, instance cloudprovider.CloudInstance) map[resourceapi.QualifiedName]resourceapi.DeviceAttribute {
+	if instance == nil || device == nil {
+		klog.Warningf("instance metadata or device is nil, cannot get provider attributes.")
```
gauravkghildiyal: nit: Can we log the pointers to better indicate what was nil.

```suggestion
		klog.Warningf("instance metadata or device is nil (instance: %v, device: %v), cannot get provider attributes.")
```

pkg/inventory/db.go
gauravkghildiyal: nit: Please remove this new line. That way, all variables not separated by a new line follow the [Mutex Hat convention](https://dmitri.shuralyov.com/idiomatic-go#mutex-hat) that all of them are protected by the mutex `mu`.

aojea:
```suggestion
```

Issue #63, opened 2026-02-06
Support RDMA NIC sharing


gauravkghildiyal:
We want to support sharing a single RDMA NIC across multiple pods

Context: (Based on a discussion on [slack](https://kubernetes.slack.com/archives/C09QYUH5W/p1770294876430649)) There are use cases within LLM inference which require scheduling multiple pods on a single Node which consume accelerator (GPU/TPU). The Node itself may only have one RDMA device, which means for multi-node inference, the RDMA device would need to be shared between all pods on the single node to allow RDMA traffic to other nodes.

E.g. 

```
CPU / PCIe Root Complex 0
 ├── GPU0
 ├── GPU1
 ├── GPU2
 ├── GPU3
 └── NIC0
```

Possible solution:
1. Create IPVLANs for each pod requesting an RDMA NIC to share the underlying RDMA device.
2. Use the Consumable Capacity ([kep.k8s.io/5075](kep.k8s.io/5075)) in DRA to model this resource sharing capability.

MikeZappa87:
> [@MikeZappa87](https://github.com/MikeZappa87) You are correct to say that setting the rdma system to shared would atleast share the rdma/infiniband devices. But, we've seen that user applications and libraries (think NCCL or XLA as an example) making use of RoCE usually also expect the presence of a linux network interfaces (that is mapped to the rdma devices). So IPVLAN devices would be able to fulfill the job of interfaces available within each pod which all map to the same physical RDMA physical device.

That part I understand. The original github issue had no context.

gauravkghildiyal:
@ngcxy is working on this.

ngcxy:
Quick update on the progress:

Following our team discussion, here is an overview of the API changes we want to introduce:

* We are introducing a new allocation mode that allows a workload to claim only the child interface, making the parent NIC shareable across multiple child interfaces.
* This new **"shared allocation"** and the original **"exclusive allocation"** mode will be modeled as two different device definitions for a single NIC inside the `ResourceSlice`.
* To achieve this pattern, we will borrow concepts from two v1.36 beta features - specifically, `allowMultipleAllocations: bool` and `capacity` fields from [Consumable Capacity](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#consumable-capacity) and `counters` field from [Partitionable Devices](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#partitionable-devices).

[...]

PR #77, opened 2026-02-28
  Add IB-only RDMA device support and AKS GB300 example                                                                                                                   

anson627:
# Summary                                                                                                                                                                 

This pull request introduces a comprehensive end-to-end demo for topologically-aware GPU and InfiniBand NIC allocation using Dynamic Resource Allocation (DRA) on Azure Kubernetes Service (AKS) with ND GB300-v6 nodes. It adds example manifests, live resource slice references, and validation logic for RDMA-only devices, enabling precise NUMA-aware resource allocation and improved workload isolation for high-performance distributed ML jobs.

**Demo and Documentation:**
* Added a detailed `README.md` explaining the AKS GB300 InfiniBand DRA demo, NUMA topology, device mapping, DRA device attributes, usage instructions, and benchmark results, including the impact of NUMA alignment on performance.

**Validation and API Enhancements:**
* Added a new validation function `ValidateRDMAOnlyConfig` to ensure that network configuration fields are not specified for RDMA-only (InfiniBand-only, no netdev) devices, improving correctness and error reporting.
* Added a new device attribute constant `AttrRDMADevice` to the API for tracking the RDMA device name.

# Test plan

  - Deploy on AKS ND GB300 v6 with ghcr.io/anson627/dranet:test
  - Verify ResourceSlice has ibOnly: true and rdmaDevice: mlx5_X for all 4 IB NICs
  - Run 1nic-aligned: confirm only uverbs0 visible, ~56 GB/s busbw
  - Run 2nic-aligned: confirm only uverbs0+uverbs1 visible, ~112 GB/s busbw
  - Run 1nic-unaligned: confirm only uverbs2 visible, ~25 GB/s busbw (GDR disabled, cross-NUMA)

### DIFF
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -171,19 +171,22 @@ func (np *NetworkDriver) runPodSandbox(_ context.Context, pod *api.PodSandbox, p
 			if err := attachRdmaToNS(config.RDMADevice.LinkDev, ns, resourceClaimStatusDevice); err != nil {
 				return err
 			}
-			// For IB-only devices (no netdev), emit the top-level Ready condition.
-			// RunPodSandbox runs once per pod, so this fires exactly once per claim.
-			if ifName == "" {
-				resourceClaimStatusDevice.WithConditions(
-					metav1apply.Condition().
-						WithType("Ready").
-						WithReason("RDMAOnlyDeviceReady").
-						WithStatus(metav1.ConditionTrue).
-						WithLastTransitionTime(metav1.Now()),
-				)
-			}
 		}
-		
+
+		// Block 3: Status conditions for IB-only devices (no netdev).
+		// In exclusive RDMA mode the RDMA link was moved above; in shared mode
+		// char-device injection (createContainer) is sufficient. Either way the
+		// device is ready, so emit the condition unconditionally.
+		if ifName == "" && config.RDMADevice.LinkDev != "" {
+			resourceClaimStatusDevice.WithConditions(
+				metav1apply.Condition().
+					WithType("Ready").
+					WithReason("RDMAOnlyDeviceReady").
+					WithStatus(metav1.ConditionTrue).
+					WithLastTransitionTime(metav1.Now()),
+			)
+		}
+
 		resourceClaimStatus.WithDevices(resourceClaimStatusDevice)
 	}
 	// do not block the handler to update the status


pkg/driver/pod_device_config.go
```diff
@@ -27,6 +27,11 @@ import (
 // network device allocated to a Pod. This includes network interface settings,
 // routes for the Pod's network namespace, and RDMA configurations.
 type PodConfig struct {
+	// IBOnly is true for InfiniBand-only devices that have no netdev interface.
+	// When set, the netdev attach/detach steps are skipped and only RDMA
+	// character devices are injected into the container.
+	IBOnly bool
```
aojea: Since we started the API from network device -> rdma device, now we have this oddness at the API level for handling this devices that does not have network interface, this is my fault, sorry about it ... 

I'd like us to think if we can avoid exposing a new boolean. Based on a quick look at the code I think that we have enough data to know when is an `IBOnly device`, we have the link type and we know that `if rdmaDevice != "" && ifName == ""` it will be `IBOnly` ... do we need to expose this boolean?

pkg/driver/nri_hooks.go
```diff
@@ -153,6 +153,21 @@ func (np *NetworkDriver) runPodSandbox(_ context.Context, pod *api.PodSandbox, p
...
 
+		// IB-only path: no netdev to move into the pod namespace.
+		// RDMA char devices are injected via CreateContainer.
+		if config.IBOnly {
+			klog.V(2).Infof("RunPodSandbox IB-only device %s: skipping netdev attach", deviceName)
+			resourceClaimStatusDevice.WithConditions(
+				metav1apply.Condition().
+					WithType("Ready").
+					WithReason("IBOnlyDeviceReady").
+					WithStatus(metav1.ConditionTrue).
```
aojea: can we really update the status here with this condition? at this point we didn't move the device yet so it seems better to update the condition on the `CreateContainer` ... although on that hook, we should make it that only update the status once, if there are multipe containers it will try to update once per container, that is not what we want

pkg/driver/nri_hooks.go
```diff
@@ -310,6 +325,10 @@ func (np *NetworkDriver) stopPodSandbox(_ context.Context, pod *api.PodSandbox,
 		}
 	}
 	for deviceName, config := range podConfig {
+		// IB-only devices have no netdev in the namespace — nothing to detach.
+		if config.IBOnly {
```
aojea: what if `!np.rdmaSharedMode && config.RDMADevice.LinkDev != "" ` should we just avoid the `nsDetachNetdev`  `if config.NetworkInterfaceConfigInPod.Interface.Name == ""`

pkg/apis/attributes.go
```diff
@@ -45,4 +45,6 @@ const (
 	AttrSRIOVVfs        = AttrPrefix + "/" + "sriovVfs"
 	AttrVirtual         = AttrPrefix + "/" + "virtual"
 	AttrRDMA            = AttrPrefix + "/" + "rdma"
+	AttrIBOnly          = AttrPrefix + "/" + "ibOnly"
```
aojea: I think we can omit this boolean if we are able to uniquivocally identify this devices , also, we need to add new validation on validation.go to avoid configurations of networking for these rdma only devices

pkg/driver/nri_hooks.go
```diff
@@ -153,6 +153,21 @@ func (np *NetworkDriver) runPodSandbox(_ context.Context, pod *api.PodSandbox, p
...
+					WithType("Ready").
+					WithReason("IBOnlyDeviceReady").
+					WithStatus(metav1.ConditionTrue).
+					WithLastTransitionTime(metav1.Now()),
+			)
+			resourceClaimStatus.WithDevices(resourceClaimStatusDevice)
+			continue
+		}
+
 		ifName := config.NetworkInterfaceConfigInHost.Interface.Name
```
aojea, line 158: we need to reorganize this code block as it is doing network and rdma device operations, we can skip network operations if there are no network interface name
```suggestion
		ifName := config.NetworkInterfaceConfigInHost.Interface.Name
		if ifName == ""  {
		    continue
		 }   
```

but we also support moving rdma devices when not in shared mode, see line 246, so we somehow need to make this two operations block independent and use the `config.NetworkInterfaceConfigInHost.Interface.Name` and `config.RDMADevice.LinkDev` to discriminate

```
// Move the RDMA device to the namespace if the host is in exclusive mode
		if !np.rdmaSharedMode && config.RDMADevice.LinkDev != "" {
			klog.V(2).Infof("RunPodSandbox processing RDMA device: %s", config.RDMADevice.LinkDev)
			err := nsAttachRdmadev(config.RDMADevice.LinkDev, ns)
			if err != nil {
				klog.Infof("RunPodSandbox error getting RDMA device %s to namespace %s: %v", config.RDMADevice.LinkDev, ns, err)
				return fmt.Errorf("error moving RDMA device %s to namespace %s: %v", config.RDMADevice.LinkDev, ns, err)
			}
			resourceClaimStatusDevice.WithConditions(
				metav1apply.Condition().
					WithType("RDMALinkReady").
					WithStatus(metav1.ConditionTrue).
					WithReason("RDMALinkReady").
					WithLastTransitionTime(metav1.Now()),
			)
		}
```

pkg/driver/nri_hooks.go
aojea: same comment as in https://github.com/kubernetes-sigs/dranet/pull/77/changes#r2876566618 , since now we support these 3 combinations:
1. netdev only
2. rdma only
3. netdev + rdma combo

we should make the code able to handle the three of them

pkg/driver/pod_device_config.go
aojea, line 53:
```suggestion
// LinkDev is the name of the RDMA link device (e.g., "mlx5_0")
// Depending on the type of device (ROCE, WARP) it may have a network device associated
LinkDev string
```

pkg/driver/pod_device_config.go
aojea: I think that to avoid amibguity  if `RDMAConfig.LinkDev` is set

pkg/inventory/db.go
```diff
@@ -294,6 +294,16 @@ func (db *DB) discoverNetworkInterfaces(pciDevices []resourceapi.Device) []resou
...
 		}
 
+		// Skip IPoIB interfaces. The underlying PCI device will be discovered
+		// as an IB-only RDMA device (no netdev) via discoverRDMADevices.
+		// Associating the IPoIB netdev with the PCI device would mask the
+		// IB-only nature of the device and prevent correct RDMA char device
+		// injection into pods.
+		if link.Type() == "ipoib" {
+			klog.V(4).Infof("Network Interface %s is IPoIB, skipping netdev association (will be discovered as IB-only RDMA device)", ifName)
+			continue
```
aojea, line 308: how will be the network interface be configured in this case?

IIUIC you need to configure the `ib0` interface

examples/aks-gb300-infiniband-dranet/mpi-job.yaml
```diff
@@ -0,0 +1,61 @@
...
+            command: ["/bin/bash", "-c"]
+            args:
+            - |
+              sleep 5
+              mpirun -np 2 \
+                --allow-run-as-root \
+                -bind-to none \
+                -x LD_LIBRARY_PATH \
+                -x NCCL_DEBUG=INFO \
+                -x NCCL_SOCKET_IFNAME=eth0 \
```
aojea, line 24: I see, so you use the pod interface for the initial exchange and then you fallback to rdma ...

pkg/inventory/db.go
MikeZappa87: This IB only, the netdev ipoib is not required for this setup. The process interacts with the uverbs directly

aojea: yeah, but I'm trying to understand how do we support both cases ... what will be the UX for the people that will use the ipoib ? will this be opt-in or opt-out? ...

aojea: > This IB only, the netdev ipoib is not required for this setup. The process interacts with the uverbs directly

independently of is required or not, it looks to me that it should move to the pod too, what is the point of having it in the root namespacce? this setup depends on how do you configure the application, moving the interface to the pod namesapce makes it work with all setups, included this

gauravkghildiyal: > MikeZappa87: This IB only, the netdev ipoib is not required for this setup. The process interacts with the uverbs directly

I can understand it being "not required", but if it is present, is that a problem for IPoIB? In other words, @anson627 can we please share more details on what this means. (Sorry the lack of access to hardware that supports IPoIB means we may have to rely on you to share such nuances)

```
> // Associating the IPoIB netdev with the PCI device would mask the
> // IB-only nature of the device and prevent correct RDMA char device
> // injection into pods.
```

> aojea: what is the point of having it in the root namespacce?

So I'm familiar with use cases where you don't want to move the actual network device tied to the physical NIC into the pod, mainly due to some monitoring/NIC-programming application running on the root-namespace that depends on the the existence of the netdev in that namespace. (Such cases may be better handled generically through something like IPVLAN as discussed in https://github.com/kubernetes-sigs/dranet/issues/63) But is that the case here or is it something else is what I'd also like to understand better.

anson627: For this PR, we intentionally skip IPoIB interfaces (link type == ipoib) during inventory discovery. The reason is that associating the ib0 netdev with the PCI device would cause dranet to treat it as a regular Ethernet-backed device, masking the IB-only nature of the underlying RDMA device and breaking the uverbs* char-device injection that this PR is designed to deliver. The current target use case (confirmed with @MikeZappa87) is pure uverbs access — NCCL and similar MPI frameworks communicate directly over the verbs API without needing an IPoIB interface in the pod.

Supporting IPoIB-in-pod is a genuinely different design problem: moving a child ib0 netdev while its parent mlx5_ib device lives in the host namespace involves kernel IPoIB semantics we haven't fully validated yet, and we'd want to be intentional about the UX (opt-in attribute? separate device class? automatic when ib0 is present?). Bundling that into this PR would significantly expand scope and risk without benefit to the primary use case.

Would it be acceptable to merge the current IB-only char-device injection as-is, and track IPoIB-in-pod support as a follow-up issue? Happy to open that issue now with the context from this thread so nothing gets lost. We can revisit the kernel semantics and UX design there once hardware access is available to verify the behavior.

pkg/driver/pod_device_config.go
anson627: agree, this field is removed now

pkg/driver/nri_hooks.go
anson627: make sense, this is fixed and verified by checking resource claim status here: https://github.com/kubernetes-sigs/dranet/pull/77/changes#diff-5b6f8d713f3adf4a701793ce47df5b634a5d06d9d47428deb41de6fb99225160R41

pkg/driver/nri_hooks.go
anson627: agree, fixed

pkg/apis/attributes.go
anson627: make sense, validation added here: https://github.com/kubernetes-sigs/dranet/pull/77/changes#diff-3825ab34bc95276bbc8521313ccb629720c3b0610ddb9cc5f56d31928e8498baR257

examples/aks-gb300-infiniband-dranet/mpi-job.yaml
anson627: yes, this is most common use case on Azure, since most of GPU VMs are InfiniBand only:
https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/gpu-accelerated/ndh100v5-series?tabs=sizebasic
https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/gpu-accelerated/nd-gb200-v6-series?tabs=sizebasic

pkg/driver/pod_device_config.go
anson627: Comment updated

pkg/inventory/db.go
aojea: > Would it be acceptable to merge the current IB-only char-device injection as-is, and track IPoIB-in-pod support as a follow-up issue? Happy to open that issue now with the context from this thread so nothing gets lost. We can revisit the kernel semantics and UX design there once hardware access is available to verify the behavior.

I think we can implement some way of using a feature flag(enabled by default) to gate this behavior, because as per your great explanation there are some semantics and behaviors we need more experimentation and I can assert from personal experience that also the kernel and driver implementations behave quiet surprisingly

pkg/driver/nri_hooks.go
anson627: that's true, I did some refactoring to move netdev related logic into a separate function attachNetdevToNS and attachRdmaToNS to make this part easier to maintain

examples/aks-gb300-infiniband-dranet/mpi-job.yaml
tamilmani1989: @aojea `NCCL_SOCKET_IFNAME=eth0` is required only when the IB network interface is moved inside the pod. Otherwise, the initial metadata exchange fails because NCCL attempts to use the IB interface for that communication. If its treated as an IB‑only device and not move IB interface into the pod, explicitly setting NCCL_SOCKET_IFNAME is not necessary.

pkg/driver/nri_hooks.go
```diff
@@ -79,11 +79,15 @@ func (np *NetworkDriver) CreateContainer(ctx context.Context, pod *api.PodSandbo
 	return adjust, update, err
 }
 
-func (np *NetworkDriver) createContainer(_ context.Context, pod *api.PodSandbox, _ *api.Container, podConfig map[string]PodConfig) (*api.ContainerAdjustment, []*api.ContainerUpdate, error) {
-	// Containers only cares about the RDMA char devices
+func (np *NetworkDriver) createContainer(_ context.Context, _ *api.PodSandbox, _ *api.Container, podConfig map[string]PodConfig) (*api.ContainerAdjustment, []*api.ContainerUpdate, error) {
```
aojea, line 82: this is the tricky part, Pods can have multiple containers and I think it will be optimal to just update once, I do not think we can have a way to detect "this is the last Container in a Pod" ...I'm now starting to think I wrongly suggested to update the state here, that is more correct, but technically can cause a more problematic behavior ... I do not know if we should just leave it as in the original proposal in RunPodSandbox status and call it a day ... @gauravkghildiyal for some thoughts

examples/aks-gb300-infiniband-dranet/mpi-job.yaml
aojea: yeah, that is why I think we should move the ib interface by default to avoid drifting the behavior, and we may also have an option for users to decide to not move it

pkg/inventory/db.go
gauravkghildiyal: > I think we can implement some way of using a feature flag(enabled by default) to gate this behavior,

+1

Adding to that, I'm thinking that if we anyway go ahead with making such an experiment enabled by default, and because we've not validated a functioning IPoIB, maybe it makes more sense to introduce a feature gate when we eventually enable IPoIB and skip one right now?

pkg/driver/dra_hooks.go
```diff
@@ -222,6 +222,46 @@ func (np *NetworkDriver) prepareResourceClaim(ctx context.Context, claim *resour
...
+			}
+			if len(errorList) > 0 {
+				continue
+			}
+			rdmaDevName, err := np.netdb.GetRDMADeviceName(result.Device)
+			if err != nil {
+				errorList = append(errorList, fmt.Errorf("failed to get RDMA device name for IB-only device %s: %v", result.Device, err))
+				continue
+			}
+			podCfg.RDMADevice.LinkDev = rdmaDevName
```
gauravkghildiyal: Can we please refactor the common handling of this RDMA device into a shared function between IB-only and with-netdev code paths?

pkg/driver/nri_hooks.go
gauravkghildiyal: I agree, let's keep this in the RunPodSandbox for now. 

In addition to dealing with multiple containers, I think this also gets re-executed if the container is restarting (for any reason), so if we keep it here, we'll have to think how to deal with that as well.

I think what we ideally want is a reliable signal that the container was created with all these device injections successfully, and perhaps once we fix this integration point with NRI https://github.com/kubernetes-sigs/dranet/issues/88, the container successfully starting should itself serve as a signal that ALL required devices were injected successfully.

examples/aks-gb300-infiniband-dranet/resource-claim-template.yaml
```diff
@@ -0,0 +1,68 @@
...
+  spec:
+    devices:
+      requests:
+      - name: gpu
+        exactly:
+          deviceClassName: gpu.nvidia.com
+          count: 1
+          selectors:
+          - cel:
+              expression: 'device.attributes["resource.kubernetes.io"]["pciBusID"] == "0008:06:00.0"'
```
gauravkghildiyal: nit: Any reason to NOT use the more idomatic way `device.attributes["resource.kubernetes.io"].pciBusID` for these expressions?

(Non blocking, just curious.)

anson627: fixed

pkg/inventory/db.go
anson627: agree, I can follow-up on IPoIB later

pkg/driver/nri_hooks.go
anson627: make sense, RunPodSandbox runs once per pod, fixed to emit the top-level Ready condition, so this fires exactly once per claim

pkg/driver/dra_hooks.go
anson627: refactor with buildRDMAConfig function added

PR #87, opened 2026-03-05
fix: update module paths to sigs.k8s.io/dranet

rbtr:
Noticed this inconsistency between the current project home and the module path. It doesn't cause any visible issues with the build because imports are all evaluated locally and the go.mod declares the same module path as all of the imports, but if you try to `go get` it, for example:

```bash
$ go get sigs.k8s.io/dranet
go: downloading sigs.k8s.io/dranet v1.0.1
go: sigs.k8s.io/dranet@upgrade (v1.0.1) requires sigs.k8s.io/dranet@v1.0.1: parsing go.mod:
        module declares its path as: github.com/google/dranet
                but was required as: sigs.k8s.io/dranet
```

Which isn't resolvable since `github.com/google/dranet` still exists.

So ran a simple find/replace for `github.com/google/dranet -> sigs.k8s.io/dranet` then a `go fmt ./...` to reorder the new imports and made sure it still builds.

### DIFF
--- a/cmd/dranet/app.go
+++ b/cmd/dranet/app.go
@@ -31,11 +31,11 @@ import (
 
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/ext"
-	"github.com/google/dranet/pkg/driver"
-	"github.com/google/dranet/pkg/inventory"
-	"github.com/google/dranet/pkg/pcidb"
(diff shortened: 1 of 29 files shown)


aojea:
/lgtm
/approve

upps, thanks

Issue #88, opened 2026-03-05
[Reliability] Race condition during NRI plugin updates allows pods to bypass plugin logic


gauravkghildiyal:
Ran into this well-known race condition involving NRI plugins, https://github.com/containerd/nri/issues/142 where when DRANET pod is undergoing some lifecycle event (e.g. recreated due to an update) and around the same time a workload-pod gets created that expects device injection, the device injection can be skipped for the duration when DRANET pods are not-started.

We should follow up on that issue to determine if there's a way to avoid this race condition.

gauravkghildiyal:
(An issue closely related to this was [reported on Slack](https://kubernetes.slack.com/archives/C0B2UG578KA/p1787567041394769))

### The problem

* When RunPodSandbox() is invoked and the DRANET NRI plugin fails to complete within the required timeout, NRI treats it as unresponsive and closes the NRI <---> DRANET connection. (https://github.com/containerd/nri/blob/8ec8b285eae3ef5d2dbab8e679aad47f5ac58da9/pkg/adaptation/plugin.go#L625-L627)
* Before DRANET can re-register, NRI continues creating the Pod because DRANET is momentarily unregistered and nothing tells NRI to wait.

This is likely a more common variant of the issue being discussed at [Proposal: Ensuring NRI Plugin Execution for DRA Workloads via CRI Annotations](https://docs.google.com/document/d/1bssPQW4L_RVyVtGhq1rWidxFSVmvN2iwEHG0t5KN7S4/edit?usp=sharing) where we need a more closer integration between DRA and NRI.

### Possible short-term mitigations:

1. Ensure containerd config has "default validator" enabled:

```
[plugins."io.containerd.nri.v1.runtime"]
  [plugins."io.containerd.nri.v1.runtime".default_validator]
    enable = true
```

2. In your workload pods, use the annotation `required-plugins.noderesource.dev: '["dra.net"]'`. Adding this annotation to workload pods should force NRI to wait for the DRANET plugin execution and prevent NRI from skipping it.

qasmi:
Hello, I would like to help on this issue, I am facing a similar issue in production of pod getting started with part of the dranet devices allocated, complete silent failure form dranet side

PR #91, opened 2026-03-10
feat: implement robust graceful shutdown logic as a mitigation for NRI plugins not executing during restart

gauravkghildiyal:
This change implements a simple graceful shutdown in Stop() as a
mitigation for https://github.com/kubernetes-sigs/dranet/issues/88.

**NOTE: A more permanent solution involving making NRI plugin invocation
mandatory for DRA drivers will require a separate KEP and will be
worked on separately.**

Key changes:
- Refactor Stop() to follow the following sequence:
    - Halt new PrepareResourceClaims via DRA plugin stop.
    - Wait indefinitely for prepared pods to hit NRI hooks (CreateContainer)
        with a 10s grace period, deferring final timeout to the Kubernetes
        terminationGracePeriodSeconds.
- Add unit test for Stop().

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/dynamic-resource-allocation/resourceslice"
 	"k8s.io/klog/v2"
 	registerapi "k8s.io/kubelet/pkg/apis/pluginregistration/v1"
+	"k8s.io/utils/clock"
 )
 
 const (
@@ -97,6 +98,8 @@ type NetworkDriver struct {
 	// Cache the rdma shared mode state
 	rdmaSharedMode bool
 	podConfigStore *PodConfigStore
+
+	clock clock.WithTicker // Injectable clock for testing
 }
 
 type Option func(*NetworkDriver)
@@ -118,6 +121,7 @@ func Start(ctx context.Context, driverName string, kubeClient kubernetes.Interfa
 		kubeClient:     kubeClient,
 		rdmaSharedMode: rdmaNetnsMode == apis.RdmaNetnsModeShared,
 		podConfigStore: NewPodConfigStore(),
+		clock:          clock.RealClock{},
 	}
 
 	for _, o := range opts {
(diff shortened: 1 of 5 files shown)


pkg/driver/driver.go
```diff
@@ -95,6 +98,21 @@ type NetworkDriver struct {
...
+	// activity timestamp.
+	//
+	// An entry is added when a Pod's ResourceClaim is prepared, updated
+	// when a container is created for that Pod, and removed only when the
+	// ResourceClaim is unprepared (typically when the Pod is deleted).
+	//
+	// This map is used during Stop() to ensure that the driver doesn't shut
+	// down while pods are still being initialized in the container runtime.
+	allocatedPodsMu sync.Mutex
+	allocatedPods   map[types.UID]time.Time
```
aojea: can we reuse the `podConfigStore` so we only have one source of truth? these things tend to drift and having the source of truth for the pod state in two different places requires later to keep them in sync

pkg/driver/nri_hooks.go
```diff
@@ -70,6 +70,17 @@ func (np *NetworkDriver) CreateContainer(ctx context.Context, pod *api.PodSandbo
 	if !ok {
 		return nil, nil, nil
 	}
+
+	defer func() {
```
aojea, line 74: thinking out loud, we saw this on the update states of a previous PR, on Pods with multiple containers, we do not really know which one will use the rdma device , or just if this is the last container  ... @samuelkarp do you know if we can get this heuristic from NRI? to know exactly when all containers were created?

samuelkarp: CRI doesn't specify when "all" containers are created.  There's no indication to containerd about what container is an init container, what container is an ephemeral debug container, etc. With [this proposal](https://docs.google.com/document/d/1J8Aq0XzN8BiNdWHXSEGA1Xw2nXcZRSKTMoi-tNh7FTc/edit?disco=AAABTOCn_uo) there can even be dynamic container additions and removals to an existing pod in the future.

pkg/driver/driver.go
gauravkghildiyal: Sure thing. Introduced a separate `LastNRIActivity` within podConfigStore. 

I feel this has overall made the change simpler.

pkg/driver/nri_hooks.go
gauravkghildiyal: > thinking out loud, we saw this on the update states of a previous PR

Yeah right, it's the same requirement. I didn't find a more native solution for that right now.

pkg/driver/driver.go
```diff
@@ -209,9 +213,104 @@ func Start(ctx context.Context, driverName string, kubeClient kubernetes.Interfa
...
+	fallbackTimeout := 5 * time.Minute
+	ticker := np.clock.NewTicker(pollInterval)
+	defer ticker.Stop()
+
+	// defaultPreviousActivity is used as a baseline for pods that have been
+	// prepared but haven't recorded any NRI activity yet.
+	defaultPreviousActivity := np.clock.Now()
+
+	for {
+		done := func() bool {
```
aojea, line 257: I find this tendency of gemini to create closures a bit annoying, but is just a comment

pkg/driver/driver.go
```diff
@@ -209,9 +213,104 @@ func Start(ctx context.Context, driverName string, kubeClient kubernetes.Interfa
...
+
+	// Step 1: Halt the DRA plugin.
+	// This stops the driver from handling new NodePrepareResources requests,
+	// stabilizing the set of pods that require NRI processing.
 	np.draPlugin.Stop()
+
+	// Step 2: Wait for prepared pods to finish NRI initialization.
+	gracePeriod := 10 * time.Second
+	pollInterval := 5 * time.Second
+	fallbackTimeout := 5 * time.Minute
```
aojea, line 248: I like this way more, at least you know eventually it exits

PR #98, opened 2026-03-10
Add --move-ib-interfaces flag to control IPoIB interface handling

tamilmani1989:
Adds a [`--move-ib-interfaces`] flag (default: `true`) to make IPoIB network interface handling configurable.

When true, IPoIB interfaces are associated with their PCI devices and moved into pod namespaces like regular network interfaces. When false, IPoIB interfaces are skipped during discovery and the underlying device is exposed as an IB-only RDMA device.

### DIFF
--- a/cmd/dranet/app.go
+++ b/cmd/dranet/app.go
@@ -57,6 +57,7 @@ var (
 	minPollInterval  time.Duration
 	maxPollInterval  time.Duration
 	pollBurst        int
+	moveIBInterfaces bool
 
 	ready atomic.Bool
 )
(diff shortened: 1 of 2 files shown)


gauravkghildiyal:
@tamilmani1989 I think you'll have to update a few things based on the state of the merged PR https://github.com/kubernetes-sigs/dranet/pull/77

- As this PR introduces a flag for the behaviour, we don't need the special logic to ignore ifName from being published for IPoIB, right? https://github.com/kubernetes-sigs/dranet/blob/0807ba0b28a18c841d12d0b834f1132de0c6eaa8/pkg/inventory/db.go#L306-L309

- Within nri_hook, you can make use of `dra.net/type` in addition to the flag to check for which devices you don't want to move the interface to the pod's namespace.

Also, would `--move-infiniband-interfaces` with a default of `true` be a simpler acceptable name for you?

Issue #52, opened 2026-01-26
Request: Expose explicit SR-IOV PF/VF classification attributes in ResourceSlice


kanlkan:
# Environment (Worker node)

* OS: Ubuntu 24.04.3 LTS
* Kernel: 6.8.0-90-generic
* Kubernetes: v1.34.3
* containerd: v2.1.6
* DRANET: v1.0.1
* SR-IOV NIC: Mellanox ConnectX-6
  * Create 2 VFs (ens1f0v0, ens1f0v1)

# Details

`dra.net/sriov` and `dra.net/virtual` attributes for SR-IOV VFs in ResourceSlice are set to `false`.
Here is an example of Mellanox ConnectX-6 VFs in ResourceSlice.

```
...
  - attributes:
      dra.net/alias:
        string: ""
      dra.net/ebpf:
        bool: false
      dra.net/encapsulation:
        string: ether
      dra.net/ifName:
        string: ens1f0v0
      dra.net/mac:
        string: f6:f5:4b:8d:55:7c
      dra.net/mtu:
        int: 1500
      dra.net/numaNode:
        int: 0
      dra.net/pciAddress:
        string: "0000:27:00.2"
      dra.net/pciDevice:
        string: MT28908 Family [ConnectX-6 Virtual Function]
      dra.net/pciSubsystem:
        string: "0009"
      dra.net/pciVendor:
        string: Mellanox Technologies
      dra.net/rdma:
        bool: true
      dra.net/sriov:
        bool: false
      dra.net/state:
        string: down
      dra.net/type:
        string: device
      dra.net/virtual:
        bool: false
      resource.kubernetes.io/pcieRoot:
        string: pci0000:26
    name: pci-0000-27-00-2
  - attributes:
      dra.net/alias:
        string: ""
      dra.net/ebpf:
        bool: false
      dra.net/encapsulation:
        string: ether
      dra.net/ifName:
        string: ens1f0v1
      dra.net/mac:
        string: 32:f8:30:7f:45:51
      dra.net/mtu:
        int: 1500
      dra.net/numaNode:
        int: 0
      dra.net/pciAddress:
        string: "0000:27:00.3"
      dra.net/pciDevice:
        string: MT28908 Family [ConnectX-6 Virtual Function]
      dra.net/pciSubsystem:
        string: "0009"
      dra.net/pciVendor:
        string: Mellanox Technologies
      dra.net/rdma:
        bool: true
      dra.net/sriov:
        bool: false
[...]

PR #100, opened 2026-03-13
add isSriovVf attribute

kanlkan:
#52 

Add `isSriovVf` attribute for easier SR-IOV VF filtering.

### DIFF
--- a/pkg/inventory/db.go
+++ b/pkg/inventory/db.go
@@ -395,8 +395,11 @@ func addLinkAttributes(device *resourceapi.Device, link netlink.Link) {
 		vfs := int64(sriovNumVFs(ifName))
 		device.Attributes[apis.AttrSRIOVVfs] = resourceapi.DeviceAttribute{IntValue: &vfs}
 	}
+
 	isSriovVirtualFunction := isSriovVf(ifName, sysnetPath)
-	device.Attributes[apis.AttrIsSriovVf] = resourceapi.DeviceAttribute{BoolValue: &isSriovVirtualFunction}
+	if isSriovVirtualFunction {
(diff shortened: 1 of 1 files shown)


pkg/inventory/sysfs_test.go
```diff
@@ -136,3 +138,53 @@ func TestPCIAddressFromPath(t *testing.T) {
...
+			t.Fatalf("failed to create device directory for %q: %v", ifName, err)
+		}
+		return deviceDir
+	}
+
+	t.Run("missing physfn", func(t *testing.T) {
+		createDeviceDir(t, "eth0")
+		want := false
+		got := isSriovVf("eth0", syspath)
+		if diff := cmp.Diff(want, got); diff != "" {
```
aojea: `want` and `got` are booleans, no need to use `cmp.Diff`, let's make it simpler with `if want != go`

kanlkan: Thank you. I fixed it.

pkg/inventory/db.go
```diff
@@ -395,6 +395,8 @@ func addLinkAttributes(device *resourceapi.Device, link netlink.Link) {
 		vfs := int64(sriovNumVFs(ifName))
 		device.Attributes[apis.AttrSRIOVVfs] = resourceapi.DeviceAttribute{IntValue: &vfs}
 	}
+	isSriovVirtualFunction := isSriovVf(ifName, sysnetPath)
+	device.Attributes[apis.AttrIsSriovVf] = resourceapi.DeviceAttribute{BoolValue: &isSriovVirtualFunction}
```
aojea: there is a cap on the number of attributes, so I think is better to publish it only if is true , so we don't publish it for any other device

kanlkan: Thank you for suggstion. I fixed it.

PR #57, opened 2026-01-28
Fix PrepareResourceClaim to read the rdmadev from sysfs path as well

neaggarwMS:
his PR fixes the PrepareResourceClaim function to read the RDMA device name from sysfs path as a fallback when rdmamap fails to detect the device.

### Changes:
- Added \`getRdmaDeviceFromNetdev\` function in \`rdmadevice.go\` that:
  - First tries to get RDMA device using \`rdmamap.GetRdmaDeviceForNetdevice()\`
  - Falls back to reading from sysfs path \`/sys/class/net/<ifName>/device/infiniband/\` if rdmamap fails
  - This addresses a known issue with detecting RDMA devices for certain Mellanox NICs: https://github.com/Mellanox/rdmamap/issues/15

- Added comprehensive unit tests in \`rdmadevice_test.go\`

### Testing:
- Unit tests pass
- Tested on AKS cluster with Azure RDMA NIC (SKU: Standard_ND96isr_H100_v5)

PR #101, opened 2026-03-13
Fallback to get rdma device from sysfs

tamilmani1989:
if rdmamap call of GetRdmaDeviceForNetDevice failed, then dranet can fallback to get rdma device name via sysfs call.

This PR fixes similar to this PR https://github.com/kubernetes-sigs/dranet/pull/9 but in different part of code 

continuation of this PR: #57

### DIFF
--- a/pkg/inventory/sysfs.go
+++ b/pkg/inventory/sysfs.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/Mellanox/rdmamap"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/inventory/sysfs.go
```diff
@@ -100,7 +101,52 @@ func sriovNumVFs(name string) int {
...
+	entries, err := os.ReadDir(rdmaDir)
+	if err != nil {
+		return "", fmt.Errorf("no RDMA device for %s: %w", ifName, err)
+	}
+
+	for _, entry := range entries {
+		if entry.IsDir() {
+			klog.V(4).Infof("Found RDMA device %s for interface %s via sysfs", entry.Name(), ifName)
+			return entry.Name(), nil // Return first RDMA device found (e.g., "mlx5_0")
+		}
```
aojea, line 144: are os.ReadDir entries order always consistent?

aojea: it seems so 

> ReadDir reads the contents of the directory associated with the file f and returns a slice of [DirEntry](https://pkg.go.dev/os#DirEntry) values in directory order. Subsequent calls on the same file will yield later DirEntry records in the directory.

PR #102, opened 2026-03-13
Add Azure placement group support

anson627:
This pull request adds [Azure placement group](https://learn.microsoft.com/en-us/azure/virtual-machines/setup-infiniband#cluster-configuration-options) awareness to the project, enabling the system to expose Azure VMSS placement group and VM size information as device attributes. This allows workloads to make scheduling decisions that respect InfiniBand fabric boundaries, preventing cross-placement-group RDMA failures in AKS clusters. 

* Added a new `pkg/cloudprovider/azure/azure.go` module that queries the Azure Instance Metadata Service (IMDS) to retrieve the VM's `placementGroupId` and `vmSize`, and exposes them as device attributes for all devices on the node.

* Added `examples/aks-gb300-placement-group/README.md` with a detailed explanation of the Azure placement group problem, the solution using device attributes, and an example of how to use CEL selectors in `ResourceClaimTemplate` to ensure scheduling within the same IB fabric.

* Added unit tests in `pkg/cloudprovider/azure/azure_test.go` to verify correct population of device attributes and behavior of the Azure provider implementation.

### DIFF
--- a/pkg/cloudprovider/azure/OWNERS
+++ b/pkg/cloudprovider/azure/OWNERS
@@ -1,4 +1,0 @@
-# See the OWNERS docs at https://go.k8s.io/owners
-
-reviewers:
-  - anson627


pkg/cloudprovider/azure/azure.go
```diff
@@ -0,0 +1,161 @@
+/*
```
aojea, line 1: I'm fine adding an OWNERS file inside this folder for azure folks to iterate quicker, although I prefer if you don;t mind to start as reviwers first so we have some sort of organized way of promoting approvers. My experience is that this quickly into a lot of approvers and we loose the best practices when independent teams trying to get their features without thinking in the overall impact

pkg/cloudprovider/azure/azure.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+const (
+	AzureAttrPrefix = "azure.dra.net"
+
+	AttrAzurePlacementGroupID = AzureAttrPrefix + "/" + "placementGroupId"
+	AttrAzureVMSize           = AzureAttrPrefix + "/" + "vmSize"
+
+	// imdsEndpoint is the Azure Instance Metadata Service endpoint.
+	imdsEndpoint = "http://169.254.169.254/metadata/instance"
```
aojea, line 42: this will bit you later when you need to support ipv6, add a todo or thinkg if you can use a fallback using some well known name

pkg/cloudprovider/azure/azure.go
```diff
@@ -0,0 +1,161 @@
...
+// device-specific network configuration via IMDS.
+func (a *AzureInstance) GetDeviceConfig(id cloudprovider.DeviceIdentifiers) *apis.NetworkConfig {
+	return nil
+}
+
+// OnAzure returns true if the code is running on an Azure VM by probing the
+// IMDS endpoint. This is a lightweight check that does not parse the full
+// response.
+func OnAzure() bool {
+	client := &http.Client{Timeout: 2 * time.Second}
```
aojea: this things use to flake on some corner cases but the result is really frustrating for the user and it needs to restart the dranet pod, better use a context and an active poll like you have below

pkg/cloudprovider/azure/azure.go
anson627: I'm afraid not, the [Azure Instance Metadata Service (IMDS)](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service) must be called using the hardcoded, well-known non-routable IPv4 address 169.254.169.254. It is not designed to be accessed via a DNS name, as it is a RESTful API reachable directly through the host network interface within the VM

pkg/cloudprovider/azure/azure.go
anson627: Done

pkg/cloudprovider/azure/azure.go
anson627: Good point, Done!

gauravkghildiyal:
> The OWNERS file contains untrusted users, which makes it INVALID. The following users are mentioned in OWNERS file(s) but are untrusted for the following reasons. One way to make the user trusted is to add them as [members](https://git.k8s.io/community/community-membership.md#member) of the kubernetes-sigs org. You can then trigger verification by writing `/verify-owners` in a comment.
> 
> * anson627
>   
>   * User is not a member of the org. Satisfy at least one of these conditions to make the user trusted.

@anson627 FYI -- You'll have to get kubernetes-sigs membership before being eligible for OWNERS https://github.com/kubernetes/community/blob/master/community-membership.md#requirements. Seeing that this PR is almost ready, perhaps we merge this without OWNERS and add them later? Or you can include some other user(s) in OWNERS for now. Whatever you decide, you'll need to update the PR with the relevant change.

PR #112, opened 2026-03-24
Changes to provide build and distro images to Dockerfile as arguments

nakshah87:
Made the Dockerfile's base images configurable by introducing GOLANG_IMAGE and BASE_IMAGE build arguments, allowing different cloud providers to specify custom Go compiler and runtime images instead of hardcoded versions. 

This enables flexibility for different build environments, security requirements, or when using alternative image registries while maintaining golang:1.26 and gcr.io/distroless/base-debian12 as defaults.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,10 @@
 # limitations under the License.
 
 # setup cross-compile env
-FROM --platform=$BUILDPLATFORM golang:1.26 AS builder
+ARG GOLANG_IMAGE=golang:1.26
+ARG BASE_IMAGE=gcr.io/distroless/base-debian12
+
+FROM --platform=$BUILDPLATFORM $GOLANG_IMAGE AS builder
 ARG TARGETARCH
 ARG GOARCH=${TARGETARCH} CGO_ENABLED=0
 
@@ -27,6 +30,6 @@ COPY . .
 RUN go build -o /go/bin/dranet ./cmd/dranet
 
 # copy binary onto base image
-FROM gcr.io/distroless/base-debian12
+FROM $BASE_IMAGE
 COPY --from=builder --chown=root:root /go/bin/dranet /dranet
 CMD ["/dranet"]
--- a/Makefile
+++ b/Makefile
@@ -57,16 +57,27 @@ TAG?=$(shell echo "$$(date +v%Y%m%d)-$$(git describe --always --dirty)")
 IMAGE?=$(REGISTRY)/$(IMAGE_NAME):$(TAG)
 PLATFORMS?=linux/amd64,linux/arm64
 
+# base images (defaults are in the Dockerfile)
+BUILD_ARGS?=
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -13,7 +13,10 @@
 # limitations under the License.
 
 # setup cross-compile env
-FROM --platform=$BUILDPLATFORM golang:1.26 AS builder
+ARG GOLANG_IMAGE=golang:1.26
```
michaelasp, line 16: Curious, why would this need to be changed? Don't mind adding an argument but wondering when we would need a different golang image since we would bump this in tandem with a golang bump.

nakshah87: I was thinking we could have the flexibility to provide the golang version of choice. It anyways needs to be >= the go version in go.mod and in case if anyone wants to compile with the latest go version until it is bumped up here then that can be done.

Let me know if that works else this one is not really a blocker.

michaelasp: I think this makes sense, my only concern would be people not building at the right go version.

nakshah87:
> > @nakshah87 can you sign the CLA?
> 
> Yeah I am working on it with the legal team to get the approval.

@michaelasp Done.

Makefile
```diff
@@ -57,16 +57,24 @@ TAG?=$(shell echo "$$(date +v%Y%m%d)-$$(git describe --always --dirty)")
 IMAGE?=$(REGISTRY)/$(IMAGE_NAME):$(TAG)
 PLATFORMS?=linux/amd64,linux/arm64
 
+# base images
+GOLANG_IMAGE?=golang:1.26
```
michaelasp: Can we not pass the args if they don't exist? Otherwise we have 2 separate declarations of the base and golang images, making it a bit more tedious when we bump. Since it's the default, let's just omit the args if they are empty.

fmuyassarov:
> > > I don’t really see an issue here, and I’m not sure why there would be a need to parameterize the Dockerfile. Typically, a Dockerfile isn’t something we expect to adjust or tune before deployment. If this were related to something like Helm configuration or other runtime settings, then it would make more sense to parameterize. But for a Dockerfile, I’m not entirely convinced. It seems simpler to update the necessary fields directly and rebuild the image when needed.
> > 
> > 
> > Hi fmuyassarov, Having the base image parameterized gives every cloud provider the flexibility to provide their internally recommended image. Having them parameterized makes it easier to replace/update instead of updating it directly.
> 
> +1, having this just adds flexibility. I am in favor of merging this. There are many times where someone may want to override the base images due to compliance/ airgapped builds, etc...

As said I'm not against this, I just didn't realize/see the actual use case initially as I'm not the one using the driver on the cloud provider.

Makefile
nakshah87: Yeah. Good point. Completely makes sense to me. I will update the diff.

PR #115, opened 2026-03-27
feat: bbolt persistent pod device configs

### LINKED ISSUE #89: [Reliability] In-memory state loss during DRANET restarts leads to pod configuration bypass
The stateful nature of the DRANET driver during the pod creation phase is kinda problematic. 

The driver maintains an in-memory [podConfigStore](https://github.com/kubernetes-sigs/dranet/blob/0f136aa9d8c3da4d040edf297719eb722ea9cc1d/pkg/driver/driver.go#L97) to track pod-specific network and device configurations. This state is populated during the DRA [NodePrepareResource](https://github.com/kubernetes-sigs/dranet/blob/0f136aa9d8c3da4d040edf297719eb722ea9cc1d/pkg/driver/dra_hooks.go#L374) phase and is subsequently used by NRI hooks ([RunPodSandbox](https://github.com/kubernetes-sigs/dranet/blob/0f136aa9d8c3da4d040edf297719eb722ea9cc1d/pkg/driver/nri_hooks.go#L116), [CreateContainer](https://github.com/kubernetes-sigs/dranet/blob/0f136aa9d8c3da4d040edf297719eb722ea9cc1d/pkg/driver/nri_hooks.go#L69)) to apply configurations.

If the DRANET pod restarts (e.g., due to an update or crash) after resources have been prepared but before the NRI hooks are executed, the driver loses the context for the pod. This results in the NRI hooks silently skipping device injection or network setup, allowing pods to bypass plugin logic entirely.

rbtr:
#### What type of PR is this?

/kind feature

Supplements the in-memory-only store with a bbolt-backed persistent store using a layered write-through cache architecture. The in-memory PodConfigStore remains the source of truth for all reads, with an optional Checkpointer implementation as a write-through persistent backend. This follows the kubelet DRA checkpoint precedent (pkg/kubelet/cm/dra/state).

 - Defines a Checkpointer interface (GetOrCreate, Store, DeletePod, Close) as the persistence contract. Adds a boltCheckpointer implementation.
 - Modifies PodConfigStore to accept an optional Checkpointer. Writes persist to the checkpointer before updating memory; if persistence fails, memory is unchanged. Deletes proceed with memory removal regardless — stale checkpoint entries are harmless and cleaned up by Synchronize on the next startup.
 - Reads (GetPodConfig, GetDeviceConfig) are served from memory in the NRI hot path.
 - Ephemeral state (LastNRIActivity) stays entirely in-memory in the PodConfigStore and never reaches the persistence layer.
 - Wires persistence in to lifecycle: WithDBPath option configures the bolt store in Start(), and Close() shuts down the database.
 - Adds config for db path: `--db-path`. Set to empty to disable persistence. Sets default path as /var/run/dranet/dranet.db and adds a corresponding hostPath volume (DirectoryOrCreate) to deployment samples. 

Fixes #89

The `--db-path` flag defaults to `/var/run/dranet/dranet.db`. Set it to empty to disable persistence and fall back to pure in-memory behavior.

```release-note
Adds optional bbolt-backed persistence for pod device configs. Configs now survive daemon restarts so pods that were already prepared do not need to be re-prepared by the kubelet. Configure with `--db-path` (default: `/var/run/dranet/dranet.db`); set to empty to disable.

### CODE REVIEW

File: pkg/driver/pod_device_config_bolt.go [hunk 1]
```diff
@@ -0,0 +1,318 @@
...
+	return s.db.Update(func(tx *bolt.Tx) error {
+		root := tx.Bucket(podConfigsBucket)
+		if root == nil {
+			return berrors.ErrBucketNotFound
+		}
+		podBucket, err := root.CreateBucketIfNotExists([]byte(podUID))
+		if err != nil {
+			return err
+		}
+		data, err := json.Marshal(config)
```
gauravkghildiyal: I'll recommend us to have another bucket (or a key-value pair if that serves better) to group the deviceConfigs, rather than storing it directly within the podUID bucket. 

So instead of `pod_configs => <POD_UID> => <ListOfDevices>`, we instead of `pod_configs => <POD_UID> => device_configs => <ListOfDevices>`

Rationale being that it's very likely we'll need a Pod level config as well which is not specific to a device. (Think of something like `NRIActivity`, due to which the existing PodConfigStore is also structured in equivalent manner)

gauravkghildiyal (commented):
Thanks for the patience @rbtr! This is looking really cool

Haven't reviewed the tests yet, but I thought I'll share this before. I was thinking about the `PodConfigStore` setup and instead of the current "either-or" approach where we choose between an in-memory or a Bolt implementation at startup, what if we just have one in-memory store that handles all the logic and just uses a persistence backend (Bolt being an implementation of that persistence backend interface)?

Something like: `driver -> PodConfigStore -> BoltStore`

There's a couple of reasons why I'm thinking of structuring things that way:

1.  GetPodConfig is in the hot path of all NRI hooks. This function is a bottleneck for `RunPodSandbox`, `CreateContainer`, and `StopPodSandbox`, which are currently invoked for each and every pod on the node, regardless of whether it uses DRA or not. In the current `BoltPodConfigStore` implementation, we are forced to fetch from the DB and perform data unmarshalling for each of these operations. Having a local cache (like `PodConfigStore`) would allow us to serve these requests from RAM instantly.

2.  As things are implemented right now, we are forced to implement ephemeral state like `NRIActivity` into the entire `BoltPodConfigStore`. This mixing of concerns smells like it needs some better structuring. If we were to have `PodConfigStore` consume a persistent storage like `BoltStore`, we could avoid implementing this in the persistence layer entirely and keep the ephemeral state strictly in the top-level store.

Also, it looks like Kubelet does something similar for DRA in [claiminfo.go](https://github.com/kubernetes/kubernetes/blob/2f82dc6dce33cb00b51ce4fbe40c49f87ea4fe19/pkg/kubelet/cm/dra/claiminfo.go#L240-L250). Basically keeping the state in RAM and using the disk as a backup so it survives restarts. 

In the proposed implementation, I imagine:
- Most of the reads in this case will be served directly from memory.
- The interface between `PodConfigStore -> BoltStore` will be quite minimal

pkg/driver/pod_device_config.go
```diff
@@ -63,36 +64,80 @@ type RDMAConfig struct {
...
+		configs:      make(map[types.UID]PodConfig),
+		checkpointer: checkpointer,
+	}
+
+	if checkpointer != nil {
+		saved, err := checkpointer.GetOrCreate()
+		if err != nil {
+			return nil, err
+		}
+		for podUID, devices := range saved {
```
gauravkghildiyal, line 125: Can we please log the podUIDs which we load back at startup

pkg/driver/nri_hooks.go
```diff
@@ -406,6 +417,7 @@ func (np *NetworkDriver) RemovePodSandbox(ctx context.Context, pod *api.PodSandb
 
 func (np *NetworkDriver) removePodSandbox(_ context.Context, pod *api.PodSandbox) error {
 	np.netdb.RemovePodNetNs(podKey(pod))
+	np.podConfigStore.DeletePod(types.UID(pod.GetUid()))
```
gauravkghildiyal: Today, podConfigs are generated as part of the dra_hooks invocation, and consequently they are also removed ruring the UnprepareResourceClaim as part of cleaning up the claims. I say lets keep things that way for now to allow creation and removal to be handled during the same lifecycle phase (DRA, instead of NRI).

README.md
```diff
@@ -103,6 +103,15 @@ helm upgrade --install dranet ./deployments/helm/dranet -n kube-system
 
 For available configuration options, see the [chart README](deployments/helm/dranet/README.md).
 
+DRANET persists prepared pod device configuration in a local bbolt database so
```
gauravkghildiyal, line 106: Thanks for adding this!

Can we include this (and any other information you find useful) in a short release notes section of the PR description (We have a new PR template now https://raw.githubusercontent.com/kubernetes-sigs/dranet/refs/heads/main/.github/PULL_REQUEST_TEMPLATE.md)

pkg/driver/pod_device_config_bolt_test.go
```diff
@@ -0,0 +1,387 @@
...
+		t.Fatalf("Store() error: %v", err)
+	}
+	data, err = cp.GetOrCreate()
+	if err != nil {
+		t.Fatalf("GetOrCreate() error: %v", err)
+	}
+	if len(data) != 1 {
+		t.Fatalf("expected 1 pod, got %d", len(data))
+	}
+	if !reflect.DeepEqual(data[podUID][deviceName], config) {
```
gauravkghildiyal: Consider using cmp.Diff instead of DeepEqual for better test log output (Applicable to other new additions in the PR). Example: https://github.com/kubernetes-sigs/dranet/blob/df8cf66aebaa379de57eef4619ecd58bcb2a47fc/pkg/cloudprovider/gce/gce_test.go#L151

pkg/driver/nri_hooks_test.go
```diff
@@ -65,6 +67,164 @@ func TestCreateContainerNoDuplicateDevices(t *testing.T) {
 	}
 }
 
+func TestCreateContainerUsesPersistedConfigAfterRestart(t *testing.T) {
```
gauravkghildiyal, line 70: Great test!

gauravkghildiyal (commented):
Thanks! Looking mostly ready, just a few small requests.

README.md
rbtr: Updated to new format and added release note

pkg/driver/nri_hooks.go
rbtr: Removed

pkg/driver/pod_device_config.go
rbtr: Added

pkg/driver/pod_device_config_bolt_test.go
rbtr: Threw away reflect and used cmp 🙂

aojea:
❤️

PR #116, opened 2026-03-30
Add OKE (Oracle Cloud) provider for dranet

dkennetzoracle:
- Adds cloud provider module that queries OCI IMDS to publish: oke.dra.net/shape, oke.dra.net/faultDomain, and oke.dra.net/availabilityDomain as device attributes on ResourceSlices.

Test matrix for CSP:

- Build and push with changes to team container registry
- Deploy to OKE cluster with active H100 bare metal nodes with RDMA backend
```bash
k get resourceslice
NAME                               NODE          DRIVER           POOL          AGE
10.0.68.50-dra.net-vr8dh           10.0.68.50    dra.net          10.0.68.50    49m
10.0.68.50-gpu.nvidia.com-cb9cz    10.0.68.50    gpu.nvidia.com   10.0.68.50    41m
```
- Check the contents to verify PCIAddress and Device as well as config variables surfaced in code
```bash
k get resourceslice 10.0.68.50-dra.net-vr8dh -o yaml
...
  - attributes:
      dra.net/numaNode:
        int: 0
      dra.net/pciAddress:
        string: 0000:0c:00.0
      dra.net/pciDevice:
        string: MT2910 Family [ConnectX-7]
      dra.net/pciSubsystem:
        string: "0038"
      dra.net/pciVendor:
        string: Mellanox Technologies
      dra.net/rdma:
        bool: true
      oke.dra.net/availabilityDomain:
        string: TrcQ:US-ASHBURN-AD-2
      oke.dra.net/faultDomain:
        string: FAULT-DOMAIN-1
      oke.dra.net/shape:
        string: BM.GPU.H100.8
      resource.kubernetes.io/pcieRoot:
        string: pci0000:07
```
- run NCCL tests with new resource slices to verify perf.

### DIFF
--- a/pkg/inventory/cloud.go
+++ b/pkg/inventory/cloud.go
@@ -29,22 +29,29 @@ import (
 	"sigs.k8s.io/dranet/pkg/cloudprovider"
 	"sigs.k8s.io/dranet/pkg/cloudprovider/azure"
 	"sigs.k8s.io/dranet/pkg/cloudprovider/gce"
+	"sigs.k8s.io/dranet/pkg/cloudprovider/oke"
 )
 
 // getInstanceProperties get the instace properties and stores them in a global variable to be used in discovery
 func getInstanceProperties(ctx context.Context) cloudprovider.CloudInstance {
 	var err error
 	var instance cloudprovider.CloudInstance
-	if metadata.OnGCE() {
+	switch {
+	case metadata.OnGCE():
 		// Get google compute instance metadata for network interfaces
 		// https://cloud.google.com/compute/docs/metadata/predefined-metadata-keys
 		klog.Infof("running on GCE")
 		instance, err = gce.GetInstance(ctx)
-	} else if azure.OnAzure(ctx) {
+	case azure.OnAzure(ctx):
 		// Get Azure instance metadata for placement group and VM size
 		// https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service
 		klog.Infof("running on Azure")
 		instance, err = azure.GetInstance(ctx)
+	case oke.OnOKE(ctx):
+		// Get OCI instance metadata for shape, fault domain, and availability domain
+		// https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/gettingmetadata.htm
+		klog.Infof("running on OKE")
+		instance, err = oke.GetInstance(ctx)
 	}
 	if err != nil {
 		klog.Infof("could not get instance properties: %v", err)
--- a/pkg/cloudprovider/oke/oke.go
+++ b/pkg/cloudprovider/oke/oke.go
@@ -1,0 +1,154 @@
+/*
+Copyright The Kubernetes Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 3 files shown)


pkg/inventory/cloud.go
```diff
@@ -29,22 +29,29 @@ import (
...
 	"sigs.k8s.io/dranet/pkg/cloudprovider/gce"
+	"sigs.k8s.io/dranet/pkg/cloudprovider/oke"
 )
 
 // getInstanceProperties get the instace properties and stores them in a global variable to be used in discovery
 func getInstanceProperties(ctx context.Context) cloudprovider.CloudInstance {
 	var err error
 	var instance cloudprovider.CloudInstance
-	if metadata.OnGCE() {
+	switch {
```
michaelasp, line 39: When is `getInstanceProperties` called? I want to be sure that we aren't adding an artificial 10-15 second delay on detecting any device now that we have 3 cloud providers. If this is called on every device, I'd like to maybe see us detect it once and default to that cloud provider. Let me look and see how this code is used again.

michaelasp: So it seems like this can start adding a decent bit of delay as we add more cloud providers, I don't think it's blocking for this change necesarily but would like to handle this before we start adding too many.

 @gauravkghildiyal wdyt is a better way to handle this? We know that we will only be on one cloud provider at any point, maybe during the scan we should set a durable value knowing which cloud provider the node resides on?

dkennetzoracle: @michaelasp it's called at startup and then values are re-used from db.instance, so in memory after. There will be an artificial delay on startup of about 5s per CSP. 

1. getInstanceProperties called once at startup before polling loop
```
pkg/inventory/db.go:171 — inside DB.Run(), before the polling loop:
  // Obtain data that will not change after the startup                                                                
  db.instance = getInstanceProperties(ctx)
  ```
2. polling loop starts at db.go:175
```
  for {                                                                                                                
      err := db.rateLimiter.Wait(ctx)                                                                                  
      ...                                                                                                              
      filteredDevices := db.scan()
```
3. db.instance is then read (never written again) in two places during each scan cycle:
- pkg/inventory/db.go:458 — addCloudAttributes() calls getProviderAttributes(device, db.instance) per device. This just reads db.instance.GetDeviceAttributes(), in-memory map lookup.
- pkg/inventory/db.go:471-482 — updateDeviceStore() calls db.instance.GetDeviceConfig() per device. Also in-memory lookup

michaelasp: ~~Yeah it's mostly in relation to adding new devices that would then take 5*(# of cloud providers) seconds to run the scan loop before updating. I'm ok with adding this for the time being but may do a refactor to make it so we only have to figure out the cloud provider once. I'll approve this though since it's no different than what was done for Azure.~~

Nvm, just traced it and saw the same as you. This is perfect as is then. Thanks!

dkennetzoracle: Got it, yeah that will kind of blow up over time. Might make sense to kick the scan logic towards the CSP in the future. IE, initial loop stores the CSP (maybe during install), and scan logic references a func map based on CSP.

michaelasp: +1, it's just that eventually this will impact startup time(much lower impact than what I thought initially but should still be addressed before too many cloud providers are added)

PR #117, opened 2026-03-30
deployments: add Helm chart for DRANET

### LINKED ISSUE #113: proposal: add Helm chart
Add a Helm chart for the driver installation + for easier parameter tuning during installation. The steps that would be taken apart from the chart itself would be adding some CI linters (pre-submit preferably), agreeing on chart hosting (I assume we could just rely on GHCR) and automating chart artifacts to be published automatically as part of a release process.
Happy to jump in on this.

/assign

fmuyassarov:
Add a Helm chart under deployments/helm/dranet as an alternative to the existing static install.yaml. Also introduce a CI workflow to run helm lint for validation. As a follow-up, I’m preparing a patch to publish chart artifacts as OCI artifacts as part of the release process. This will require some additional work on the registry side which I will take care, as well as integration with [Artifact Hub.](https://artifacthub.io/).

fixes: https://github.com/kubernetes-sigs/dranet/issues/113

### DIFF
--- a/.github/workflows/helm-lint.yml
+++ b/.github/workflows/helm-lint.yml
@@ -1,0 +1,18 @@
+name: Helm Lint
+on:
+  pull_request:
+    branches: [ main ]
+    paths:
+      - 'deployments/helm/**'
+
+jobs:
+  helm-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Install Helm
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5.0.0
+
+      - name: Run Helm lint
+        run: make helm-lint
--- a/deployments/helm/dranet/Chart.yaml
+++ b/deployments/helm/dranet/Chart.yaml
@@ -1,0 +1,18 @@
+apiVersion: v2
+name: dranet
+description: Helm chart for deploying DRANET, a Kubernetes network driver using Dynamic Resource Allocation (DRA)
+type: application
+version: 0.1.0
+appVersion: "stable"
+home: https://github.com/kubernetes-sigs/dranet
+sources:
+  - https://github.com/kubernetes-sigs/dranet
+keywords:
+  - networking
+  - dra
+  - dynamic-resource-allocation
+maintainers:
+  - name: aojea
+  - name: gauravkghildiyal
+  - name: michaelasp
+  - name: fmuyassarov
--- a/deployments/helm/dranet/templates/daemonset.yaml
+++ b/deployments/helm/dranet/templates/daemonset.yaml
@@ -1,0 +1,96 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "dranet.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "dranet.labels" . | nindent 4 }}
+    tier: node
+    app: {{ include "dranet.name" . }}
+    k8s-app: {{ include "dranet.name" . }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "dranet.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "dranet.labels" . | nindent 8 }}
+        {{- include "dranet.selectorLabels" . | nindent 8 }}
+        tier: node
+        k8s-app: {{ include "dranet.name" . }}
+        {{- with .Values.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    spec:
+      hostNetwork: true
+      serviceAccountName: {{ include "dranet.serviceAccountName" . }}
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      tolerations:
+        {{- toYaml .Values.tolerations | nindent 8 }}
+      containers:
+        - name: dranet
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args:
+            - /dranet
+            - --v={{ .Values.logVerbosity }}
+            - --hostname-override=$(NODE_NAME)
+          env:
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          resources:
+            requests:
+              {{- toYaml .Values.resources.requests | nindent 14 }}
+            {{- with .Values.resources.limits }}
+            limits:
+              {{- toYaml . | nindent 14 }}
+            {{- end }}
+          securityContext:
+            privileged: true
+          readinessProbe:
+            {{- toYaml .Values.readinessProbe | nindent 12 }}
+          volumeMounts:
+            - name: device-plugin
+              mountPath: /var/lib/kubelet/plugins
+            - name: plugin-registry
+              mountPath: /var/lib/kubelet/plugins_registry
+            - name: nri-plugin
+              mountPath: /var/run/nri
+            - name: netns
+              mountPath: /var/run/netns
+              mountPropagation: HostToContainer
+            - name: infiniband
+              mountPath: /dev/infiniband
+              mountPropagation: HostToContainer
+            - name: bpf-programs
+              mountPath: /sys/fs/bpf
+              mountPropagation: HostToContainer
+      volumes:
+        - name: device-plugin
+          hostPath:
+            path: /var/lib/kubelet/plugins
+        - name: plugin-registry
+          hostPath:
+            path: /var/lib/kubelet/plugins_registry
+        - name: nri-plugin
+          hostPath:
+            path: /var/run/nri
+        - name: netns
+          hostPath:
+            path: /var/run/netns
+        - name: infiniband
+          hostPath:
+            path: /dev/infiniband
+        - name: bpf-programs
+          hostPath:
+            path: /sys/fs/bpf
--- a/deployments/helm/dranet/values.yaml
+++ b/deployments/helm/dranet/values.yaml
@@ -1,0 +1,35 @@
+nameOverride: ""
+fullnameOverride: ""
+
+image:
+  repository: registry.k8s.io/networking/dranet
+  tag: stable
+  pullPolicy: IfNotPresent
+
+imagePullSecrets: []
+
+rbac:
+  create: true
+
+podAnnotations: {}
+podLabels: {}
+
+logVerbosity: 4
+
+tolerations:
+  - operator: Exists
+    effect: NoSchedule
+
+resources:
+  requests:
+    cpu: 100m
+    memory: 50Mi
+  limits: {}
+
+readinessProbe:
+  httpGet:
+    path: /healthz
+    port: 9177
+
+serviceAccount:
+  annotations: {}
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,9 @@ e2e-test:
 lint:
 	hack/lint.sh
 
+helm-lint:
+	helm lint --strict deployments/helm/dranet
+
 update:
 	go mod tidy
 
--- a/README.md
+++ b/README.md
@@ -95,6 +95,14 @@ Install the latest stable version of DRANET using the provided manifest:
 kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/dranet/refs/heads/main/install.yaml
 ```
 
+To install using the Helm chart from a local tree:
+
+```sh
+helm upgrade --install dranet ./deployments/helm/dranet -n kube-system
+```
+
+For available configuration options, see the [chart README](deployments/helm/dranet/README.md).
+
 ### How to Use It
 
 Once DRANET is running, you can inspect the network interfaces and their
--- a/deployments/helm/dranet/.helmignore
+++ b/deployments/helm/dranet/.helmignore
@@ -1,0 +1,6 @@
+.DS_Store
+.git/
+.gitignore
+*.swp
+*.bak
+*.tmp
--- a/deployments/helm/dranet/README.md
+++ b/deployments/helm/dranet/README.md
@@ -1,0 +1,40 @@
+# DRANET Helm Chart
+
+## Installation
+
+From a local checkout:
+
+```sh
+helm upgrade --install dranet ./deployments/helm/dranet -n kube-system
+```
+
+## Configuration
+
+The following table lists the configurable parameters and their default values:
+
+| Parameter | Description | Default |
+|-----------|-------------|---------|
+| `nameOverride` | Override the chart name | `""` |
+| `fullnameOverride` | Override the full release name | `""` |
+| `image.repository` | Container image repository | `registry.k8s.io/networking/dranet` |
+| `image.tag` | Container image tag | `stable` |
+| `image.pullPolicy` | Image pull policy | `IfNotPresent` |
+| `imagePullSecrets` | List of image pull secrets | `[]` |
+| `rbac.create` | Create RBAC resources | `true` |
+| `podAnnotations` | Annotations to add to pods | `{}` |
+| `podLabels` | Labels to add to pods | `{}` |
+| `logVerbosity` | Log verbosity level | `4` |
+| `tolerations` | Pod tolerations | `[{operator: Exists, effect: NoSchedule}]` |
+| `resources.requests.cpu` | CPU resource request | `100m` |
+| `resources.requests.memory` | Memory resource request | `50Mi` |
+| `resources.limits.cpu` | CPU resource limit | `""` (not set) |
+| `resources.limits.memory` | Memory resource limit | `""` (not set) |
+| `readinessProbe.httpGet.path` | Readiness probe HTTP path | `/healthz` |
+| `readinessProbe.httpGet.port` | Readiness probe HTTP port | `9177` |
+
+Parameters can be set at install time using `--set` or a custom values file:
+
+```sh
+helm upgrade --install dranet ./deployments/helm/dranet -n kube-system --set logVerbosity=6
+helm upgrade --install dranet ./deployments/helm/dranet -n kube-system -f my-values.yaml
+```
--- a/deployments/helm/dranet/templates/_helpers.tpl
+++ b/deployments/helm/dranet/templates/_helpers.tpl
@@ -1,0 +1,36 @@
+{{- define "dranet.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{- define "dranet.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- if contains $name .Release.Name -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "dranet.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{- define "dranet.labels" -}}
+helm.sh/chart: {{ include "dranet.chart" . }}
+app.kubernetes.io/name: {{ include "dranet.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end -}}
+
+{{- define "dranet.selectorLabels" -}}
+app: {{ include "dranet.name" . }}
+{{- end -}}
+
+{{- define "dranet.serviceAccountName" -}}
+{{- include "dranet.fullname" . -}}
+{{- end -}}
--- a/deployments/helm/dranet/templates/rbac.yaml
+++ b/deployments/helm/dranet/templates/rbac.yaml
@@ -1,0 +1,54 @@
+{{- if .Values.rbac.create }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "dranet.fullname" . }}
+  labels:
+    {{- include "dranet.labels" . | nindent 4 }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - nodes
+    verbs:
+      - get
+  - apiGroups:
+      - resource.k8s.io
+    resources:
+      - resourceslices
+    verbs:
+      - list
+      - watch
+      - create
+      - update
+      - delete
+  - apiGroups:
+      - resource.k8s.io
+    resources:
+      - resourceclaims
+      - deviceclasses
+    verbs:
+      - get
+  - apiGroups:
+      - resource.k8s.io
+    resources:
+      - resourceclaims/status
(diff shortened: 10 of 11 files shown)


.github/workflows/helm-lint.yml
```diff
@@ -0,0 +1,18 @@
...
+      - 'deployments/helm/**'
+
+jobs:
+  helm-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Install Helm
+        uses: azure/setup-helm@v4
```
michaelasp: Can we use the sha like above? Just to prevent security issues in github actions.

deployments/helm/dranet/values.yaml
```diff
@@ -0,0 +1,43 @@
...
+  privileged: true
+  capabilities:
+    add: []
+
+readinessProbe:
+  httpGet:
+    path: /healthz
+    port: 9177
+
+containerdNRI:
```
michaelasp: Maybe make this dranetHandleEnableNRI?

This specifically creates the init container that handles enabling NRI for the node, if we enable this we want the user to understand that DRANet will be toggling flags in the node relating to NRI to not catch them off guard.

deployments/helm/dranet/values.yaml
```diff
@@ -0,0 +1,43 @@
+nameOverride: ""
+fullnameOverride: ""
+
+image:
+  repository: registry.k8s.io/networking/dranet
+  tag: stable
```
michaelasp, line 6: Should we make this a semver tag we bump with version changes? @aojea or @gauravkghildiyal feel free to chime in. I think this is fine though

deployments/helm/dranet/values.yaml
```diff
@@ -0,0 +1,43 @@
...
+    effect: NoSchedule
+
+resources:
+  requests:
+    cpu: 100m
+    memory: 50Mi
+  limits: {}
+
+securityContext:
+  privileged: true
```
michaelasp: Not against this but we always have to run this container as privileged right? Is there a point in having this configurable?

michaelasp (commented):
Lgtm overall, minor comments.

deployments/helm/dranet/values.yaml
aojea: I use the trick of tag `stable` always to the latest 

https://github.com/kubernetes/k8s.io/blob/0e8e0d9cc755db67a387e4497834b87eb936e80f/registry.k8s.io/images/k8s-staging-networking/images.yaml#L3

deployments/helm/dranet/values.yaml
aojea: I do not think we need to use this, NRI is enabled by default in containerd 2.x and new crio versions ... and this init container was needed for supporting a GKE version that still has contianerd 1.7.x, let's remove the init container

deployments/helm/dranet/templates/daemonset.yaml
```diff
@@ -0,0 +1,137 @@
...
+                 printf '%s\n' "[plugins.\"io.containerd.nri.v1.nri\"]" "  disable = false" "  disable_connections = false" "  plugin_config_path = \"/etc/nri/conf.d\"" "  plugin_path = \"/opt/nri/plugins\"" "  plugin_registration_timeout = \"5s\"" "  plugin_request_timeout = \"5s\"" "  socket_path = \"/var/run/nri/nri.sock\"" >> /etc/containerd/config.toml
+                 echo "restarting containerd"
+                 nsenter -t 1 -m -u -i -n -p -- systemctl restart containerd
+              fi
+          securityContext:
+            privileged: true
+          volumeMounts:
+            - name: etc
+              mountPath: /etc
+      {{- end }}
```
aojea:
```suggestion
```

let's remove this hack for enabling nri, it only applies to pre 2.x containerd versions

fmuyassarov: Just heads up, this is optional... and doesn't run by default. My reasoning for adding this was [install-containerd-1.7yaml](https://github.com/kubernetes-sigs/dranet/blob/main/install-containerd-1.7yaml) file.

But I totally agree that, we should remove it anyways from the driver code base, as CRI-O has it on by default for a long time and containerd too.

deployments/helm/dranet/values.yaml
michaelasp: Sgtm, lets keep this then

deployments/helm/dranet/values.yaml
fmuyassarov: I fully agree that it is only in rare cases that NRI may not be turned on and for that reason I thought I would still add it as optional and disabled by default (in other words running init container). So, you prefer to have it totally skipped right @aojea ?

deployments/helm/dranet/values.yaml
fmuyassarov: You are right about privileged mode. However, I added this part so that people can add capabilities depending on their setup. And I was not fully sure about giving possibility to tune capabilities or should we expect capabilities to be different at all? And my answer was yes - based on quick search I found, for example this one: https://github.com/kubernetes-sigs/dranet/blob/eb0eddaf041fd633b813c8d70b7679b8f108fe18/site/content/docs/user/gke-rdma.md?plain=1#L198

fmuyassarov: But it is worth of removing `privileged: true` at least...

deployments/helm/dranet/templates/daemonset.yaml
fmuyassarov: ok, I've now cleaned up this part to keep it tide.

.github/workflows/helm-lint.yml
fmuyassarov: certainly, fixed it.

deployments/helm/dranet/values.yaml
fmuyassarov: removed everything around init container

deployments/helm/dranet/values.yaml
fmuyassarov: @michaelasp I've now kept only 
```YAML
securityContext:
  capabilities:
    add: []
```
but I'm also fine to drop the whole section of capabilities if we think that people won't need to alter them

michaelasp: ~~I think that should be good!~~

I think we still need `privileged: true` in the pod spec right?

fmuyassarov: actually not necessarily, capabilities can be added directly, example https://github.com/kubernetes-sigs/dranet/blob/eb0eddaf041fd633b813c8d70b7679b8f108fe18/examples/dranetctl-install.yaml#L121-L123. 

But if you think this should be done separately outside of this PR, happy to do it on a follow up

fmuyassarov:
> I think we still need the priviliged: true in the pod spec

replied on the [thread](https://github.com/kubernetes-sigs/dranet/pull/117#discussion_r3011950067) above, it can be and probably should be dropped actually, unless there is a good reason for keeping it.

deployments/helm/dranet/values.yaml
michaelasp: In that case, should we have it as a default so the user doesn't have to override the capabilities? Otherwise it wouldn't work out the box with a helm install right?

fmuyassarov: I think it should work because 
```YAML
resources:
  requests:
    cpu: 100m
    memory: 50Mi
securityContext:
readinessProbe:
  httpGet:
    path: /healthz
    port: 9177
```
is syntactically valid YAML but might not make things like Kyverno or kube-linter happy...
I can fix this quickly.

fmuyassarov: okay, now we don't end up with empty key in case user doesn't specify a capability to add. Is this better now WDYT @michaelasp ?

fmuyassarov: now it will looks something like this when specified
```YAML
securityContext:
  capabilities:
    add:
      - NET_ADMIN
      - SYS_PTRACE
      - SYS_ADMIN
```
otherwise no empty `securityContext` key.

deployments/helm/dranet/values.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+resources:
+  requests:
+    cpu: 100m
+    memory: 50Mi
+  limits: {}
+
+securityContext:
+  capabilities:
+    add: []
```
michaelasp: I'm still unclear here, don't we need to add the capabilities here or else we will have a fully unprivileged pod by default? Or am I missing something obvious 😅

We want the helm chart to by default grant those privileges right?

fmuyassarov: Sorry for any confusion earlier - let me clarify:
- At the moment, we don’t require any specific capabilities in install.yaml, so I followed the same approach in Helm by not enabling any capabilities by default.
- The current install.yaml includes privileged: true, which I intentionally omitted since it doesn’t seem necessary in our case. This is the only difference between install.yaml and the Helm configuration. If you think we should not do it here, happy to submit a separate patch for thatl
- I’m adding an option in values.yaml to allow users to specify capabilities if needed - for example, similar to the [dranetctl](https://github.com/kubernetes-sigs/dranet/blob/d99fb225d74465e89442c4717a262c4bb07a2c27/examples/dranetctl-install.yaml#L123) use case.

Example run from my current tree on kind cluster

```
~/dranet$ helm upgrade --install dranet ./deployments/helm/dranet -n kube-system
Release "dranet" does not exist. Installing it now.
NAME: dranet
LAST DEPLOYED: Mon Mar 30 21:18:53 2026
NAMESPACE: kube-system
STATUS: deployed
REVISION: 1
TEST SUITE: None

~/dranet$ kubectl get pods -A
NAMESPACE            NAME                                         READY   STATUS    RESTARTS   AGE
kube-system          coredns-7d764666f9-9wvdb                     1/1     Running   0          14h
kube-system          coredns-7d764666f9-j9jw4                     1/1     Running   0          14h
kube-system          dranet-5xqhx                                 1/1     Running   0          4s
kube-system          dranet-gxchm                                 1/1     Running   0          4s
kube-system          dranet-rwk46                                 1/1     Running   0          4s
kube-system          etcd-kind-control-plane                      1/1     Running   0          14h
kube-system          kindnet-7q6fk                                1/1     Running   0          14h
kube-system          kindnet-mlr6l                                1/1     Running   0          14h
kube-system          kindnet-n6d9d                                1/1     Running   0          14h
kube-system          kube-apiserver-kind-control-plane            1/1     Running   0          14h
kube-system          kube-controller-manager-kind-control-plane   1/1     Running   0          14h
kube-system          kube-proxy-6bpbq                             1/1     Running   0          14h
kube-system          kube-proxy-dxcc5                             1/1     Running   0          14h
kube-system          kube-proxy-vn9q9                             1/1     Running   0          14h
kube-system          kube-scheduler-kind-control-plane            1/1     Running   0          14h
local-path-storage   local-path-provisioner-84669cbbb8-qj485      1/1     Running   0          14h
```


```
kubectl logs -n kube-system  dranet-5xqhx -f
I0330 21:18:54.358937       1 app.go:199] dranet go go1.24.11 build: bb7e8b7af0c571b9e474981064e26ad9dc85c21f time: 2025-11-26T21:00:20Z
I0330 21:18:54.359071       1 app.go:85] FLAG: --add_dir_header="false"
I0330 21:18:54.359079       1 app.go:85] FLAG: --alsologtostderr="false"
I0330 21:18:54.359082       1 app.go:85] FLAG: --bind-address=":9177"
I0330 21:18:54.359087       1 app.go:85] FLAG: --filter="!(\"dra.net/type\" in attributes) || attributes[\"dra.net/type\"].StringValue  != \"veth\""
I0330 21:18:54.359096       1 app.go:85] FLAG: --hostname-override="kind-worker2"
I0330 21:18:54.359098       1 app.go:85] FLAG: --inventory-max-poll-interval="1m0s"
I0330 21:18:54.359102       1 app.go:85] FLAG: --inventory-min-poll-interval="2s"
I0330 21:18:54.359105       1 app.go:85] FLAG: --inventory-poll-burst="5"
I0330 21:18:54.359110       1 app.go:85] FLAG: --kubeconfig=""
I0330 21:18:54.359112       1 app.go:85] FLAG: --log_backtrace_at=":0"
I0330 21:18:54.359119       1 app.go:85] FLAG: --log_dir=""
I0330 21:18:54.359122       1 app.go:85] FLAG: --log_file=""
I0330 21:18:54.359124       1 app.go:85] FLAG: --log_file_max_size="1800"
I0330 21:18:54.359127       1 app.go:85] FLAG: --logtostderr="true"
I0330 21:18:54.359129       1 app.go:85] FLAG: --one_output="false"
I0330 21:18:54.359131       1 app.go:85] FLAG: --skip_headers="false"
I0330 21:18:54.359133       1 app.go:85] FLAG: --skip_log_headers="false"
I0330 21:18:54.359135       1 app.go:85] FLAG: --stderrthreshold="2"
I0330 21:18:54.359146       1 app.go:85] FLAG: --v="4"
I0330 21:18:54.359154       1 app.go:85] FLAG: --vmodule=""
I0330 21:18:54.360608       1 pcidb.go:55] Successfuly set value of PCIDB_PATH="/tmp/pcidb2052773308/pci.ids.gz"
I0330 21:18:54.364148       1 envvar.go:172] "Feature gate default state" feature="ClientsAllowCBOR" enabled=false
I0330 21:18:54.364196       1 envvar.go:172] "Feature gate default state" feature="ClientsPreferCBOR" enabled=false
I0330 21:18:54.364213       1 envvar.go:172] "Feature gate default state" feature="InOrderInformers" enabled=true
I0330 21:18:54.364222       1 envvar.go:172] "Feature gate default state" feature="InformerResourceVersion" enabled=false
I0330 21:18:54.364235       1 envvar.go:172] "Feature gate default state" feature="WatchListClient" enabled=false
I0330 21:18:54.370916       1 driver.go:109] RDMA subsystem in mode: shared
I0330 21:18:54.371080       1 draplugin.go:597] "Starting"
I0330 21:18:54.371256       1 nonblockinggrpcserver.go:90] "GRPC server started" logger="dra" endpoint="/var/lib/kubelet/plugins/dra.net/dra.sock"
I0330 21:18:54.371333       1 nonblockinggrpcserver.go:90] "GRPC server started" logger="registrar" endpoint="/var/lib/kubelet/plugins_registry/dra.net-reg.sock"
time="2026-03-30T21:18:55Z" level=info msg="Created plugin 00-dra.net (dranet, handles RunPodSandbox,StopPodSandbox,RemovePodSandbox,CreateContainer)"
I0330 21:18:55.372688       1 app.go:174] driver started
I0330 21:18:55.372740       1 dra_hooks.go:56] Publishing resources
time="2026-03-30T21:18:55Z" level=info msg="Registering plugin 00-dra.net..."
time="2026-03-30T21:18:55Z" level=info msg="Configuring plugin 00-dra.net for runtime containerd/v2.2.1..."
time="2026-03-30T21:18:55Z" level=info msg="Started plugin 00-dra.net..."
I0330 21:18:55.393920       1 nri_hooks.go:42] Synchronized state with the runtime (4 pods, 3 containers)...
I0330 21:18:55.395231       1 nri_hooks.go:46] Synchronize Pod kube-system/kube-proxy-dxcc5 UID b3382c71-dfa5-44ff-8eb2-8cab1fbbf7dd
I0330 21:18:55.395263       1 nri_hooks.go:47] pod kube-system/kube-proxy-dxcc5: namespace= ips=[]
I0330 21:18:55.395273       1 nri_hooks.go:46] Synchronize Pod kube-system/dranet-5xqhx UID dceb3d53-94f8-4b81-8103-90c7ef185fb2
I0330 21:18:55.395280       1 nri_hooks.go:47] pod kube-system/dranet-5xqhx: namespace= ips=[]
I0330 21:18:55.395285       1 nri_hooks.go:46] Synchronize Pod kube-system/kindnet-7q6fk UID d76323c1-79ac-4cde-9a23-faaa29c04eb4
I0330 21:18:55.395290       1 nri_hooks.go:47] pod kube-system/kindnet-7q6fk: namespace= ips=[]
I0330 21:18:55.395295       1 nri_hooks.go:46] Synchronize Pod kube-system/dranet-sznnd UID ca5ed312-d645-4849-ba19-646928945471
I0330 21:18:55.395299       1 nri_hooks.go:47] pod kube-system/dranet-sznnd: namespace= ips=[]
I0330 21:18:55.395773       1 db.go:155] Default gateway interfaces: [eth0]
I0330 21:18:55.558407       1 db.go:293] Network Interface lo is a loopback interface, excluding it from discovery
W0330 21:18:55.565989       1 cloud.go:53] instance metadata is nil, cannot get provider attributes.
I0330 21:18:55.566017       1 db.go:197] Ignoring interface eth0 from discovery since it is an uplink interface
I0330 21:18:55.566029       1 db.go:207] Found 10 devices
I0330 21:18:55.566065       1 dra_hooks.go:60] Got 10 devices from inventory: pci-0000-02-00-0, pci-0000-02-00-1, pci-0000-0d-00-0, pci-0000-0d-00-1, pci-0000-0d-00-2, pci-0000-0d-00-3, pci-0000-b5-00-0, pci-0000-b5-00-1, pci-0000-b5-00-2, pci-0000-b5-00-3
I0330 21:18:55.566162       1 dra_hooks.go:62] After filtering, publishing 10 devices in ResourceSlice(s): pci-0000-02-00-0, pci-0000-02-00-1, pci-0000-0d-00-0, pci-0000-0d-00-1, pci-0000-0d-00-2, pci-0000-0d-00-3, pci-0000-b5-00-0, pci-0000-b5-00-1, pci-0000-b5-00-2, pci-0000-b5-00-3
I0330 21:18:55.566298       1 resourceslicecontroller.go:528] "Starting ResourceSlice informer and waiting for it to sync" logger="ResourceSlice controller"
I0330 21:18:55.566414       1 reflector.go:358] "Starting reflector" type="*v1.ResourceSlice" resyncPeriod="0s" reflector="pkg/mod/k8s.io/client-go@v0.34.2/tools/cache/reflector.go:290"
I0330 21:18:55.566425       1 reflector.go:404] "Listing and watching" type="*v1.ResourceSlice" reflector="pkg/mod/k8s.io/client-go@v0.34.2/tools/cache/reflector.go:290"
I0330 21:18:55.580763       1 reflector.go:436] "Caches populated" type="*v1.ResourceSlice" reflector="pkg/mod/k8s.io/client-go@v0.34.2/tools/cache/reflector.go:290"
I0330 21:18:56.567501       1 resourceslicecontroller.go:543] "ResourceSlice informer has synced" logger="ResourceSlice controller"
I0330 21:18:56.567679       1 resourceslicecontroller.go:184] "Starting" logger="ResourceSlice controller"
I0330 21:19:22.363068       1 nri_hooks.go:327] RemovePodSandbox Pod kube-system/dranet-sznnd UID ca5ed312-d645-4849-ba19-646928945471
I0330 21:19:55.714771       1 db.go:293] Network Interface lo is a loopback interface, excluding it from discovery
W0330 21:19:55.722419       1 cloud.go:53] instance metadata is nil, cannot get provider attributes.
I0330 21:19:55.722436       1 db.go:197] Ignoring interface eth0 from discovery since it is an uplink interface
I0330 21:19:55.722445       1 db.go:207] Found 10 devices
I0330 21:19:55.722474       1 dra_hooks.go:60] Got 10 devices from inventory: pci-0000-02-00-0, pci-0000-02-00-1, pci-0000-0d-00-0, pci-0000-0d-00-1, pci-0000-0d-00-2, pci-0000-0d-00-3, pci-0000-b5-00-0, pci-0000-b5-00-1, pci-0000-b5-00-2, pci-0000-b5-00-3
I0330 21:19:55.722526       1 dra_hooks.go:62] After filtering, publishing 10 devices in ResourceSlice(s): pci-0000-02-00-0, pci-0000-02-00-1, pci-0000-0d-00-0, pci-0000-0d-00-1, pci-0000-0d-00-2, pci-0000-0d-00-3, pci-0000-b5-00-0, pci-0000-b5-00-1, pci-0000-b5-00-2, pci-0000-b5-00-3

```

deployments/helm/dranet/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - dra
+  - dynamic-resource-allocation
+  - cni
+maintainers:
+  - name: aojea
+  - name: bowei
+  - name: danwinship
+  - name: gauravkghildiyal
+  - name: michaelasp
+  - name: thockin
```
aojea: just add yourself, me, micheal and gaurav ...  and remove the others

deployments/helm/dranet/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+version: 0.1.0
+appVersion: "stable"
+home: https://github.com/kubernetes-sigs/dranet
+sources:
+  - https://github.com/kubernetes-sigs/dranet
+keywords:
+  - networking
+  - dra
+  - dynamic-resource-allocation
+  - cni
```
aojea:
```suggestion
```

remove cni, is a misleading concept that leaked into kubernetes as the networking part, that is not true

fmuyassarov: makes sense, removed it.

deployments/helm/dranet/Chart.yaml
fmuyassarov: done.

deployments/helm/dranet/values.yaml
michaelasp: Huh, do we really not need any capabilities? This feels incorrect, I thought we needed them to move devices back and forth from network namespaces. @aojea wdyt? 

Currently this is only checking whether or not we can discover them, but we should ensure the actual functionality works as well. You can run the dra shell tests against a kind cluster built with this helm chart to see if that works.

IMO, I think we need either `privileged: true` or some set of network capabilities.

gauravkghildiyal: +1
We WILL need `priviledged: true`.

fmuyassarov: Okay, to not make things complicated and drifted let's keep privileged: true in Helm as we have it in install.yaml and we can continue the discussion about capabilities separately or here as well. 


How I saw this initially was, capabilities might differ and we should give an option to user to dictate required capabilities. But after your last question, I think it is us, in other words driver writers who really know what capabilities are needed. And in that sense we should not expect users to modify them but just consume them :)

fmuyassarov: added `privileged: true`, back @michaelasp  and thanks for the questions.

michaelasp: Sounds good, sorry for the back and forth!

deployments/helm/dranet/templates/daemonset.yaml
```diff
@@ -0,0 +1,100 @@
...
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      tolerations:
+        {{- toYaml .Values.tolerations | nindent 8 }}
+      containers:
+        - name: dranet
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          args:
```
gauravkghildiyal, line 42: Is it possible to parameterize the additional arguments that get passed here?

deployments/helm/dranet/values.yaml
fmuyassarov: no problem, it was good to clarify it before landing...

deployments/helm/dranet/templates/daemonset.yaml
gauravkghildiyal: (Non blocking)

fmuyassarov: It is, but I don't see a reason - unless we have some args apart from log verbosity which is already parameterized  (`Values.logVerbosity`) and the `hostname-override` which is not expected to be changed. If we happen to add args to the driver we could ensure to add those new args as well similarly to log verbosity.

gauravkghildiyal: We have a couple of args defined at https://github.com/kubernetes-sigs/dranet/blob/d99fb225d74465e89442c4717a262c4bb07a2c27/cmd/dranet/app.go#L66-L73

It may be useful to atleast parameterize the "filter" arg that can help avoid some ugliness at https://github.com/kubernetes-sigs/dranet/blob/d99fb225d74465e89442c4717a262c4bb07a2c27/tests/setup_suite.bash#L44

gauravkghildiyal: We can take this up separately as well. I'd prefer not to extend this PR when it's almost ready to be merged :)

fmuyassarov: Ow my bad then, let's take it separately. I will file a PR to extend the list both in the install and helm. Thanks for catching this @gauravkghildiyal

PR #118, opened 2026-03-30
Provide the ability to hint the cloud provider that is expected to lower startup time.

michaelasp:
When reviewing https://github.com/kubernetes-sigs/dranet/pull/116 I saw that we don't run these getters in parallel. Since all of them time out after ~5 seconds, worst case scenario we take 5*(# of providers) to start up. Run these in parallel, the network overhead should be minimal and we already reduce the worst case startup time by 10 seconds.

```release-note
Add the cloud-provider-hint command line flag that allows the administrator to specify what cloud they are running on. If specified it will skip discovery, allowing for faster startup.
```

### DIFF
--- a/cmd/dranet/app.go
+++ b/cmd/dranet/app.go
@@ -50,14 +50,15 @@ const (
 )
 
 var (
-	hostnameOverride string
-	kubeconfig       string
-	bindAddress      string
-	celExpression    string
-	minPollInterval  time.Duration
-	maxPollInterval  time.Duration
-	pollBurst        int
-	moveIBInterfaces bool
+	hostnameOverride  string
+	kubeconfig        string
+	bindAddress       string
+	celExpression     string
+	minPollInterval   time.Duration
+	maxPollInterval   time.Duration
+	pollBurst         int
+	moveIBInterfaces  bool
+	cloudProviderHint string
 
 	ready atomic.Bool
 )
@@ -71,6 +72,7 @@ func init() {
 	flag.DurationVar(&maxPollInterval, "inventory-max-poll-interval", 1*time.Minute, "The maximum interval between two consecutive polls of the inventory.")
 	flag.IntVar(&pollBurst, "inventory-poll-burst", 5, "The number of polls that can be run in a burst.")
 	flag.BoolVar(&moveIBInterfaces, "move-ib-interfaces", true, "If true, InfiniBand (IPoIB) network interfaces associated with PCI devices are moved into pod network namespace. If false, moving IB network interfaces are skipped and the underlying device is exposed as an IB-only RDMA device.")
+	flag.StringVar(&cloudProviderHint, "cloud-provider-hint", "", "Hint for the cloud provider that will be used to select the appropriate provider plugin. Supported values: (GCE, AZURE, OKE, NONE). If left unset, the cloud provider is auto-detected (which may delay startup) and defaults to 'NONE' if detection fails.")
 
 	flag.Usage = func() {
 		fmt.Fprint(os.Stderr, "Usage: dranet [options]\n\n")
@@ -165,6 +167,7 @@ func main() {
 		inventory.WithRateLimiter(rate.NewLimiter(rate.Every(minPollInterval), pollBurst)),
 		inventory.WithMaxPollInterval(maxPollInterval),
 		inventory.WithMoveIBInterfaces(moveIBInterfaces),
+		inventory.WithCloudProviderHint(cloudProviderHint),
 	)
 	opts = append(opts, driver.WithInventory(db))
 	dranet, err := driver.Start(ctx, driverName, clientset, nodeName, opts...)
--- a/pkg/inventory/cloud.go
+++ b/pkg/inventory/cloud.go
@@ -32,27 +32,58 @@ import (
 	"sigs.k8s.io/dranet/pkg/cloudprovider/oke"
 )
 
-// getInstanceProperties get the instace properties and stores them in a global variable to be used in discovery
-func getInstanceProperties(ctx context.Context) cloudprovider.CloudInstance {
-	var err error
-	var instance cloudprovider.CloudInstance
-	switch {
-	case metadata.OnGCE():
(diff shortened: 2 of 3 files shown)


pkg/inventory/cloud.go
```diff
@@ -32,27 +32,72 @@ import (
...
+			if err != nil {
+				klog.Infof("could not get instance properties: %v", err)
+				return nil
+			}
+			if instance != nil {
+				break
+			}
+		}
+	} else {
+		provider, ok := providers[cloudProviderHint]
```
michaelasp: In the case of not finding the cloud provider we want, should we fall back to detection of other cloud providers, or should we consider that a misconfig?

Currently we log it as an error and don't try other cloud providers.

pkg/inventory/cloud.go
```diff
@@ -32,27 +32,72 @@ import (
...
+// If the hint is empty, it attempts to discover the cloud provider automatically.
+func getInstanceProperties(ctx context.Context, cloudProviderHint string) cloudprovider.CloudInstance {
+	if cloudProviderHint == CloudProviderHintNone {
+		klog.Infof("cloud provider hint is none, skipping instance properties retrieval")
+		return nil
+	}
+
+	providers := map[string]func(context.Context) (cloudprovider.CloudInstance, error){
+		CloudProviderHintGCE: func(ctx context.Context) (cloudprovider.CloudInstance, error) {
+			if metadata.OnGCE() {
```
gauravkghildiyal: If the flag is explicitly stating an explicit cloud-provider, we should not try to perform an additional check and proceed with what the flag indicated.

With that idea in mind, I think it'll be neater if we decouple things into:
1. Identify the cloudProvider if flag is empty
2. Separately mapping their GetInstance calls.

pkg/inventory/cloud.go
gauravkghildiyal: Better to be explicit with the behaviour and consider this fatal if someone explicitly specified a flag value and it was not found. 

It seems like you are already doing validation in `WithCloudProviderHint()` so we can exit when we don't identify an acceptable value (acceptable value = [<empty>, "NONE"", <other_providers>])

cmd/dranet/app.go
```diff
@@ -71,6 +72,7 @@ func init() {
 	flag.DurationVar(&maxPollInterval, "inventory-max-poll-interval", 1*time.Minute, "The maximum interval between two consecutive polls of the inventory.")
 	flag.IntVar(&pollBurst, "inventory-poll-burst", 5, "The number of polls that can be run in a burst.")
 	flag.BoolVar(&moveIBInterfaces, "move-ib-interfaces", true, "If true, InfiniBand (IPoIB) network interfaces associated with PCI devices are moved into pod network namespace. If false, moving IB network interfaces are skipped and the underlying device is exposed as an IB-only RDMA device.")
+	flag.StringVar(&cloudProviderHint, "cloud-provider-hint", "", "Hint for the cloud provider that will be used to select the appropriate provider plugin. Supported values: (OCI, GCP, AWS, NONE). If not set, the cloud provider will be auto-detected, but may take longer to start.")
```
gauravkghildiyal: Supported values need to be updated to match the actual possible values.

Also, consider:

_"If left unset, the cloud provider is auto-detected (which may delay startup) and defaults to 'NONE' if detection fails."_

pkg/inventory/db.go
```diff
@@ -92,6 +92,11 @@ type DB struct {
 	// When false, IPoIB interfaces are skipped and the underlying device is
 	// exposed as an IB-only RDMA device.
 	moveIBInterfaces bool
+
+	// cloudProviderHint is a hint for the cloud provider that will be used to
+	// select the appropriate provider plugin. Supported values: (OCI, GCP, AWS, NONE).
```
gauravkghildiyal: Requires update of supported values.

pkg/inventory/cloud.go
```diff
@@ -32,27 +32,72 @@ import (
...
 )
 
-// getInstanceProperties get the instace properties and stores them in a global variable to be used in discovery
-func getInstanceProperties(ctx context.Context) cloudprovider.CloudInstance {
+const (
+	CloudProviderHintGCE  = "GCE"
+	CloudProviderHintAzure = "AZURE"
+	CloudProviderHintOKE  = "OKE"
+	CloudProviderHintNone = "NONE"
+)
```
aojea, line 42: Shouldn;t this be an enum?

pkg/inventory/cloud.go
michaelasp: Made it a fatal error in validation in `WithCloudProviderHint()` lmk if this makes sense.

pkg/inventory/cloud.go
michaelasp: I like this approach. Split this into 2 functions with explicit discover vs lookup modes.

cmd/dranet/app.go
michaelasp: Thanks, that's a good way of wording it.

pkg/inventory/db.go
michaelasp: Yep, updated! Don't know why I put AWS 😵‍💫

pkg/inventory/cloud.go
michaelasp: I don't think so, made it a specific type however to make it more understandable. Otherwise we'd have to create mappings from the string to enum value.

pkg/inventory/cloud.go
```diff
@@ -32,27 +32,59 @@ import (
...
+			return hint
+		}
+	}
+
+	klog.Warning("could not discover cloud provider")
+	return CloudProviderHintNone
+}
+
+// getInstanceProperties gets the instance properties using the cloud provider hint if provided.
+// If the hint is empty, it attempts to discover the cloud provider automatically.
```
gauravkghildiyal: This comment may need one tiny correction and we can merge.

michaelasp: Done! Also added a comment above to explain what the other function does.

PR #123, opened 2026-03-31
Create an inventory startup time metric

michaelasp:
Add a startup time metric for the device driver.

Traces to overall latency of startup, everything from creating the plugin path to discovering interfaces the first time around.

### DIFF
--- a/pkg/inventory/metrics.go
+++ b/pkg/inventory/metrics.go
@@ -1,0 +1,40 @@
+/*
+Copyright The Kubernetes Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package inventory
+
+import (
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var registerMetricsOnce sync.Once
+
+func registerMetrics() {
+	registerMetricsOnce.Do(func() {
+		prometheus.MustRegister(inventoryStartupDurationSeconds)
+	})
+}
+
+var (
+	inventoryStartupDurationSeconds = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "dranet",
+		Subsystem: "inventory",
+		Name:      "inventory_startup_duration_seconds",
+		Help:      "The duration of the inventory startup.",
+	})
+)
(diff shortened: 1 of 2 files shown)


pkg/driver/metrics.go
```diff
@@ -87,4 +87,10 @@ var (
 		Name:      "last_published_time_seconds",
 		Help:      "The timestamp of the last successful resource publication.",
 	})
+	startupTime = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "dranet",
+		Subsystem: "driver",
+		Name:      "startup_time_seconds",
+		Help:      "The latency of the driver startup.",
```
fmuyassarov: nit: and most probably sort of a personal preference as such feel free to ignore

```suggestion
		Name:      "startup_duration_second",
		Help:      "The latency of the driver startup.",
```
assuming we try to describe how long it took for the startup and it is closer so standard prometheus naming pattern

pkg/driver/driver.go
```diff
@@ -106,6 +106,10 @@ type Option func(*NetworkDriver)
 
 func Start(ctx context.Context, driverName string, kubeClient kubernetes.Interface, nodeName string, opts ...Option) (*NetworkDriver, error) {
 	registerMetrics()
+	begin := time.Now()
+	defer func() {
+		startupTime.Set(time.Since(begin).Seconds())
+	}()
```
fmuyassarov: we need to register it in [registerMetrics()](https://github.com/kubernetes-sigs/dranet/blob/df472dd1f29a464f15b23fdbe829f3d535450a78/pkg/driver/metrics.go#L42)

michaelasp: Good catch, thanks!

pkg/driver/metrics.go
michaelasp: Agreed, time can be a bit subjective. Updated the naming.

fmuyassarov (approved):
LGTM

pkg/inventory/metrics.go
```diff
@@ -0,0 +1,40 @@
...
+	registerMetricsOnce.Do(func() {
+		prometheus.MustRegister(inventoryStartupDurationSeconds)
+	})
+}
+
+var (
+	inventoryStartupDurationSeconds = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "dranet",
+		Subsystem: "inventory",
+		Name:      "inventory_startup_duration_seconds",
```
gauravkghildiyal, line 37: The final metric name for this would end up being `dranet_inventory_inventory_startup_duration_seconds`. Should we strip the name to `startup_duration_seconds`?

Also, inventory startup name doesn't seem to be doing justice to what its measuring, given it excludes the initial scan completely. If our intent is to measure the cloud-provider detection latency, should we change the name to focus on the cloud provider? (And if we do that, then maybe we can wrap the metric within getInstanceProperties?)

michaelasp: That makes sense, it's more about the cloud provider latency here. Let me update this after the refactor being done with cloud provider hints https://github.com/kubernetes-sigs/dranet/pull/118

gauravkghildiyal: Sounds good!

gauravkghildiyal:
Closing this out since it's not needed at the moment.

PR #124, opened 2026-03-31
chore: expose daemon flags as optional chart values

fmuyassarov:
Expose daemon flags as optional Helm chart values and add schema validation to enforce correct value types to prevent misconfiguration.
Example:
```bash
helm upgrade --install dranet ./deployments/helm/dranet -n kube-system --set args.inventoryMinPollInterval=10

Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
dranet:
- at '/args/inventoryMinPollInterval': got number, want string
```

### DIFF
--- a/deployments/helm/dranet/values.schema.json
+++ b/deployments/helm/dranet/values.schema.json
@@ -1,0 +1,114 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "nameOverride": {
+      "type": "string"
+    },
+    "fullnameOverride": {
+      "type": "string"
+    },
+    "image": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "repository": {
+          "type": "string"
+        },
+        "tag": {
+          "type": "string"
+        },
+        "pullPolicy": {
+          "type": "string",
+          "enum": ["Always", "IfNotPresent", "Never"]
+        }
+      }
+    },
+    "imagePullSecrets": {
+      "type": "array"
+    },
+    "rbac": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "create": {
+          "type": "boolean"
+        }
+      }
+    },
+    "podAnnotations": {
+      "type": "object"
+    },
+    "podLabels": {
+      "type": "object"
+    },
+    "metricsPort": {
+      "type": ["integer", "null"],
+      "minimum": 1,
+      "maximum": 65535
+    },
+    "metricsPath": {
+      "type": "string"
+    },
+    "logVerbosity": {
+      "type": "integer",
+      "minimum": 0
+    },
+    "args": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "filter": {
+          "type": "string",
+          "description": "CEL expression to filter network interface attributes"
+        },
+        "inventoryMinPollInterval": {
+          "type": "string",
+          "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
+          "description": "Go duration string, e.g. '2s', '500ms'"
+        },
+        "inventoryMaxPollInterval": {
+          "type": "string",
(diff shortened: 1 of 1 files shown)


fmuyassarov:
This is a follow up to https://github.com/kubernetes-sigs/dranet/pull/117 based on the [discussion](https://github.com/kubernetes-sigs/dranet/pull/117#discussion_r3016969335)

deployments/helm/dranet/values.yaml
```diff
@@ -15,6 +15,16 @@ podAnnotations: {}
 podLabels: {}
 
 logVerbosity: 4
+metricsPort: ~
+metricsPath: /healthz
+
+# dranet daemon arguments — omit any field to use the binary's built-in default
+args: {}
```
michaelasp, line 22: nit, maybe `daemonArgs just to express that these are args specific to the dranet daemon.

fmuyassarov: I’m not against to this, but longer term are we okay with something like this instead?
```diff
-| `args.filter` | CEL expression to filter network interface attributes | see binary default |
-| `args.inventoryMinPollInterval` | Minimum interval between two consecutive inventory polls | binary default: `2s` |
-| `args.inventoryMaxPollInterval` | Maximum interval between two consecutive inventory polls | binary default: `1m` |
-| `args.inventoryPollBurst` | Number of inventory polls that can be run in a burst | binary default: `5` |
-| `args.moveIBInterfaces` | If true, InfiniBand (IPoIB) interfaces are moved into the pod network namespace | binary default: `true` |
+| `daemonArgs.filter` | CEL expression to filter network interface attributes | see binary default |
+| `daemonArgs.inventoryMinPollInterval` | Minimum interval between two consecutive inventory polls | binary default: `2s` |
+| `daemonArgs.inventoryMaxPollInterval` | Maximum interval between two consecutive inventory polls | binary default: `1m` |
+| `daemonArgs.inventoryPollBurst` | Number of inventory polls that can be run in a burst | binary default: `5` |
+| `daemonArgs.moveIBInterfaces` | If true, InfiniBand (IPoIB) interfaces are moved into the pod network namespace | binary default: `true` |
```
`args{}` is shorter and a bit easier to work with (especially with Helm being case-sensitive). `daemonArgs` is clearer, but I wonder if it’s more verbosity than we need right now.

michaelasp: Yeah, I don't feel too strongly on this. Let's keep it simpler since for the time being there's only one container we pass args to.

deployments/helm/dranet/README.md
```diff
@@ -24,13 +24,21 @@ The following table lists the configurable parameters and their default values:
...
 | `resources.limits.cpu` | CPU resource limit | `""` (not set) |
 | `resources.limits.memory` | Memory resource limit | `""` (not set) |
-| `readinessProbe.httpGet.path` | Readiness probe HTTP path | `/healthz` |
-| `readinessProbe.httpGet.port` | Readiness probe HTTP port | `9177` |
+| `args.filter` | CEL expression to filter network interface attributes | see binary default |
+| `args.inventoryMinPollInterval` | Minimum interval between two consecutive inventory polls | binary default: `2s` |
+| `args.inventoryMaxPollInterval` | Maximum interval between two consecutive inventory polls | binary default: `1m` |
+| `args.inventoryPollBurst` | Number of inventory polls that can be run in a burst | binary default: `5` |
+| `args.moveIBInterfaces` | If true, InfiniBand (IPoIB) interfaces are moved into the pod network namespace | binary default: `true` |
+| `args.cloudProviderHint` | Hint for the cloud provider plugin (`GCE`, `AZURE`, `OKE`, `NONE`); auto-detected if unset | binary default: auto-detect |
```
michaelasp: `binary default: auto-detect`

This is an unsupported value right? Shouldn't it just be empty? This is just the readme but we should probably have it empty?

fmuyassarov: Correct and sorry about that. I mistakenly added it as an allowed value for the provider, even though it doesn’t exist in the enum. Good catch, it is fixed now.

PR #130, opened 2026-04-06
chore: add helm chart packaging & pushing to release process

fmuyassarov:
#### What type of PR is this?
/kind feature
/kind dependency
Add Helm chart publishing steps to the release pipeline so that dranet charts are automatically packaged and pushed to the OCI registry on every tagged release.
Part of https://github.com/kubernetes-sigs/dranet/issues/128

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,14 @@ update:
 ensure-buildx:
 	./hack/init-buildx.sh
 
+HELM_VERSION_SHA?=a2369ca71c0ef633bf6e4fccd66d634eb379b371 # v3.20.1
+.PHONY: ensure-helm
+ensure-helm:
+	@if ! helm version >/dev/null 2>&1; then \
+		echo "Helm not found, installing helm@$(HELM_VERSION_SHA) ..."; \
+		go install helm.sh/helm/v3/cmd/helm@$(HELM_VERSION_SHA); \
+	fi
+
 # get image name from directory we're building
 IMAGE_NAME=dranet
 # docker image registry, default to upstream
@@ -58,6 +66,10 @@ REGISTRY?=gcr.io/k8s-staging-networking
 TAG?=$(shell echo "$$(date +v%Y%m%d)-$$(git describe --always --dirty)")
 # the full image tag
 IMAGE?=$(REGISTRY)/$(IMAGE_NAME):$(TAG)
+CHART_REGISTRY?=$(REGISTRY)/charts/$(IMAGE_NAME)
+HELM_TAG?=$(shell git describe --tags --exact-match 2>/dev/null)
+# for helm chart version strip 'v' to have valid semver (example: v0.1.0 → 0.1.0)
+CHART_VERSION=$(shell echo "$(HELM_TAG)" | sed 's/^v//')
 PLATFORMS?=linux/amd64,linux/arm64
 
 # base images (defaults are in the Dockerfile)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -50,6 +50,13 @@ update:
 ensure-buildx:
 	./hack/init-buildx.sh
 
+HELM_VERSION?=v3.20.1
+.PHONY: ensure-helm
+ensure-helm:
+	@if ! helm version >/dev/null 2>&1; then \
+		go install helm.sh/helm/v3/cmd/helm@$(HELM_VERSION); \
```
michaelasp: Can we use sha here? Just want to avoid scenarios like https://www.docker.com/blog/trivy-supply-chain-compromise-what-docker-hub-users-should-know/

I think guidance is generally in favor of sha now.

fmuyassarov: Yes, I tend to agree. Replaced the version by `a2369ca71c0ef633bf6e4fccd66d634eb379b371` based on the output as below

```
go list -m -json helm.sh/helm/v3@v3.20.1
{
	"Path": "helm.sh/helm/v3",
	"Version": "v3.20.1",
	"Time": "2026-03-11T23:22:18Z",
	"GoMod": "/Users/est/go/pkg/mod/cache/download/helm.sh/helm/v3/@v/v3.20.1.mod",
	"GoVersion": "1.25.0",
	"Origin": {
		"VCS": "git",
		"URL": "https://github.com/helm/helm",
		"Hash": "a2369ca71c0ef633bf6e4fccd66d634eb379b371",
		"Ref": "refs/tags/v3.20.1"
	}
}
```

cloudbuild.yaml
```diff
@@ -8,4 +8,5 @@ steps:
   env:
   - REGISTRY=gcr.io/k8s-staging-networking
   - IMAGE_NAME=dranet
-  args: ['release']
+  - TAG=$TAG_NAME
```
gauravkghildiyal: Non-blocking: I'm a bit uncertain if the TAG will work out correctly (and I also recognize that the current behaviour is also not correct where we don't correctly tag images). Something we'll need to verify separately from this issue whether `TAG=$TAG_NAME` was sufficient here to fix things.

cloudbuild.yaml
```diff
@@ -8,4 +8,5 @@ steps:
...
+  args: ['ensure-helm', 'release']
```
gauravkghildiyal: Why do we have `ensure-helm` separately rather than making it as a dependency of one of the Make targets (similar to `ensure-buildx`)?

gauravkghildiyal (commented):
Thanks @fmuyassarov.

cloudbuild.yaml
fmuyassarov: I just realized that this variable is not needed at all. Because, we just execute the `release` target and the tag is [determined](https://github.com/Nordix/dranet/blob/88a0a4f5a3c5020eab06d48f19764132acdacbdb/Makefile#L69) within the Makefile anyways. Did a quick local test as well:

```sh
$ git tag -a v1.1.1 -m "v1.1.1"
$ git checkout v1.1.1 
$ make helm-package
$ helm package deployments/helm/dranet \
		--version "1.1.1" \
		--app-version "v1.1.1" \
		--destination /Users/est/dranet/bin
Successfully packaged chart and saved it to: /Users/est/dranet/bin/dranet-1.1.1.tgz

$ tar -xvf dranet-1.1.1.tgz
```

```sh
$ cat dranet/Chart.yaml

apiVersion: v2
appVersion: v1.1.1
description: Helm chart for deploying DRANET, a Kubernetes network driver using Dynamic
  Resource Allocation (DRA)
home: https://github.com/kubernetes-sigs/dranet
keywords:
- networking
- dra
- dynamic-resource-allocation
maintainers:
- name: aojea
- name: gauravkghildiyal
- name: michaelasp
- name: fmuyassarov
name: dranet
sources:
- https://github.com/kubernetes-sigs/dranet
type: application
version: 1.1.11.1
```

cloudbuild.yaml
fmuyassarov: Good point. I had it initially as dependency to `helm-package` but then I realized that it is a bad idea because people may run the target locally which will force them to execute the helm installation. But I've now added it as part of `release` target.

aojea:
@fmuyassarov this is failing https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-dranet-image/2042591551317610496

fmuyassarov:
Ah sorry. Looking into it

Issue #111, opened 2026-03-24
Add OCI Support


dkennetzoracle:
Hi all,

OCI here - would love to work with you all to support OCI (OKE) with dranet. Would love to see what changes would be required to build that integration.

Thanks,
Dennis

gauravkghildiyal:
Awesome to see the interest. Assigned the issue to you. Feel free to comment here as you test things out and need to discuss something.

dkennetzoracle:
Awesome - thanks!

gauravkghildiyal:
Thanks for the PR @dkennetzoracle. 
I'm keeping this issue open for now to track adding an example.

PR #138, opened 2026-04-08
Feat: Add real topology data and Working ipv6.

dkennetzoracle:
#### What type of PR is this?

/kind feature

Joint PR (happy to split up, just wanted to get eyes on it and work with you all to do so).

1. **OKE IMDS topology data** - (HPC island, network block, local block, rack, GPU memory fabric) is exposed as device attributes for topology-aware scheduling.
2. **OKE Example GB200** - adds an example doc for GB200 with performance numbers and importance of aligned NIC selection
3. **OCI's RA daemon injects IPv6 default routes** - onto every RDMA NIC, causing them to be misclassified as uplink interfaces and filtered from discovery. A new `NonUplinkChecker` interface allows the OKE provider to override this for RDMA devices on GPU fabric shapes.
4. **Pod namespaces have IPv6 disabled** (`net.ipv6.conf.all.disable_ipv6=1`), but RoCEv2 on OKE requires a globally-routable IPv6 address on RDMA interfaces to populate a routable GID (GID index >= 2). A new `EnableIPv6` field on `InterfaceConfig` triggers per-interface IPv6 enablement in the pod namespace via sysctl override, followed by re-application of the IPv6 address.

#111 - doesn't close

**Note**: Again, happy to split this up, but did want to get it out to start the process if needed. dranet on OKE won't work for all GPU SKUs without this.

##### Why OKE needs these changes but AKS and GCE do not

The IPv6 and uplink-filtering changes are driven by OKE's bare-metal
infrastructure model, which differs fundamentally from how AKS and GCE expose
RDMA networking:

**AKS (Azure):** GB300 nodes use InfiniBand VFs, not RoCEv2. IB GIDs are derived
from the port GUID (a hardware identifier burned into the HCA), not from IP
addresses. There is no dependency on IPv6 addressing for GID generation --
`ibv_modify_qp` resolves paths using GUID-based GIDs and the IB subnet manager.
Because IB VFs have no Ethernet netdev, there is no RA daemon, no IPv6 address,
and no default-route side-effect to filter around. The dranet IB-only path
(char-device injection without netdev movement) handles this cleanly.

**GCE (Google Cloud):** GPU nodes use SR-IOV VFs for RDMA. The VFs are presented
as standard Ethernet NICs with IPv4 addresses managed by the GCE metadata
service. There is no RA daemon injecting IPv6 routes, and GIDs (when RoCEv2 is
used) are populated from the IPv4 address. The pod namespace does not need IPv6
enabled because the GID is IPv4-derived.

**OKE (Oracle Cloud) -- why it is different:** BM.GPU.GB200-v3.4 are bare-metal
shapes. The 8 ConnectX-8 NICs are full physical functions (PFs), not SR-IOV VFs.
OCI manages the RDMA fabric at the infrastructure level by running an RA
(Router Advertisement) daemon on the host that:

1. Assigns a globally-routable IPv6 address to each RDMA NIC (e.g.
   `fdcd:0:2a29:3032:364e:3a64:6758:428a/64`). This address populates a RoCEv2
   GID at index >= 2 in the NIC's GID table. The OCI fabric routes traffic using
   these IPv6-derived GIDs -- link-local GIDs (`fe80::`) are not routable
   (`ENETUNREACH` on `ibv_modify_qp`).

2. Injects an IPv6 default route via RA (`proto ra`) into the main routing table
   for every RDMA NIC. This is an infrastructure side-effect, not a signal that
   the NIC is an uplink/transit interface. Without the `NonUplinkChecker`
   override, dranet's gateway-interface filter removes these NICs from the
   ResourceSlice entirely.

The IPv6 problem compounds because Kubernetes single-stack IPv4 clusters (the
default on OKE) configure `net.ipv6.conf.all.disable_ipv6=1` in pod network
namespaces. When dranet moves the RDMA NIC into the pod namespace and attempts
to apply the RA-assigned IPv6 address, the kernel returns EACCES. Without the
address, the NIC has no routable GID and NCCL falls back to TCP or fails with
ENETUNREACH.

The `EnableIPv6` mechanism solves this with a per-interface sysctl override
(`net.ipv6.conf.<ifname>.disable_ipv6=0`) that enables IPv6 on just the RDMA
NIC without affecting the pod's primary network interface. This is safe because
the RDMA NIC is isolated in the pod namespace and the IPv6 address is only used
for GID generation, not for IP-level routing.

None of this applies to AKS (IB, GUID-based GIDs, no netdev) or GCE (VFs,
IPv4-based GIDs, no RA daemon).

##### Architecture context

On GB200, GPUs connect to the Grace CPU via NVLink C2C while NICs connect via
PCIe. `nvidia-smi topo -m` reports SYS for all GPU-NIC pairs. Despite this,
NCCL enables GDR for NUMA-local NICs via `NCCL_NET_GDR_C2C=1`. The meaningful
topology distinction is same-NUMA vs cross-NUMA, not PCIe host bridge alignment.

##### IPv6 flow

The `EnableIPv6` mechanism is intentionally outside `nsAttachNetdev` to keep
that function general-purpose. The sequence is:

1. `nsAttachNetdev` moves NIC to pod namespace, soft-fails IPv6 address (EACCES)
2. `attachNetdevToNS` enables IPv6 per-interface sysctl, re-applies IPv6 address
3. RDMA device generates routable GID from the applied IPv6 address

##### Known issue -- NIC orphaning - #137 

RDMA NICs are not reliably returned to the host namespace on pod deletion. This
is a pre-existing CRI-O / dranet bug (not introduced here). The example README documents
the PCI rebind recovery procedure.

##### Benchmark results

2-node `all_reduce_perf`, 1 GPU/worker, BM.GPU.GB200-v3.4:

| Template | NIC(s) | NUMA | Channels | GDR | Avg busbw |
|---|---|---|---|---|---|
| `1nic-aligned` | rdma3 (NUMA 0) | same | 4 | yes | ~46 GB/s |
| `2nic-aligned` | rdma2+rdma3 (NUMA 0) | same | 8 | yes | ~96 GB/s |
| `1nic-unaligned` | rdma4 (NUMA 1) | cross | 2 | no | ~25 GB/s |

```release-note
Add OKE cloud provider support for BM.GPU.GB200 RoCEv2 RDMA networking. dranet
now correctly discovers RDMA NICs on OKE GPU shapes (despite RA-injected default
routes), enables per-interface IPv6 in pod namespaces for routable RoCEv2 GID
generation, and exposes OCI RDMA topology attributes (HPC island, network block,
local block, rack, GPU memory fabric) for topology-aware scheduling.
```

### DIFF
--- a/pkg/cloudprovider/cloud.go
+++ b/pkg/cloudprovider/cloud.go
@@ -27,6 +27,18 @@ type DeviceIdentifiers struct {
 	MAC        string
 	PCIAddress string
 	Name       string
+	RDMA       bool
+}
+
+// NonUplinkChecker is an optional extension to CloudInstance. A provider may
+// implement this to exempt specific device classes from the default-gateway
+// uplink filter, even when those interfaces carry a default route. This is
+// needed on platforms where infrastructure daemons (e.g. RA) inject default
+// routes onto workload RDMA NICs as a side-effect of their setup.
+type NonUplinkChecker interface {
+	// IsNonUplink returns true if the device should be included in the
+	// ResourceSlice regardless of any default gateway route on that interface.
+	IsNonUplink(id DeviceIdentifiers) bool
 }
 
 // CloudInstance defines the generic interface for all cloud providers.
--- a/pkg/cloudprovider/oke/oke.go
+++ b/pkg/cloudprovider/oke/oke.go
@@ -19,9 +19,11 @@ package oke
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -35,54 +37,123 @@ import (
 const (
 	OKEAttrPrefix = "oke.dra.net"
 
-	AttrOKEShape              = OKEAttrPrefix + "/" + "shape"
-	AttrOKEFaultDomain        = OKEAttrPrefix + "/" + "faultDomain"
-	AttrOKEAvailabilityDomain = OKEAttrPrefix + "/" + "availabilityDomain"
+	AttrOKEHPCIslandId     = OKEAttrPrefix + "/" + "hpcIslandId"
+	AttrOKENetworkBlockId  = OKEAttrPrefix + "/" + "networkBlockId"
+	AttrOKELocalBlockId    = OKEAttrPrefix + "/" + "localBlockId"
+	AttrOKERackId          = OKEAttrPrefix + "/" + "rackId"
+	AttrOKEGpuMemoryFabric = OKEAttrPrefix + "/" + "gpuMemoryFabricId"
 
 	// imdsEndpoint is the Oracle Cloud Instance Metadata Service endpoint.
 	imdsEndpoint = "http://169.254.169.254/opc/v2"
 )
 
-// imdsInstanceMetadata contains the fields we care about from the OCI IMDS
-// instance metadata response.
-type imdsInstanceMetadata struct {
-	Shape              string `json:"shape"`
-	FaultDomain        string `json:"faultDomain"`
-	AvailabilityDomain string `json:"availabilityDomain"`
+// imdsHostRDMATopologyData contains the RDMA topology fields from the OCI
+// IMDS host metadata response. This is only populated when RDMA topology
+// data is enabled for the tenancy.
+type imdsHostRDMATopologyData struct {
+	CustomerGpuMemoryFabric string `json:"customerGpuMemoryFabric"`
+	CustomerHPCIslandId     string `json:"customerHPCIslandId"`
+	CustomerHostId          string `json:"customerHostId"`
+	CustomerLocalBlock      string `json:"customerLocalBlock"`
+	CustomerNetworkBlock    string `json:"customerNetworkBlock"`
+}
+
+// imdsHostMetadata contains the fields we care about from the OCI IMDS
+// host metadata response at /opc/v2/host/.
+type imdsHostMetadata struct {
+	NetworkBlockId   string                    `json:"networkBlockId"`
+	RackId           string                    `json:"rackId"`
+	RDMATopologyData *imdsHostRDMATopologyData `json:"rdmaTopologyData"`
 }
 
 var _ cloudprovider.CloudInstance = (*OKEInstance)(nil)
+var _ cloudprovider.NonUplinkChecker = (*OKEInstance)(nil)
 
-// OKEInstance holds OCI/OKE specific instance data.
+// OKEInstance holds OCI/OKE specific instance topology data.
 type OKEInstance struct {
-	Shape              string
-	FaultDomain        string
-	AvailabilityDomain string
+	HPCIslandId    string
+	NetworkBlockId string
+	LocalBlockId   string
+	RackId         string
+	// GpuMemoryFabric is only populated on shapes that use a GPU memory fabric
+	// interconnect (e.g. BM.GPU.GB200, BM.GPU.GB300). It will be empty on all
+	// other shapes such as BM.GPU.H100.8.
+	GpuMemoryFabric string
 }
 
-// GetDeviceAttributes returns OKE-specific attributes for a device.
-// These are node-level attributes applied to all devices since OCI IMDS
-// does not expose per-RDMA-NIC metadata.
+// GetDeviceAttributes returns OKE-specific topology attributes for a device.
+// These are node-level attributes applied to all devices since the OCI IMDS
+// host endpoint exposes per-node topology, not per-NIC metadata.
 func (o *OKEInstance) GetDeviceAttributes(id cloudprovider.DeviceIdentifiers) map[resourceapi.QualifiedName]resourceapi.DeviceAttribute {
 	attributes := make(map[resourceapi.QualifiedName]resourceapi.DeviceAttribute)
 
-	if o.Shape != "" {
-		attributes[AttrOKEShape] = resourceapi.DeviceAttribute{StringValue: &o.Shape}
+	if o.HPCIslandId != "" {
+		attributes[AttrOKEHPCIslandId] = resourceapi.DeviceAttribute{StringValue: &o.HPCIslandId}
 	}
-	if o.FaultDomain != "" {
-		attributes[AttrOKEFaultDomain] = resourceapi.DeviceAttribute{StringValue: &o.FaultDomain}
+	if o.NetworkBlockId != "" {
+		attributes[AttrOKENetworkBlockId] = resourceapi.DeviceAttribute{StringValue: &o.NetworkBlockId}
 	}
-	if o.AvailabilityDomain != "" {
-		attributes[AttrOKEAvailabilityDomain] = resourceapi.DeviceAttribute{StringValue: &o.AvailabilityDomain}
+	if o.LocalBlockId != "" {
+		attributes[AttrOKELocalBlockId] = resourceapi.DeviceAttribute{StringValue: &o.LocalBlockId}
+	}
+	if o.RackId != "" {
+		attributes[AttrOKERackId] = resourceapi.DeviceAttribute{StringValue: &o.RackId}
+	}
+	if o.GpuMemoryFabric != "" {
+		attributes[AttrOKEGpuMemoryFabric] = resourceapi.DeviceAttribute{StringValue: &o.GpuMemoryFabric}
 	}
 
 	return attributes
 }
 
-// GetDeviceConfig returns nil as OCI does not provide device-specific
-// network configuration through IMDS.
+// ocidSuffix returns the unique identifier suffix of an OCI OCID — the segment
+// after the last '.'. DRA string attributes are capped at 64 bytes, but full
+// OCIDs are ~90+ characters; the suffix is always 60 characters and is unique
+// per resource within a tenancy, making it safe to use as an attribute value.
+func ocidSuffix(s string) (string, error) {
+	if s == "" {
+		return "", nil
+	}
+	if !strings.Contains(s, "ocid") {
+		return "", fmt.Errorf("not a valid OCID (missing 'ocid' prefix): %q", s)
+	}
+	i := strings.LastIndex(s, ".")
+	if i < 0 {
+		return "", fmt.Errorf("not a valid OCID (missing '.' separator): %q", s)
+	}
+	suffix := s[i+1:]
+	if len(suffix) > 60 {
+		suffix = suffix[len(suffix)-60:]
+	}
+	return suffix, nil
+}
+
+// GetDeviceConfig returns OKE-specific per-device network configuration.
+// On OKE GPU RDMA shapes (BM.GPU.GB200, BM.GPU.GB300, …), RDMA NICs receive a
+// globally-routable IPv6 address via Router Advertisement. That address is required
+// to populate a routable RoCEv2 GID (GID index ≥ 2) in the pod namespace.
+// EnableIPv6 signals dranet to override the pod-level disable_ipv6=1 sysctl for
+// this specific interface so the RA-assigned global address can be applied.
 func (o *OKEInstance) GetDeviceConfig(id cloudprovider.DeviceIdentifiers) *apis.NetworkConfig {
+	if o.GpuMemoryFabric != "" && id.RDMA {
+		enableIPv6 := true
+		return &apis.NetworkConfig{
+			Interface: apis.InterfaceConfig{
+				EnableIPv6: &enableIPv6,
+			},
+		}
+	}
 	return nil
+}
+
+// IsNonUplink returns true for RDMA-capable devices on OKE GPU RDMA shapes
+// (BM.GPU.GB200, BM.GPU.GB300, …). On these shapes OCI's Router Advertisement
+// daemon injects an IPv6 default route into the main routing table for every
+// RDMA interface as an infrastructure side-effect. Those interfaces are
+// workload NICs, not transit uplinks — without this override the default
+// gateway filter would remove them from the ResourceSlice entirely.
+func (o *OKEInstance) IsNonUplink(id cloudprovider.DeviceIdentifiers) bool {
+	return o.GpuMemoryFabric != "" && id.RDMA
 }
 
 // OnOKE returns true if running on an Oracle Cloud Infrastructure instance.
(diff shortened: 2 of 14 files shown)


dkennetzoracle:
@gauravkghildiyal thanks for clarifying that GCP runs BMs, I didn't know this! OKE does support dual-stack, the cluster I was testing on was just running single stack. I've reached out internally to see if we could get another one setup with dual stack which may alleviate a lot of this.

todo.md
anson627, line 1: rm? since it looks like personal notes

pkg/cloudprovider/oke/oke_test.go
```diff
@@ -100,11 +118,38 @@ func TestGetDeviceAttributes(t *testing.T) {
...
+		})
+	}
+}
+
 func TestGetDeviceConfig(t *testing.T) {
 	instance := &OKEInstance{
-		Shape:              "BM.GPU.H100.8",
-		FaultDomain:        "FAULT-DOMAIN-1",
-		AvailabilityDomain: "TrcQ:US-ASHBURN-AD-2",
+		HPCIslandId:    "fake-island-id",
```
anson627: add test case to cover new code path where GpuMemoryFabric != "" and id.RDMA == true should return a config with EnableIPv6: true

pkg/cloudprovider/oke/oke.go
```diff
@@ -35,56 +36,115 @@ import (
 const (
 	OKEAttrPrefix = "oke.dra.net"
 
-	AttrOKEShape              = OKEAttrPrefix + "/" + "shape"
```
anson627, line 38: removing these fields could cause existing usage to break, right?

pkg/cloudprovider/oke/oke.go
```diff
@@ -106,44 +166,55 @@ func OnOKE(ctx context.Context) bool {
...
-			klog.Infof("could not reach OCI IMDS ... retrying: %v", err)
+			klog.Infof("could not reach OCI IMDS host endpoint ... retrying: %v", err)
 			return false, nil
 		}
 		defer resp.Body.Close()
 
 		if resp.StatusCode != http.StatusOK {
-			klog.Infof("OCI IMDS returned status %d ... retrying", resp.StatusCode)
-			return false, nil
+			return false, fmt.Errorf("Please turn on TopologyData for your Tenancy")
```
anson627: would there be any other error can be retried, e.g. network error?

pkg/cloudprovider/oke/oke.go
```diff
@@ -35,56 +36,115 @@ import (
...
 }
 
-// GetDeviceConfig returns nil as OCI does not provide device-specific
-// network configuration through IMDS.
+// ocidSuffix returns the unique identifier suffix of an OCI OCID — the segment
+// after the last '.'. DRA string attributes are capped at 64 bytes, but full
+// OCIDs are ~90+ characters; the suffix is always 60 characters and is unique
+// per resource within a tenancy, making it safe to use as an attribute value.
+// Non-OCID strings (e.g. the rackId hex hash) are returned unchanged.
+func ocidSuffix(s string) string {
```
anson627: nit: add OCID validation: e.g. 60 characters length, string contains '.'?

dkennetzoracle:
Thanks for the review @anson627 - I am addressing some of the feedback now!

pkg/cloudprovider/oke/oke.go
dkennetzoracle: They will, but they were placeholders. The API call made there doesn't return anything useful for topology awareness and is a separate api call then the topology data one. To maintain both, I'd need to call out to 2 APIs, which is fine and still provides some value because it gives us the PCIe of the NICs purely from dranet but it doesn't provide any value beyond that (for topology). My preference for someone on OCI to fail loudly if topology awareness is not enabled, rather than just logging it and moving on, which is why I removed the old stuff.

This is still very new on the OCI side, so I don't expect to have any users for our cloud yet so I'm not concerned with the backwards compatibility aspect.

pkg/cloudprovider/oke/oke.go
dkennetzoracle: Adding the empty string return because GpuMemoryFabric is only present for rack-scale shapes like Grace Blackwell. Basically, if GpuMemoryFabric exists, it should proceed through the validation. If not, it should just get skipped.

pkg/cloudprovider/oke/oke_test.go
dkennetzoracle: Done!

pkg/cloudprovider/oke/oke.go
dkennetzoracle: I modified this to retry rather than return an error. If the context deadline is exceeded, it'll error.

pkg/cloudprovider/oke/oke.go
anson627: make sense, thanks for the clarification!

anson627 (approved):

pkg/driver/dra_hooks.go
```diff
@@ -563,11 +563,18 @@ func getRouteInfo(nlHandle nlwrap.Handle, ifName string, link netlink.Link) ([]a
...
-			// Discard IPv6 proto=kernel routes
+			// Discard IPv6 proto=kernel routes (auto-generated connected routes).
 			if route.Protocol == unix.RTPROT_KERNEL {
 				klog.V(5).Infof("Skipping IPv6 proto=kernel route %s for interface %s", route.String(), ifName)
 				continue
 			}
+			// Discard IPv6 RA-assigned routes. These are injected by the cloud
+			// provider's RA daemon as an infrastructure side-effect (e.g. OCI on
+			// RDMA NICs) and must not be propagated into pod namespaces.
+			if route.Protocol == unix.RTPROT_RA {
```
aojea, line 574: I'm not sure we can generalize this pattern

pkg/cloudprovider/cloud.go
```diff
@@ -27,6 +27,18 @@ type DeviceIdentifiers struct {
...
+
+// NonUplinkChecker is an optional extension to CloudInstance. A provider may
+// implement this to exempt specific device classes from the default-gateway
+// uplink filter, even when those interfaces carry a default route. This is
+// needed on platforms where infrastructure daemons (e.g. RA) inject default
+// routes onto workload RDMA NICs as a side-effect of their setup.
+type NonUplinkChecker interface {
+	// IsNonUplink returns true if the device should be included in the
+	// ResourceSlice regardless of any default gateway route on that interface.
+	IsNonUplink(id DeviceIdentifiers) bool
```
aojea, line 41: we have a flag for providers to implement custom filtering based on attributes, can't we use that?

This model of special casting by creating interfaces does not look very sustainable long term

pkg/driver/hostdevice.go
```diff
@@ -150,6 +150,13 @@ func nsAttachNetdev(hostIfName string, containerNsPAth string, interfaceConfig a
 		}
 		err = nhNs.AddrAdd(nsLink, &netlink.Addr{IPNet: &net.IPNet{IP: ip, Mask: ipnet.Mask}})
 		if err != nil {
+			// IPv6 may be disabled in the pod namespace (single-stack IPv4 clusters set
+			// net.ipv6.conf.all.disable_ipv6=1). Soft-fail so that the caller can enable
+			// IPv6 per-interface and re-apply the address after moving the device.
+			if ip.To4() == nil && (errors.Is(err, unix.EACCES) || errors.Is(err, unix.EPERM)) {
+				klog.V(4).Infof("skipping IPv6 address %s on %s in namespace %s: IPv6 is disabled (will retry after per-interface enable)", address, nsLink.Attrs().Name, containerNsPAth)
```
aojea, line 157: I can't remember right now the exactl logic, where is this retried?

pkg/driver/nri_hooks.go
```diff
@@ -275,6 +275,20 @@ func attachNetdevToNS(pod *api.PodSandbox, ns, deviceName string, config DeviceC
...
+	if config.NetworkInterfaceConfigInPod.Interface.EnableIPv6 != nil &&
+		*config.NetworkInterfaceConfigInPod.Interface.EnableIPv6 {
+		if err := enableIPv6ForInterface(ns, ifNameInNs); err != nil {
+			return fmt.Errorf("failed to enable IPv6 for %s in namespace %s: %w", ifNameInNs, ns, err)
+		}
+		if err := reapplyIPv6Addresses(ns, ifNameInNs, config.NetworkInterfaceConfigInPod.Interface.Addresses); err != nil {
+			return fmt.Errorf("failed to re-apply IPv6 addresses for %s in namespace %s: %w", ifNameInNs, ns, err)
+		}
+	}
+
```
aojea, line 291: oh, I see, https://github.com/kubernetes-sigs/dranet/pull/138/changes#r3063956282

can't we do better and set the sysctl before the addresses?

pkg/inventory/db.go
```diff
@@ -229,10 +229,28 @@ func (db *DB) scan() []resourceapi.Device {
...
+				}
+				if rdmaAttr, ok := device.Attributes[apis.AttrRDMA]; ok && rdmaAttr.BoolValue != nil {
+					id.RDMA = *rdmaAttr.BoolValue
+				}
+				if checker.IsNonUplink(id) {
+					klog.V(4).Infof("Interface %s has a default gateway route but is classified as a non-uplink by the cloud provider; including in discovery", *ifName)
+					filteredDevices = append(filteredDevices, device)
+					continue
+				}
+			}
```
aojea, line 253: this is the thing I prefer to think more about it...  the gwInterfaces was added based on an assumption that it seems to no longer be true, can we make the `gwInterfaces` detection logic more resilient? I want us to think in a better architecture that is sustainable long term rather in just solving the problem at hand

aojea:
/hold

I'm +1 on the PR but I want us to think more holistically, holding to avoid an unexpected merge?

dkennetzoracle:
@aojea thanks for the review! To your point:

> I'm +1 on the PR but I want us to think more holistically, holding to avoid an unexpected merge?

I agree with this and many of the things added in the PR feel like patches rather than design decisions. I just can't test on other systems outside of OCI, so it is hard for me to abstract some of this stuff. I'll spend some time reviewing and provide some feedback for what could be a bit more robust.

pkg/cloudprovider/cloud.go
dkennetzoracle: A challenge in this case with using the `FilterDevices` was that the CEL filter runs after the gwInterfaces exclusion so by the time `FilterDevices` sees the devices, the RDMA NICs with default RA routes have already been removed so the filter can't include them back..

Should probably relate to your comment about gwInterfaces and improving that.

pkg/driver/nri_hooks.go
dkennetzoracle: Yeah, good point. We should be able to cut out the reapply by just doing:

```
enableIPv6ForInterface()
nsAttachNetdev()
```

dkennetzoracle: Oh but that has some problems. Currently `enableIPv6ForInterface` operates inside the pod ns on the interface after it's been moved so we'd need to restructure so the sysctl is set on the interface name inside the pod netns before address application.

nsAttachNetdev does move + address in one call, so we'd either need to:
- split nsAttachNetDev into move + configure phases
- have it accept a pre-configure callback
- enable IPV6 per-instance inside nsAttachNetDev when EnableIPv6 is set before applying addresses

The last one is probably easiest.

anson627:
> @aojea thanks for the review! To your point:
> 
> > I'm +1 on the PR but I want us to think more holistically, holding to avoid an unexpected merge?
> 
> I agree with this and many of the things added in the PR feel like patches rather than design decisions. I just can't test on other systems outside of OCI, so it is hard for me to abstract some of this stuff. I'll spend some time reviewing and provide some feedback for what could be a bit more robust.

maybe we split OCI specific part out into a separate PR, which is relatively straightforward and can merge quickly, and spend some more time on refactoring & testing changes on the common code path

pkg/driver/nri_hooks.go
anson627: agree handling EnableIPv6 inside nsAttachNetdev directly — since it already does move + address in one call, this should avoid the soft-fail + reapply dance entirely

dkennetzoracle:
@aojea - aligned! I thought that might be the outcome of the PR. Are you okay with me including the OCI + example in this PR even though example isn't reproducible until we get ipv6 stuff in? Or should I put the example in the ipv6 PR?

aojea:
@dkennetzoracle we are planning to do a release soon? is this ipv6 something you will want as part of the release?

dkennetzoracle:
@aojea - possibly, how long are the release cycles generally? If I missed this one, how long until the next one?

aojea:
> @aojea - possibly, how long are the release cycles generally? If I missed this one, how long until the next one?

right now releases are ad hoc, in this case the only friction is about the mltiple gateway UX, the `EnableIPv6` sounds good, is clearly an `interface` option and we have other `sysctl` interface options ... I think we can improve the implementation to be more streamlined and not rely on retries

The multiple gateway thing I prefer we put some more thought on it, the original idea is that we do make it sane by default, so no customer breaks their nodes by putting the vm interface on the pod ... but with the situation you are describing is unclear to me if we can be smart about it ... I feel like we should be able to

tamilmani1989:
Thanks for raising this PR. We are hitting a very similar issue on Azure dualstack clusters with multiple NICs (e.g. eth0, eth1). Both interfaces carry a default IPv6 gateway route, so dranet default interface filter incorrectly excludes both from discovery. In practice, only one of those interfaces is the VM's default NIC; the other should be allocatable to pods.

Could we reuse this `NonUplinkChecker` mechanism in the Azure cloud provider to solve this case as well? In our scenario both are physical uplink interfaces, but only one is actually used by the VM for management traffic -- the other NIC is intended for pod/container workloads.

One naming thought: since this isn't strictly about `"uplink vs non-uplink"` but rather about identifying the primary/default interface that should be excluded, would something like `IsDefaultInterface` better capture the intent across providers? On OKE the distinction is "RDMA NIC vs management NIC"; on Azure it's "primary NIC vs secondary NIC"

aojea:
> would something like `IsDefaultInterface` better capture the intent across providers?

if we all can agree that exist a "default" interface and that will always be skipped that is great ... now I wonder, is there an automated way to unequivocally identify it across providers without having to call the provider specific? I want to avoid we start to "overuse" the interfaces if possible, in this case I have the feeling we should be able to do it, at the end of the day all machines need to have a default interface

pkg/inventory/db.go
tamilmani1989: should we consider defining a node-level CRD that explicitly declares which devices dranet is responsible for managing?

Today, dranet discovers and publishes all PCI/network devices on the node into ResourceSlices, even those it doesn't actually manage. This becomes a problem when running dranet alongside other DRA drivers that manage their own subset of devices. Both drivers end up publishing overlapping ResourceSlices, leading to conflicts.

Rather than adding per-provider api like NonUplinkChecker to decide which interfaces to include or exclude, a node CRD could serve as the explicit source of truth: "these are the devices dranet should discover and manage on this node." This would cleanly solve the uplink filtering problem (the management NIC simply wouldn't be listed), the multi-driver coexistence problem, and make the behavior transparent and configurable per node.

Cilium DRA proposal also takes a similar approach - using a node-scoped CRD to define which devices the DRA driver should manage.

dkennetzoracle:
I think the word Interface is a bit convoluted here @aojea and @tamilmani1989 - I generally avoid using it for things which are not interfaces in the programmatic sense. This is an accurate statement for us on OKE:

> On OKE the distinction is "RDMA NIC vs management NIC";

I'm okay with another name, and I do think there should be an automated way (or a provider specific way) to provide the default interface since the detection mechanism may likely change across providers. I'll dedicate some time to this tomorrow, and can come back with some ideas.

gauravkghildiyal:
> if we all can agree that exist a "default" interface and that will always be skipped that is great

I'm wondering whether DRANET identifying "multiple" interfaces as the "default" is a signal that our detection of the "default" is not reliable and that we should use a different strategy (= cloud provider specific)?

@dkennetzoracle , @tamilmani1989 -- can you both share some definition/identification of what "default" interface means for you? Maybe let's start with finding something common and see if that's generic enough?

pkg/inventory/db.go
gauravkghildiyal: > Today, dranet discovers and publishes all PCI/network devices on the node into ResourceSlices, even those it doesn't actually manage. 

This is something that I've been thinking about lately as well.

I'm wondering if a good step in that direction is to allow cloudprovider to dynamically filter devices based on some discovered property (rather than just being able to add attributes to a ResourceSlice). Basically something more generic which allows more properties of the ResourceSlice to be modifiable by CloudProvider logic and filtering. We eventually will need it when we want to support things like DRA Device Taints and Tolerations https://github.com/kubernetes/enhancements/issues/5055 which cloud providers may want to configure.

Defining a new CRD is something I want to avoid.

tamilmani1989:
> I'm wondering whether DRANET identifying "multiple" interfaces as the "default" is a signal that our detection of the "default" is not reliable and that we should use a different strategy (= cloud provider specific)?

Correct, this works for single-stack IPv4 clusters. However, in dual-stack clusters, it incorrectly identifies both NICs as default gateway interfaces. This happens because dranet detects default interfaces based on default gateway routes, and in the IPv6 case, there are two default IPv6 gateway routes - one per NIC causing both to be classified as default interfaces.

ip -6 route:
>default via <gw_ipv6> dev eth0 metric 100 
default via <gw_ipv6> dev eth1 metric 200

aojea:
I think we do not need to overcomplicate, the detect gw logic was very simplistic and IMHO it is a bug, @dkennetzoracle @tamilmani1989 can you please check if this patch https://github.com/kubernetes-sigs/dranet/pull/153 solves your problem and review it?

dkennetzoracle:
@aojea I'm not sure it solves all my problems, but it helps. It doesn't touch the "ipv6" enablement part of the PR, but it may solve the NonUplinkChecker part for us but I don't think so.

Basically, the RA daemon installs a `proto ra` ipv6 default route on each of the 8 NICs. RA-learned defaults use the kernel's RA priority and there is no mechanism for differentiating the 8 RDMA NICs from each other by metric or prio then. So in #153 , all 8 RDMA NICs tied at the RA metric are all winners, so they are still excluded.

My scope that still remains:

- Per-interface IPv6 enablement in pod namespaces (`net.ipv6.conf.<ifname>.disable_ipv6=0`) - orthogonal to the gateway detector.
- Re-application of the RA-assigned IPv6 address after the NIC crosses into the pod netns which is the `EnableIPv6` flow on `InterfaceConfig`

Issue #86, opened 2026-03-03
Add AWS support


aojea:
AWS uses EFA and (I think) ENA devices

https://docs.aws.amazon.com/eks/latest/userguide/node-efa.html

nakshah87:
Hi @aojea @gauravkghildiyal. What is the release cadence you have to cut a new release?

nakshah87:
Note: Request to cut a release tag by 04/16

PR #139, opened 2026-04-09
Changes to support EFA enabled AWS instances on dranet

nakshah87:
#### What type of PR is this?
Added AWS cloud provider support to enable EFA (Elastic Fabric Adapter) device discovery and configuration for AWS EC2 instances with NVIDIA GPU and AWS Neuron accelerators that support EFA devices.

/kind feature

The implementation:
* Detects AWS EC2 instance types via IMDS, 
* Identifies Neuron instances (Trainium/Inferentia) by instance type prefix
* Leverages the [connected-device-maps-over-efa-for-neuron](https://github.com/aws-neuron/connected-device-maps-over-efa-for-neuron) library to retrieve device group attributes, enabling closely connected device allocation for EFA and Neuron devices.

The PR includes extensive Unit Tests and integration tests have been run to ensure e2e pass.

For the calls to IMDS:
* The aws sdk call is configured with maxRetries set to 10 and maxTimeout set to 10secs
* In OnAws, the overall timeout is set to 5 secs.
* In GetInstance, the overall timeout is set to 15 secs.

In OnAws and GetInstance the timeout comes into picture only when IMDS accepts the TCP connection but never responds which is very rare.
In most cases the attempt resolves fast
- IMDS is up: responds in <100ms. No retry needed.                                                                                                       
- IMDS is down (connection refused): TCP RST comes back in <1ms. Attempt fails instantly, SDK moves to next retry with backoff wait.                     
- IMDS is unreachable (no route): OS returns error in <1ms. Same — instant failure, backoff, next retry.
                                                                                                                                                           
So the 10s HTTP timeout is a per-attempt ceiling, not the per-attempt duration.

[#86](https://github.com/kubernetes-sigs/dranet/issues/86)

NONE


pkg/cloudprovider/aws/aws.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// getEFADeviceGroupIDs is a variable so tests can override it.
+var getEFADeviceGroupIDs = lookup.GetEFADeviceGroupIDs
+
+// GetDeviceAttributes fetches all attributes related to the provided device.
+func (a *AWSInstance) GetDeviceAttributes(id cloudprovider.DeviceIdentifiers) map[resourceapi.QualifiedName]resourceapi.DeviceAttribute {
+	attributes := make(map[resourceapi.QualifiedName]resourceapi.DeviceAttribute)
+
+	if a.IsNeuronInstance && isEFADevice(id.PCIAddress) {
```
anson627, line 87: is it meant to add general EFA support (e.g. p4d/p5 instance type) on AWS? this implementation only covers the Neuron-specific subset

nakshah87: It covers both instances with Neuron-EFA (eg: trn and inf) and also instances with NVIDIA-EFA (eg: p5). The additional attributes are added for Neuron only.
The existing attributes suffice for the NVIDIA instances.

pkg/inventory/cloud.go
```diff
@@ -79,11 +83,6 @@ func getInstanceProperties(ctx context.Context, cloudProviderHint CloudProviderH
 		return nil
 	}
 	instance, err := provider(ctx)
-	if err != nil {
```
aojea, line 82: oh, this was duplicate

pkg/inventory/cloud.go
```diff
@@ -36,6 +37,7 @@ type CloudProviderHint string
 
 const (
 	CloudProviderHintGCE   CloudProviderHint = "GCE"
+	CloudProviderHintAWS   CloudProviderHint = "AWS"
```
michaelasp, line 40: This comes from a command line flag, can you update the help comment there to add AWS.

pkg/inventory/cloud.go
michaelasp: Oops, that was probably me 😅, thanks for the cleanup.

pkg/inventory/cloud.go
nakshah87: Done.

aojea:
/lgtm
/approve

without knowing the specific details of this provider, the code structures looks ok

pkg/cloudprovider/aws/aws.go
anson627: thanks for the clarification!

Issue #152, opened 2026-04-15
Improving Device Discovery and Filtering Control in DRANET


gauravkghildiyal:
DRANET currently uses a static [--filter](https://github.com/kubernetes-sigs/dranet/blob/81bd6fe8fb1f498ab900b33dbda3ae1db6771cd8/cmd/dranet/app.go#L72) flag (CEL expression) to decide which network devices to publish. While this works for simple cases, it is proving too rigid for complex cloud environments.

The current "static" model has two main limitations:
1.  **No Per-Node Flexibility:** The filter is set in the manifest and is the same for every node in a pool, making it impossible to handle node-specific hardware state.
2.  **Attribute Bloat:** To make the filter "dynamic," we currently have to add extra attributes to devices just so the filter can see them. This can lead to unnecessary bloat in the `ResourceSlice`, and we are already approaching the limit on the max number of attributes allowed per device.

### Why we need it
Recent discussions (e.g. https://github.com/kubernetes-sigs/dranet/pull/138#discussion_r3063973756, https://github.com/kubernetes-sigs/dranet/pull/138#issuecomment-4255910316) have shown that standard logic (like "exclude anything with a default gateway") isn't enough. On these platforms, multiple interfaces might carry default routes even if they are meant for workload traffic.

### Possible approaches (amongst others, feel free to comment)

Instead of just letting cloud providers *add* attributes to devices, we could allow them to **filter** the device list during discovery. 

This keeps the logic simple and localized:
*   Cloud providers can use their existing knowledge (metadata, specific hardware IDs) to explicitly exclude management or host-reserved NICs.
*   We avoid "special-casing" new interfaces for every unique cloud side-effect.
*   We keep the `ResourceSlice` clean by not publishing "internal-only" attributes just for filtering purposes.

/cc @aojea @dkennetzoracle @tamilmani1989

aojea:
Ah, so you mean the cloud provider has a way to return the dynamic filter cel expression?

dkennetzoracle:
To me this reads a bit backwards:

> Cloud providers can use their existing knowledge (metadata, specific hardware IDs) to explicitly exclude management or host-reserved NICs.

OKE needs the opposite of exclusion, while I think Azure does need exclusion which I think is addressed by @aojea PR corrections. But generally speaking, a fairly robust fix that could providers could extend and could participate in device discovery could be a 3 state filter like:

```go
[...]

PR #153, opened 2026-04-16
accurately detect default gateways

aojea:
Address several critical edge cases in default interface detection and introduces a robust, rootless networking test framework.

1. Point-to-Point Interfaces: Removed the `r.Gw == nil` check. Previously, this caused the agent to completely ignore active VPNs and tunnels (like Wireguard or tun/tap) because P2P links route directly out of the device without a Gateway IP.
2. Route Metrics: The kernel relies on metrics (Priority) to determine the active route in a multi-WAN setup. The old code ignored this, returning all interfaces. It now correctly isolates the lowest metric for IPv4 and IPv6 independently, while preserving ECMP support.
3. Multipath Link Lookup: Fixed a bug where multipath routes were queried using the parent route's link index (which is often 0) rather than the nexthop's link index.

<!--  Thanks for sending a pull request!  Here are some tips for you:

1. If this is your first time, please read our contributor guidelines:
   https://git.k8s.io/community/contributors/guide/first-contribution.md#your-first-contribution
   and developer guide
   https://git.k8s.io/community/contributors/devel/development.md#development-guide
2. Please label this pull request according to what type of issue you are
   addressing, especially if this is a release targeted pull request. For
   reference on required PR/issue labels, read here:
   https://git.k8s.io/community/contributors/devel/sig-release/release.md#issuepr-kind-label
3. Ensure you have added or ran the appropriate tests for your PR:
   https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
4. If the PR is unfinished, see how to mark it:
   https://git.k8s.io/community/contributors/guide/pull-requests.md#marking-unfinished-pull-requests

/kind bug

```release-note
Address several critical edge cases in default interface detection
```

Fixes: https://github.com/kubernetes-sigs/dranet/issues/152

### DIFF
--- a/internal/testutils/userns.go
+++ b/internal/testutils/userns.go
@@ -1,0 +1,118 @@
+/*
+Copyright The Kubernetes Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testutils
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+	"sync"
+	"syscall"
+	"testing"
+)
+
+var (
+	isSupported bool
+	checkOnce   sync.Once
+)
+
+// IsSupported checks if unprivileged user namespaces are enabled on the host system.
+// It caches the result after the first check.
+func IsSupported() bool {
+	checkOnce.Do(func() {
+		cmd := exec.Command("sleep", "1")
+
+		// Attempt to map ourselves to root inside the userns.
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Cloneflags:  syscall.CLONE_NEWUSER,
+			UidMappings: []syscall.SysProcIDMap{{ContainerID: 0, HostID: os.Getuid(), Size: 1}},
+			GidMappings: []syscall.SysProcIDMap{{ContainerID: 0, HostID: os.Getgid(), Size: 1}},
+		}
+
+		if err := cmd.Start(); err != nil {
+			return // User namespaces are likely restricted or not supported
+		}
+
+		defer func() {
+			_ = cmd.Process.Kill()
+			_ = cmd.Wait()
+		}()
+
+		isSupported = true
+	})
+
+	return isSupported
+}
+
+// Run executes the given test function inside a user namespace where the
+// current user is mapped to root. This provides capabilities to create network
+// namespaces and netfilter rules without running as actual root on the host.
+//
+// extraCloneflags can be used to request additional namespace types
+// (e.g., syscall.CLONE_NEWNET).
+func Run(t *testing.T, f func(t *testing.T), extraCloneflags ...uintptr) {
+	const subprocessEnvKey = "GO_USERNS_SUBPROCESS_KEY"
+
+	// 1. If we are already inside the subprocess, run the actual test logic.
+	if testIDString, ok := os.LookupEnv(subprocessEnvKey); ok && testIDString == "1" {
+		t.Run("subprocess", f)
+		return
+	}
+
+	// 2. If we are on the host, verify support before attempting to spawn.
+	if !IsSupported() {
+		t.Skip("Unprivileged user namespaces are not supported on this system")
+	}
+
+	// 3. Prepare the command to re-execute the current test binary.
+	cmd := exec.Command(os.Args[0])
+	cmd.Args = []string{os.Args[0], "-test.run=" + t.Name() + "$", "-test.v=true"}
+
+	for _, arg := range os.Args {
+		if strings.HasPrefix(arg, "-test.testlogfile=") {
+			cmd.Args = append(cmd.Args, arg)
+		}
+	}
+
+	cmd.Env = append(os.Environ(), subprocessEnvKey+"=1")
+	// Include sbin in PATH, as some networking commands are not found otherwise.
+	cmd.Env = append(cmd.Env, "PATH=/usr/local/sbin:/usr/sbin:/sbin:"+os.Getenv("PATH"))
+	cmd.Stdin = os.Stdin
+
+	// 4. Configure the namespace clone flags.
+	cloneflags := uintptr(syscall.CLONE_NEWUSER)
+	for _, flag := range extraCloneflags {
+		cloneflags |= flag
+	}
+
+	// Map ourselves to root inside the new user namespace.
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Cloneflags:  cloneflags,
+		UidMappings: []syscall.SysProcIDMap{{ContainerID: 0, HostID: os.Getuid(), Size: 1}},
+		GidMappings: []syscall.SysProcIDMap{{ContainerID: 0, HostID: os.Getgid(), Size: 1}},
+	}
+
+	// 5. Execute and capture output.
+	out, err := cmd.CombinedOutput()
+
+	// Prepending a newline makes the nested test output much easier to read
+	// in the standard `go test` log format.
+	t.Logf("\n%s", out)
+	if err != nil {
+		t.Fatalf("Subprocess execution failed: %v", err)
+	}
+}
--- a/pkg/inventory/net.go
+++ b/pkg/inventory/net.go
@@ -17,6 +17,8 @@ limitations under the License.
 package inventory
 
 import (
+	"math"
+
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
 	"github.com/vishvananda/netlink"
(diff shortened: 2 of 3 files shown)


pkg/inventory/net.go
```diff
@@ -28,19 +30,25 @@ import (
 )
 
 // getDefaultGwInterfaces returns a set of interface names that are configured
-// as default gateways in the main routing table. It identifies these by querying
-// the main routing table for routes with an unspecified destination (0.0.0.0/0
-// for IPv4 or ::/0 for IPv6).
+// as active default gateways in the main routing table, respecting route metrics.
```
dkennetzoracle, line 33: just a note / side-effect, any route matching Dst == 0.0.0.0/0 or ::/0 in the main table now classifies the link as an uplink, which is a new semantic and probably worth a comment?

internal/testutils/userns.go
```diff
@@ -0,0 +1,102 @@
+package testutils
```
dkennetzoracle, line 17: do you need a license header? I see them everywhere else.

aojea: yes, good catch

pkg/inventory/net.go
aojea: is that heuristic not correct?

aojea: it is the same behavior as before, isn;t it? 
it just expands the logic to take into consideration the weights, but previous logic about the 0.0.0.0 and ::/0 remains the same

dkennetzoracle: It is the same logic, but the route no longer has to be a gateway. Any route with scope=link for example but with no gateway but still matches 0.0.0.0/0 or ::/0 will also match here.

pkg/inventory/net_test.go
```diff
@@ -0,0 +1,204 @@
...
+			name: "Independent families (IPv4 and IPv6 on different interfaces)",
+			setupRoutes: func() error {
+				if err := netlink.RouteAdd(&netlink.Route{Family: netlink.FAMILY_V4, Dst: defaultIPv4, Gw: gwIPv4, LinkIndex: links["eth0"].Attrs().Index, Priority: 100, Table: unix.RT_TABLE_MAIN}); err != nil {
+					return err
+				}
+				return netlink.RouteAdd(&netlink.Route{Family: netlink.FAMILY_V6, Dst: defaultIPv6, Gw: gwIPv6, LinkIndex: links["eth2"].Attrs().Index, Priority: 100, Table: unix.RT_TABLE_MAIN})
+			},
+			expectedResult: sets.New[string]("eth0", "eth2"),
+		},
+		{
```
dkennetzoracle, line 144: I'd add a test for Mixed-family default on a single interface. Basically, when you merge sets for ipv4 and ipv6 this asserts that the merge behaves (dual stacking an interface).
```go
{
    name: "Same interface wins both families",
    setupRoutes: func() error {
        if err := netlink.RouteAdd(&netlink.Route{Family: netlink.FAMILY_V4, Dst: defaultIPv4, Gw: gwIPv4, LinkIndex: links["eth0"].Attrs().Index, Priority: 100, Table: unix.RT_TABLE_MAIN}); err != nil {
            return err
        }
        return netlink.RouteAdd(&netlink.Route{Family: netlink.FAMILY_V6, Dst: defaultIPv6, Gw: gwIPv6, LinkIndex: links["eth0"].Attrs().Index, Priority: 100, Table: unix.RT_TABLE_MAIN})
    },
    expectedResult: sets.New[string]("eth0"),
},
```

pkg/inventory/net_test.go
```diff
@@ -0,0 +1,204 @@
...
+			name: "Point-to-Point Interface (No Gateway IP)",
+			setupRoutes: func() error {
+				// Intentionally leaving Gw == nil here because Point-to-Point links don't have gateways
+				if err := netlink.RouteAdd(&netlink.Route{Family: netlink.FAMILY_V4, Dst: defaultIPv4, LinkIndex: links["wg0"].Attrs().Index, Priority: 50, Scope: netlink.SCOPE_LINK, Table: unix.RT_TABLE_MAIN}); err != nil {
+					return err
+				}
+				return netlink.RouteAdd(&netlink.Route{Family: netlink.FAMILY_V4, Dst: defaultIPv4, Gw: gwIPv4, LinkIndex: links["eth0"].Attrs().Index, Priority: 100, Table: unix.RT_TABLE_MAIN})
+			},
+			expectedResult: sets.New[string]("wg0"),
+		},
```
dkennetzoracle, line 181: Would it be worth also having a pinned Priority: 0 for P2P case?

You re using Prio 50 for wg0 and this exercises the GW == nil removal as well as the the lowest metric wins logic. However, it doesn't exercise a specific case where a P2P route has Priority: 0 which is the ipv4 default for a route installed without an explicit metric. This case would just lock in that 0 is a real comparable value.

OK with not doing it, it just feels a bit more explicit and defensive.

```go
{
    name: "P2P default with Priority 0 wins over gateway'd default",
    setupRoutes: func() error {
        if err := netlink.RouteAdd(&netlink.Route{Family: netlink.FAMILY_V4, Dst: defaultIPv4, LinkIndex: links["wg0"].Attrs().Index, Priority: 0, Scope: netlink.SCOPE_LINK, Table: unix.RT_TABLE_MAIN}); err != nil {
            return err
        }
        return netlink.RouteAdd(&netlink.Route{Family: netlink.FAMILY_V4, Dst: defaultIPv4, Gw: gwIPv4, LinkIndex: links["eth0"].Attrs().Index, Priority: 100, Table: unix.RT_TABLE_MAIN})
    },
    expectedResult: sets.New[string]("wg0"),
},
```

internal/testutils/userns.go
```diff
@@ -0,0 +1,102 @@
...
+var (
+	isSupported bool
+	checkOnce   sync.Once
+)
+
+// IsSupported checks if unprivileged user namespaces are enabled on the host system.
+// It caches the result after the first check.
+func IsSupported() bool {
+	checkOnce.Do(func() {
+		cmd := exec.Command("sleep", "1")
```
dkennetzoracle, line 37: A small nit: this requires `sleep` to be on the system in `PATH`. If it's not, you'd get an error and it would be because of a missing binary not because user namespaces are not supported. Could check it?

internal/testutils/userns.go
```diff
@@ -0,0 +1,102 @@
...
+	if !IsSupported() {
+		t.Skip("Unprivileged user namespaces are not supported on this system")
+	}
+
+	// 3. Prepare the command to re-execute the current test binary.
+	cmd := exec.Command(os.Args[0])
+	cmd.Args = []string{os.Args[0], "-test.run=" + t.Name() + "$", "-test.v=true"}
+
+	for _, arg := range os.Args {
+		if strings.HasPrefix(arg, "-test.testlogfile=") {
```
dkennetzoracle, line 86: would there be any reason to add `-test.timeout` here? Any of these long?

dkennetzoracle (approved):
Some nits / questions, but I think it's a great add and a correctness fix. I'm not 100% sure it fixes the `NonUplinkChecker` situation on #138 on OKE, I'd need to double check. RA injects IPv6 defaults for us. If they share a metric (which I think they do) I still need something like NonUplinkChecker. If they don't I can drop NonUplinkChecker.

This definitely addresses one of the sub-issues in that PR, and is correct. LGTM!

gauravkghildiyal:
> I'm not 100% sure it fixes the NonUplinkChecker situation on https://github.com/kubernetes-sigs/dranet/pull/138 on OKE, I'd need to double check.

@dkennetzoracle I'm very much interested in hearing about this after you've had the opportunity to try this out

dkennetzoracle:
@gauravkghildiyal - I will let you know! It's very hard for me to get access to GB200+ so I need to make the most of my time on them haha. I'm pretty sure each NIC gets a default route for ipv6 and these use the kernel's RA priority (so 1024) so we wouldn't be able to differentiate the 8 RDMA NICs from each other by metric. They all look like peer's from he kernel's pov.

So they'd all get excluded still here, haha. But it should work for the Azure case

pkg/inventory/net.go
anson627: +1 on semantic broadening — "default route with a gateway" to "any default route"

internal/testutils/userns.go
anson627: maybe `exec.Command(os.Args[0], "-test.run=^$")` to avoid the external sleep dependency

pkg/inventory/net.go
aojea: what is the difference, from the routing perspective the fact of having a next hop direction or just send to via the corresponding interface is not important, what is important is that any packet in the host will go through that specific interface ... that is why we don't want to expose it in the ResourceSlice

dkennetzoracle: I don't feel super strongly about it, (to me) it just adds clarity.

anson627:
I just verified this PR on AWS p4d.24large, the primary NiC with default gateway/route (e.g. ens32) is excluded, and 4 EFA/ENA devices are properly included in the resource slice:

  | Device           | Type | Interface | PCI Address  | NUMA | IP                  |
|------------------|------|-----------|--------------|------|---------------------|
| pci-0000-10-01-0 | ENA  | ens33     | 0000:10:01.0 | 0    | 192.168.157.168/19  |
| pci-0000-10-1b-0 | EFA  | —         | 0000:10:1b.0 | 0    | rdmap16s27          |
| pci-0000-20-01-0 | ENA  | ens65     | 0000:20:01.0 | 0    | 192.168.155.210/19  |
| pci-0000-20-1b-0 | EFA  | —         | 0000:20:1b.0 | 0    | rdmap32s27          |
| pci-0000-90-01-0 | ENA  | ens129    | 0000:90:01.0 | 1    | 192.168.145.129/19  |
| pci-0000-90-1b-0 | EFA  | —         | 0000:90:1b.0 | 1    | rdmap144s27         |
| pci-0000-a0-01-0 | ENA  | ens161    | 0000:a0:01.0 | 1    | 192.168.132.181/19  |
| pci-0000-a0-1b-0 | EFA  | —         | 0000:a0:1b.0 | 1    | rdmap160s27         |

anson627:
verified this PR on Azure GB300, the primary NiC (eth0) with default gateway is excluded, while Azure accelerated networking VF (e.g. enP8051s1) together with other mlx5_* NiCs are included

| # | Name             | RDMA Device | PCI Address  | NUMA | Vendor   | RDMA |
|---|------------------|-------------|--------------|------|----------|------|
| 1 | pci-0101-00-00-0 | mlx5_0      | 0101:00:00.0 | 0    | Mellanox | true |
| 2 | pci-0102-00-00-0 | mlx5_1      | 0102:00:00.0 | 0    | Mellanox | true |
| 3 | pci-0103-00-00-0 | mlx5_2      | 0103:00:00.0 | 1    | Mellanox | true |
| 4 | pci-0104-00-00-0 | mlx5_3      | 0104:00:00.0 | 1    | Mellanox | true |
| 5 | pci-1f73-00-02-0 | —           | 1f73:00:02.0 | 0    | Mellanox | true |

/lgtm

anson627 (approved):

gauravkghildiyal:
Holding to have time to go through this and allow resolution of a few comments.

/hold

aojea:
> RA injects IPv6 defaults for us. If they share a metric (which I think they do) I still need something like NonUplinkChecker. If they don't I can drop NonUplinkChecker.

From a kubernetes architectural and security perspective, we like to have certain guardrails so users can not shoot themselves on the feet. Any logic that allows an interface with a default route to be unmounted or modified based on an opportunistic RA injection is very risky.:

- In the Kubernetes networking model, there is typically no difference between the control plane and the dataplane traffic. If an interface intended for the dataplane receives a default route via RA, it will be 'hijacking' control plane traffic, so the kubelet and pods trying to connect to the apiserver will fail (unless this secondary interface has access too, but this looks a very convoluted setup with routing loops). This will effectively disconnect the entire node from the cluster.

- We already were hit by security issues "abusing" RA, so right now is RA processing disabled by default in all container interfaces, the same attack can be performed at the node level

> A vulnerability was found in all versions of containernetworking/plugins before version 0.8.6, that allows malicious containers in Kubernetes clusters to perform man-in-the-middle (MitM) attacks. A malicious container can exploit this flaw by sending rogue IPv6 router advertisements to the host or other containers, to redirect traffic to the malicious container. ([CVE-2020-10749](https://www.cve.org/CVERecord?id=CVE-2020-10749))

I still think we need a better mechanism for improving filtering, so I encourage you to check @gauravkghildiyal proposal in https://github.com/kubernetes-sigs/dranet/issues/152 , that I think will be able to accomodate the `NonUplinkChecker` functionality perfectly

dkennetzoracle:
@aojea aligned with @gauravkghildiyal 's issue, it seems like a good fix and a generally better way to approach the problem!

pkg/inventory/net.go
aojea: fair

pkg/inventory/net_test.go
```diff
@@ -0,0 +1,204 @@
...
+					Gw:        gwIPv4, // Added Gateway
+					LinkIndex: links["eth0"].Attrs().Index,
+					Priority:  100,
+					Table:     unix.RT_TABLE_MAIN,
+				})
+			},
+			expectedResult: sets.New[string]("eth0"),
+		},
+		{
+			name: "Lower metric wins (IPv4)",
```
tamilmani1989, line 105: should we add one test for ipv6 with 2 default routes?

tamilmani1989 (approved):
changes lgtm. left a comment, i will leave to your discretion

internal/testutils/userns.go
aojea: this setup was been working for more than one year in other projects like kube-network-policies, so I always prefer to wait for evidence to do changes rather than premature optimizations

internal/testutils/userns.go
aojea: same as https://github.com/kubernetes-sigs/dranet/pull/153#discussion_r3102972405

aojea:
pushed addressing latest comments, great reviews, caught some flaws:
- clarify function scope explicitly indicating it queries for the unspecified routes 0.0.0.0/0, ::/0 or nil that are the kernel default networks
- adding new unit tests

pkg/inventory/net_test.go
aojea: added

pkg/inventory/net_test.go
aojea: added

pkg/inventory/net_test.go
aojea: added as "P2P default with Priority 0 wins over gateway default"

dkennetzoracle:
@aojea @gauravkghildiyal I can confirm this fixed our issues (besides NIC orphaning) on OCI. I pulled in our OKE changes to this branch locally and deployed, and didn't need any additional changes. We made a few additional changes on our side, and fixed our dual ipv6 stack deployment. With these changes, the only thing we are missing is the cri upstream changes.

aojea:
/hold cancel

based on the feedback from OKE and Azure, @gauravkghildiyal for final review

gauravkghildiyal:
Perfect, thanks for validating @anson627 and @dkennetzoracle 

Thanks for the change @aojea 

/lgtm
/approve

PR #154, opened 2026-04-16
fix: Remove 'dranet' from registry name to avoid duplication

gauravkghildiyal:
#### What type of PR is this?

/kind cleanup

Without this change, the image is pushed to `gcr.io/k8s-staging-networking/charts/dranet/dranet:1.2.0` ([example prow job](https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-dranet-image/2044876539446693888)). With this change, we avoid the duplicate 'dranet' at the end

```release-note
NONE
```

/cc @fmuyassarov


fmuyassarov (approved):
Oh my bad. I totally missed the fact that helm push automatically appends the chart name to the destination repository path. Thanks for the fix @gauravkghildiyal 
/lgtm

gauravkghildiyal:
@fmuyassarov No worries at all. Glad to see it's working smoothly. Thanks again

PR #156, opened 2026-04-18
Azure: generate per-device network config from IMDS metadata

tamilmani1989:
Implements `GetDeviceConfig` for the Azure cloud provider by fetching
network interface metadata from IMDS and generating policy routing
rules and routes for each device based on its MAC address.

## Changes

- Fetch network interface metadata from Azure IMDS (`/metadata/instance/network`)
  during instance initialization via `fetchNetworkInterfaces()`
- Implement `GetDeviceConfig()` to generate per-device `NetworkConfig` with
  IPv4 and IPv6 policy routing rules and routes derived from IMDS subnet info
- Match devices to IMDS interfaces using MAC address (case-insensitive,
  separator-agnostic)
- Skip IPv6 link-local addresses (fe80::/10) when generating IPv6 routes
- Add helper functions: `normalizeMAC`, `subnetFirstAddress`, `isIPv6LinkLocal`
- Add unit tests for changes introduced by this PR

#### What type of PR is this?
/kind feature

```release-note
Add azure specific rules and routes to be configured for secondary nic.
```

### DIFF
--- a/pkg/cloudprovider/azure/azure.go
+++ b/pkg/cloudprovider/azure/azure.go
@@ -21,13 +21,19 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"net/netip"
+	"strings"
 	"time"
 
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 
 	resourceapi "k8s.io/api/resource/v1"
+	"sigs.k8s.io/dranet/internal/nlwrap"
 	"sigs.k8s.io/dranet/pkg/apis"
 	"sigs.k8s.io/dranet/pkg/cloudprovider"
 )
@@ -42,6 +48,10 @@ const (
 	imdsEndpoint = "http://169.254.169.254/metadata/instance"
 	// imdsAPIVersion is the API version used for IMDS queries.
 	imdsAPIVersion = "2021-02-01"
+	// imdsPathCompute is the IMDS path for compute metadata.
+	imdsPathCompute = "compute"
+	// imdsPathNetwork is the IMDS path for network metadata.
+	imdsPathNetwork = "network"
 )
 
 // imdsComputeMetadata contains the fields we care about from the Azure IMDS
@@ -56,12 +66,53 @@ type imdsResponse struct {
 	Compute imdsComputeMetadata `json:"compute"`
 }
 
+// imdsNetworkResponse represents the IMDS network metadata response.
+type imdsNetworkResponse struct {
+	Interface []networkInterface `json:"interface"`
+}
+
+// networkInterface represents a single network interface from IMDS.
+type networkInterface struct {
+	IPv4       ipv4Config `json:"ipv4"`
+	IPv6       ipv6Config `json:"ipv6"`
+	MacAddress string     `json:"macAddress"`
+}
+
+// ipv4Config represents the IPv4 configuration from IMDS.
+type ipv4Config struct {
+	IPAddress []ipv4Address `json:"ipAddress"`
+	Subnet    []subnet      `json:"subnet"`
+}
+
+// ipv6Config represents the IPv6 configuration from IMDS.
+type ipv6Config struct {
+	IPAddress []ipv6Address `json:"ipAddress"`
+}
+
+// ipv4Address represents a single IPv4 address from IMDS.
+type ipv4Address struct {
+	PrivateIPAddress string `json:"privateIpAddress"`
+	PublicIPAddress  string `json:"publicIpAddress"`
+}
+
+// ipv6Address represents a single IPv6 address from IMDS.
+type ipv6Address struct {
+	PrivateIPAddress string `json:"privateIpAddress"`
+}
+
+// subnet represents a subnet from IMDS.
+type subnet struct {
+	Address string `json:"address"`
+	Prefix  string `json:"prefix"`
+}
+
 var _ cloudprovider.CloudInstance = (*AzureInstance)(nil)
 
 // AzureInstance holds Azure-specific instance data retrieved from IMDS.
 type AzureInstance struct {
 	PlacementGroupID string
 	VMSize           string
+	Interfaces       []networkInterface
 }
 
 // GetDeviceAttributes returns Azure-specific attributes for a device.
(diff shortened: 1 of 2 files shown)


aojea:
/ok-to-test

@tamilmani1989 you should apply for kubernetes-sig org membership

anson627:
> /assign @anson627 @MikeZappa87

thanks @tamilmani1989 for the PR! I will test and verify, and provide feedbacks today.

tamilmani1989:
> /ok-to-test
> 
> @tamilmani1989 you should apply for kubernetes-sig org membership

@aojea would you able to approve my request https://github.com/kubernetes/org/issues/6306

pkg/cloudprovider/azure/azure.go
```diff
@@ -81,10 +126,106 @@ func (a *AzureInstance) GetDeviceAttributes(id cloudprovider.DeviceIdentifiers)
 	return attributes
 }
 
-// GetDeviceConfig returns nil as Azure does not currently provide
-// device-specific network configuration via IMDS.
+const (
+	// azureIPv6Gateway is the fixed IPv6 gateway for Azure.
+	azureIPv6Gateway = "fe80::1234:5678:9abc"
+	// routingTable is the routing table ID used for policy routing.
+	routingTable = 100
```
anson627: would this conflicting 0.0.0.0/0 default routes in one table from different NiC? can it be derived from the NiC index?

pkg/cloudprovider/azure/azure.go
```diff
@@ -81,10 +126,106 @@ func (a *AzureInstance) GetDeviceAttributes(id cloudprovider.DeviceIdentifiers)
 	return attributes
 }
 
-// GetDeviceConfig returns nil as Azure does not currently provide
-// device-specific network configuration via IMDS.
+const (
+	// azureIPv6Gateway is the fixed IPv6 gateway for Azure.
+	azureIPv6Gateway = "fe80::1234:5678:9abc"
```
anson627: add a link to this magic ipv6 gateway?

pkg/cloudprovider/azure/azure.go
```diff
@@ -81,10 +126,106 @@ func (a *AzureInstance) GetDeviceAttributes(id cloudprovider.DeviceIdentifiers)
...
+
+	config := &apis.NetworkConfig{}
+
+	// IPv4 rules and routes
+	if len(iface.IPv4.Subnet) > 0 {
+		subnet := iface.IPv4.Subnet[0]
+		gateway := subnetFirstAddress(subnet.Address)
+		if gateway != "" {
+			config.Rules = append(config.Rules, apis.RuleConfig{
+				Source: subnet.Address + "/" + subnet.Prefix,
```
anson627, line 176: check subnet.Prefix == "" case?

pkg/cloudprovider/azure/azure.go
```diff
@@ -162,5 +303,61 @@ func GetInstance(ctx context.Context) (cloudprovider.CloudInstance, error) {
...
+	if err := fetchNetworkInterfaces(ctx, client, instance); err != nil {
+		klog.Warningf("Failed to retrieve Azure IMDS network metadata: %v", err)
+	}
+
 	return instance, nil
 }
+
+// fetchNetworkInterfaces queries IMDS for network interface metadata and
+// populates the instance's Interfaces field.
+func fetchNetworkInterfaces(ctx context.Context, client *http.Client, instance *AzureInstance) error {
```
anson627: refactor with a helper to handle http with retry, and have both GetInstance and fetchNetworkInterfaces use the help?

pkg/cloudprovider/azure/azure.go
```diff
@@ -56,12 +59,54 @@ type imdsResponse struct {
 	Compute imdsComputeMetadata `json:"compute"`
 }
 
+// imdsNetworkResponse represents the IMDS network metadata response.
+type imdsNetworkResponse struct {
+	Interface []networkInterface `json:"interface"`
+}
+
+// networkInterface represents a single network interface from IMDS.
+
```
anson627: rm empty line?

pkg/cloudprovider/azure/azure.go
```diff
@@ -81,10 +126,106 @@ func (a *AzureInstance) GetDeviceAttributes(id cloudprovider.DeviceIdentifiers)
...
+// GetDeviceConfig returns Azure-specific network configuration (rules and routes)
+// for a device identified by its MAC address. It reads subnet info from IMDS
+// and computes policy routing rules and routes for both IPv4 and IPv6.
 func (a *AzureInstance) GetDeviceConfig(id cloudprovider.DeviceIdentifiers) *apis.NetworkConfig {
-	return nil
+	if id.MAC == "" {
+		return nil
+	}
+
+	normalizedMAC := normalizeMAC(id.MAC)
```
anson627, line 149: maybe use net.ParseMAC instead, since it already handles format validation

pkg/cloudprovider/azure/azure_test.go
```diff
@@ -95,12 +96,214 @@ func TestGetDeviceAttributes(t *testing.T) {
...
 func TestGetDeviceConfig(t *testing.T) {
-	instance := &AzureInstance{
-		PlacementGroupID: "test-group",
-		VMSize:           "Standard_D4s_v3",
+	tests := []struct {
+		name     string
+		instance *AzureInstance
+		id       cloudprovider.DeviceIdentifiers
+		want     *apis.NetworkConfig
+	}{
```
anson627, line 151: add a negative test for subnetFirstAddress("10.0.0.255") etc

gauravkghildiyal:
Seems like `tamilmani1989:tm/azureNetworkConfig` branch is pretty far behind and rebasing to head will hopefully fix the doc build errors.

Also, please fix the labels from the CL description

pkg/cloudprovider/azure/azure.go
tamilmani1989: couldn't find any public doc on this address.

pkg/cloudprovider/azure/azure.go
tamilmani1989: net.ParseMAC supported formats:

```
//	00:00:5e:00:53:01
//	02:00:5e:10:00:00:00:01
//	00:00:00:00:fe:80:00:00:00:00:00:00:02:00:5e:10:00:00:00:01
//	00-00-5e-00-53-01
//	02-00-5e-10-00-00-00-01
//	00-00-00-00-fe-80-00-00-00-00-00-00-02-00-5e-10-00-00-00-01
//	0000.5e00.5301
//	0200.5e10.0000.0001
//	0000.0000.fe80.0000.0000.0000.0200.5e10.0000.0001
```
it does not support `00123456789abc` format. azure imds returns mac address in this format

pkg/cloudprovider/azure/azure.go
tamilmani1989: yep good catch. routing table id can be calculated from nic index

anson627:
I tested on GB300 nodes, and can see `mac` address but not `ipv4` address, my test setup did not enable dual-stack, so no `ipv6` address is expected
```
  - attributes:
      azure.dra.net/placementGroupId:
        string: c0104dd3-6665-45bf-afde-b2f35e590b93
      azure.dra.net/vmSize:
        string: Standard_ND96isr_MI300X_v5
      dra.net/alias:
        string: ""
      dra.net/ebpf:
        bool: false
      dra.net/encapsulation:
        string: ether
      dra.net/ifName:
        string: enP63016s1
      dra.net/mac:
        string: 7c:ed:8d:6d:bb:4c
      dra.net/mtu:
        int: 1500
      dra.net/numaNode:
        int: 0
      dra.net/pciAddress:
        string: f628:00:02.0
      dra.net/pciDevice:
        string: MT28800 Family [ConnectX-5 Ex Virtual Function]
      dra.net/pciSubsystem:
        string: "0127"
      dra.net/pciVendor:
        string: Mellanox Technologies
      dra.net/rdma:
        bool: true
      dra.net/sriov:
        bool: false
      dra.net/state:
        string: up
      dra.net/type:
        string: device
      dra.net/virtual:
        bool: false
    name: pci-f628-00-02-0
  driver: dra.net
  nodeName: aks-gpu-37973606-vmss000000
  pool:
    generation: 1
    name: aks-gpu-37973606-vmss000000
```

anson627:
I did a bit more testing by creating a resource claim template and test pod and found RDMA device `mlx5_8` under /dev/infiniband is injected into the pod as expected, but I also noticed that host uplink VF `enP63016s1` is injected into pod as netdev, which will cause issue as `enP63016s1` is attached under `eth0` on the host, and it is not an independent NIC:
- /sys/class/net/enP63016s1/master -> .../net/eth0
- /sys/class/net/eth0/lower_enP63016s1 -> .../net/enP63016s1

the doc also confirmed this: https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works
@tamilmani1989

tamilmani1989:
> I did a bit more testing by creating a resource claim template and test pod and found RDMA device `mlx5_8` under /dev/infiniband is injected into the pod as expected, but I also noticed that host uplink VF `enP63016s1` is injected into pod as netdev, which will cause issue as `enP63016s1` is attached under `eth0` on the host, and it is not an independent NIC:
> 
> * /sys/class/net/enP63016s1/master -> .../net/eth0
> * /sys/class/net/eth0/lower_enP63016s1 -> .../net/enP63016s1
> 
> the doc also confirmed this: https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works @tamilmani1989

Yes, this bug needs to be addressed separately. The same issue occurs when assigning RDMA/IB nics on Azure. Currently, dranet only ignores eth0 (default gateway interface) but not its slave enP* interface, which is the VF exposed explicitly by azure. Ignoring the default gateway interface should also ignore the corresponding enP* VF interface. I'll open a separate PR for this since the bug is not specific to the changes in this PR.

pkg/cloudprovider/azure/azure.go
aojea: eventually someone will find out and will depend on that, we got bitten by this in the past, my suggestion is if you are going to build on that you ask the internal team to make clear on the public docs the expectation about this magic ip

anson627:
> > I did a bit more testing by creating a resource claim template and test pod and found RDMA device `mlx5_8` under /dev/infiniband is injected into the pod as expected, but I also noticed that host uplink VF `enP63016s1` is injected into pod as netdev, which will cause issue as `enP63016s1` is attached under `eth0` on the host, and it is not an independent NIC:
> > 
> > * /sys/class/net/enP63016s1/master -> .../net/eth0
> > * /sys/class/net/eth0/lower_enP63016s1 -> .../net/enP63016s1
> > 
> > the doc also confirmed this: https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works @tamilmani1989
> 
> Yes, this bug needs to be addressed separately. The same issue occurs when assigning RDMA/IB nics on Azure. Currently, dranet only ignores eth0 (default gateway interface) but not its slave enP* interface, which is the VF exposed explicitly by azure. Ignoring the default gateway interface should also ignore the corresponding enP* VF interface. I'll open a separate PR for this since the bug is not specific to the changes in this PR.

issue opened to track the bug: https://github.com/kubernetes-sigs/dranet/issues/175

pkg/cloudprovider/azure/azure.go
rbtr: We fixed this, I think it shipped in Go 1.26
https://github.com/golang/go/issues/66682

pkg/cloudprovider/azure/azure.go
tamilmani1989: one other option is to retrieve this from default gateway route configured on the node itself.. then no need to hardcode this. i can update that..

tamilmani1989: fixed it

pkg/cloudprovider/azure/azure.go
tamilmani1989: still dranet use go 1.25. will update once it start using 1.26

anson627 (approved):

pkg/cloudprovider/azure/azure.go
```diff
@@ -81,10 +132,152 @@ func (a *AzureInstance) GetDeviceAttributes(id cloudprovider.DeviceIdentifiers)
...
+	ipv4 := ipNet.IP.To4()
+	if ipv4 == nil {
+		return "", fmt.Errorf("%s is not an IPv4 address", subnetAddr)
+	}
+	gateway := make(net.IP, 4)
+	binary.BigEndian.PutUint32(gateway, binary.BigEndian.Uint32(ipv4)+1)
+	if !ipNet.Contains(gateway) {
+		return "", fmt.Errorf("gateway %s falls outside subnet %s/%s", gateway, subnetAddr, prefix)
+	}
+	return gateway.String(), nil
```
aojea: this is much easier with `netip.Addr`

https://github.com/kubernetes/kubernetes/blob/8629b872110afd405b71000525d3a5a46ec8ac7b/pkg/registry/core/service/ipallocator/ipallocator.go#L105

```
	ipFirst := prefix.Masked().Addr().Next()

```

aojea:
let me know if you want to simplify the code or get it merged ... overall looks good, can't judge the logic specific to the provider

tamilmani1989:
> let me know if you want to simplify the code or get it merged ... overall looks good, can't judge the logic specific to the provider

I fixed it. Thanks for your review!

pkg/cloudprovider/azure/azure.go
```diff
@@ -81,10 +132,151 @@ func (a *AzureInstance) GetDeviceAttributes(id cloudprovider.DeviceIdentifiers)
...
+		} else {
+			config.Rules = append(config.Rules, apis.RuleConfig{
+				Source: subnet.Address + "/" + subnet.Prefix,
+				Table:  tableID,
+			})
+			config.Routes = append(config.Routes, apis.RouteConfig{
+				Destination: "0.0.0.0/0",
+				Gateway:     gateway,
+				Table:       tableID,
+			})
```
aojea, line 183: what I noted is that if you follow the convention that each Interfaces has its gateway as the first address of the subnet, when you add two interfaces to a pod ... you are going to have two upstream gateways?

tamilmani1989: yes but not in main routing table. this default gateway gets added in custom routing table. Each secondary nic projected inside pod by dranet will have its own routing table.

tamilmani1989:
@aojea i responded to your comment. Please let know if any other concerns

gauravkghildiyal:
@tamilmani1989 The PR description says this is a "/kind feature". Does this need any release notes, and if yes could you please update the template or include NONE in the release notes section.

(I took the liberty of removing all the other labels that were mistakenly added to the PR)

pkg/cloudprovider/azure/azure.go
aojea: doh, right, I see tableID now , sorry ... 

/hold cancel
/lgtm
/approve

Thanks

PR #160, opened 2026-04-20
helm: default image tag to Chart.AppVersion

fmuyassarov:
#### What type of PR is this?
/kind cleanup

helm-package sets `--app-version $(HELM_TAG)` (example v0.1.0) which updates `Chart.appVersion` in the packaged chart. But the DaemonSet template uses `.Values.image.tag`, which is hardcoded to stable in values.yaml so the packaged chart still pulls the stable image. To fix this, use the Helm pattern by making `image.tag` default to  `.Chart.AppVersion` when not explicitly set.
Sorry for missing this earlier. Just came to realize while testing out few other things related to the driver. 

```release-note
NONE
```

### DIFF
--- a/deployments/helm/dranet/templates/daemonset.yaml
+++ b/deployments/helm/dranet/templates/daemonset.yaml
@@ -45,7 +45,7 @@ spec:
         {{- toYaml .Values.tolerations | nindent 8 }}
       containers:
         - name: dranet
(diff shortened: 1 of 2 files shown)


gauravkghildiyal:
Thanks!

/lgtm
/approve

PR #172, opened 2026-04-21
Group non-k8s dependency for dependabot PRs

gauravkghildiyal:
#### What type of PR is this?

/kind cleanup

This PR makes dependabot group gomod dependency bumps for non-k8s libraries into a single PR, thereby reducing PR noise. This should be preferable for now. If need for having a different grouping shows up in the future, we can switch then.

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,9 +8,15 @@ updates:
       k8s-deps:
         patterns:
           - "*k8s.io*"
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
Looks good and cleaner approach especially for people reviewing the patches.
/lgtm

PR #176, opened 2026-04-22
prevent uplink child virtual function injected into pods

### LINKED ISSUE #175: Do not move uplink child virtual functions into pod netns
## Summary

On AKS, `dranet` excludes the default-route uplink interface itself (for example `eth0`), but it can still publish a child VF attached to that uplink (for example `enP63016s1`). The child VF is specific for Azure accelerated networking as it is documented here: https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works

If a pod claims that VF, `dranet` attempts to move the VF into the pod network namespace, which causes pod sandbox setup to fail.

## Problem

The current inventory logic filters only the exact default gateway interface name.

That means:

- `eth0` is excluded from discovery
- `enP63016s1` is still published in the `ResourceSlice`
- a `ResourceClaim` can allocate `pci-f628-00-02-0`
- `dranet` then treats `enP63016s1` as an allocatable pod netdev and attempts to move it into the pod

This is incorrect because `enP63016s1` is part of the host uplink datapath.

## Evidence

Host topology on the failing node:

- `/sys/class/net/enP63016s1/master -> .../net/eth0`
- `/sys/class/net/eth0/lower_enP63016s1 -> .../net/enP63016s1`
- `/sys/class/net/enP63016s1/upper_eth0 -> .../net/eth0`

`enP63016s1` is genuinely RDMA-capable:

- `/sys/class/net/enP63016s1/device/infiniband/mlx5_8` exists

`ResourceSlice` published:

- `pci-f628-00-02-0`
- `dra.net/ifName: enP63016s1`
- `dra.net/rdma: true`
- `dra.net/rdmaDevice: mlx5_8`

`dranet` logs during pod startup:

```text
RunPodSandbox processing device: pci-f628-00-02-0
RunPodSandbox processing Network device: enP63016s1
```

Pod sandbox failure:

```text
Failed to create pod sandbox: rpc error: code = Unknown desc = NRI RunPodSandbox failed:
rpc error: code = Unknown desc = error moving network device pci-f628-00-02-0 to namespace
... link not found for interface enP63016s1 ...
```

## Reproduction

1. Create a `ResourceClaim` that selects `device.attributes["dra.net"].pciAddress == "f628:00:02.0"`
2. Schedule a pod using that claim onto `aks-gpu-37973606-vmss000000`
3. Observe pod sandbox creation fail during `RunPodSandbox`

## Proposed fix

- detect interfaces that are children or descendants of excluded uplinks by using Linux link hierarchy rather than cloud-specific heuristics
- for non-RDMA uplink child devices, exclude them from publication entirely

This preserves verbs access while preventing `RunPodSandbox` from calling the netdev move path for a host-uplink-owned VF.

anson627:
#### What type of PR is this?
/kind bug. 

This PR prevents `dranet` from treating RDMA-capable child interfaces of excluded uplinks as movable pod network devices.

On AKS, the default-route uplink (`eth0`) is excluded today, but its child VF (`enP63016s1`) can still be published in the `ResourceSlice`. When a pod claims that device, `dranet` tries to move the VF into the pod netns and pod sandbox creation fails.

Fixes #175

```release-note
dranet no longer injects child VFs attached to an excluded uplink as pod network interfaces.
```

### DIFF
--- a/pkg/inventory/net.go
+++ b/pkg/inventory/net.go
@@ -127,44 +127,34 @@ func getExcludedUplinkInterfaces() sets.Set[string] {
 		return excluded
 	}
 
-	// Index links by their interface index so we can resolve MasterIndex
-	// lookups in O(1) and walk the hierarchy without re-querying netlink.
-	linksByIndex := make(map[int]netlink.Link, len(links))
-	nameByIndex := make(map[int]string, len(links))
+	// Build a parent-index -> children adjacency map in a single pass so we
+	// can cull whole families in one mutating walk instead of re-scanning the
+	// link list per level of nesting.
+	childrenOf := make(map[int][]netlink.Link)
+	var seeds []int
 	for _, l := range links {
-		idx := l.Attrs().Index
-		linksByIndex[idx] = l
-		nameByIndex[idx] = l.Attrs().Name
-	}
-
-	excludedIndices := sets.New[int]()
-	for idx, name := range nameByIndex {
-		if excluded.Has(name) {
-			excludedIndices.Insert(idx)
+		attrs := l.Attrs()
+		if attrs.MasterIndex != 0 {
+			childrenOf[attrs.MasterIndex] = append(childrenOf[attrs.MasterIndex], l)
+		}
+		if excluded.Has(attrs.Name) {
+			seeds = append(seeds, attrs.Index)
 		}
 	}
 
-	// Expand the excluded set by repeatedly walking the link list and adding
-	// any link whose master is already excluded. Stop once a full pass adds
-	// nothing new (handles nested hierarchies like vf -> vf-child -> uplink).
-	for {
-		added := false
-		for _, l := range links {
-			attrs := l.Attrs()
-			if excluded.Has(attrs.Name) {
-				continue
-			}
-			if attrs.MasterIndex == 0 {
-				continue
-			}
-			if excludedIndices.Has(attrs.MasterIndex) {
-				excluded.Insert(attrs.Name)
-				excludedIndices.Insert(attrs.Index)
-				added = true
-			}
+	// BFS from each excluded uplink through the adjacency map. Deleting the
+	// entry after visiting guarantees each child is processed at most once
+	// even if the hierarchy is deep (vf -> vf-child -> uplink).
+	for i := 0; i < len(seeds); i++ {
+		children, found := childrenOf[seeds[i]]
+		if !found {
+			continue
 		}
-		if !added {
-			break
+		delete(childrenOf, seeds[i])
+		for _, child := range children {
+			attrs := child.Attrs()
+			excluded.Insert(attrs.Name)
+			seeds = append(seeds, attrs.Index)
 		}
 	}
 
--- a/pkg/inventory/net.go
+++ b/pkg/inventory/net.go
@@ -117,7 +117,14 @@ func getDefaultGwInterfaces() sets.Set[string] {
 
 // getExcludedUplinkInterfaces returns the set of interface names that must be
 // excluded from the inventory: the active default-gateway uplinks plus every
-// netdev that is a descendant of one of those uplinks in the link hierarchy.
+// netdev that is a descendant of one of those uplinks. A child tied to a
+// parent through MasterIndex (bond/team slave, bridge port, VF enslaved to
+// its PF, ...) shares its forwarding state with that parent, so moving just
+// the child into a pod netns strands it from the parent that owns that
+// state; when the parent is the host's default-gw uplink this also degrades
+// host connectivity. There is no scenario where relocating only the child of
+// a default-gw uplink is correct, so the entire MasterIndex-linked subtree
+// rooted at each uplink should be excluded.
 func getExcludedUplinkInterfaces() sets.Set[string] {
 	excluded := getDefaultGwInterfaces()
 
--- a/pkg/inventory/net.go
+++ b/pkg/inventory/net.go
@@ -115,6 +115,62 @@ func getDefaultGwInterfaces() sets.Set[string] {
 	return interfaces
 }
 
+// getExcludedUplinkInterfaces returns the set of interface names that must be
+// excluded from the inventory: the active default-gateway uplinks plus every
+// netdev that is a descendant of one of those uplinks in the link hierarchy.
+func getExcludedUplinkInterfaces() sets.Set[string] {
+	excluded := getDefaultGwInterfaces()
+
+	links, err := nlwrap.LinkList()
+	if err != nil {
+		klog.Errorf("Failed to list links for uplink child exclusion: %v", err)
+		return excluded
+	}
+
+	// Index links by their interface index so we can resolve MasterIndex
+	// lookups in O(1) and walk the hierarchy without re-querying netlink.
+	linksByIndex := make(map[int]netlink.Link, len(links))
+	nameByIndex := make(map[int]string, len(links))
+	for _, l := range links {
+		idx := l.Attrs().Index
+		linksByIndex[idx] = l
+		nameByIndex[idx] = l.Attrs().Name
+	}
+
+	excludedIndices := sets.New[int]()
+	for idx, name := range nameByIndex {
+		if excluded.Has(name) {
+			excludedIndices.Insert(idx)
+		}
+	}
+
+	// Expand the excluded set by repeatedly walking the link list and adding
+	// any link whose master is already excluded. Stop once a full pass adds
+	// nothing new (handles nested hierarchies like vf -> vf-child -> uplink).
+	for {
+		added := false
+		for _, l := range links {
+			attrs := l.Attrs()
+			if excluded.Has(attrs.Name) {
+				continue
+			}
+			if attrs.MasterIndex == 0 {
+				continue
+			}
+			if excludedIndices.Has(attrs.MasterIndex) {
+				excluded.Insert(attrs.Name)
+				excludedIndices.Insert(attrs.Index)
+				added = true
+			}
+		}
+		if !added {
+			break
+		}
+	}
+
+	return excluded
+}
+
 func getTcFilters(link netlink.Link) ([]string, bool) {
 	isTcEBPF := false
 	filterNames := sets.Set[string]{}
--- a/pkg/inventory/net_test.go
+++ b/pkg/inventory/net_test.go
@@ -317,6 +317,30 @@ func testGetExcludedUplinkInterfaces_Namespaced(t *testing.T) {
 			},
 			expectedResult: sets.New[string]("br0"),
 		},
+		{
+			// macvlan links to its lower device through ParentIndex, not
+			// MasterIndex. It carries its own forwarding state and can be
+			// relocated into a pod netns without stranding the host uplink,
+			// so the exclusion walk (which follows MasterIndex) must leave
+			// it allocatable even when the parent is the default-gw uplink.
+			name: "Macvlan child of uplink stays allocatable",
+			setup: func(t *testing.T) {
+				uplink := addDummyUplink(t, "eth0", bridgeAddr, defaultIPv4, gwIPv4)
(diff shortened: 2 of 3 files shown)


bowei:
@gauravkghildiyal

tamilmani1989 (approved):

pkg/inventory/net.go
```diff
@@ -115,6 +115,62 @@ func getDefaultGwInterfaces() sets.Set[string] {
...
+	excludedIndices := sets.New[int]()
+	for idx, name := range nameByIndex {
+		if excluded.Has(name) {
+			excludedIndices.Insert(idx)
+		}
+	}
+
+	// Expand the excluded set by repeatedly walking the link list and adding
+	// any link whose master is already excluded. Stop once a full pass adds
+	// nothing new (handles nested hierarchies like vf -> vf-child -> uplink).
```
rbtr: If we walk the link list once and build a `map[id][]link` of parent -> children links, then we only need to do one (mutating) pass of excluded that culls the whole families. Simplified example:

```go
for i := 0; i < len(excluded); i++ {
  parentID := excluded[i].id
  if children, found := childrenOf[parentID]; found {
    // Move them all to the excluded slice
    excluded = append(excluded, children...)
    // Remove from map so we don't process the same children twice if the ID appears again for some reason.
    delete(childrenOf, parentID)
  }
}
```

It should be more like O(N+M) instead of O(N×M).

gauravkghildiyal:
Thanks for the change @anson627 

Can you please share some more details on the issue that leads to DRANET failing to move this interface to the pod's network namespace? Your linked issue includes this error message `link not found for interface enP63016s1` and I'm trying to understand why the "link could not be found".

The behaviour that we ideally want is that, for the devices that DRANET discovers and advertises, it should be able to ALL of them allocated to a pod. 

1. DRANET should NOT advertise devices which it knows for a fact it will fail to allocate to the pod. 
2. There can be devices that DRANET successfully discovers and can even allocate to pod, but for various reasons we don't want them to be available to pods and thus wish to exclude them from ResourceSlice. The case of "default gateway interfaces" which is excluded already is a special one and falls under this category. We don't want "default gateway interfaces" to be dedicated to a pod because they are usually shared by all pods on the Node. And I'm hoping for cloud providers to be able to further extend this category through https://github.com/kubernetes-sigs/dranet/issues/152 for their own specific needs.

I'm guessing you are saying the case of this PR falls under category number 1, and thus it'd be nice for us to verify why this failure is generic and why we can claim it will always fail in any platform.

anson627:
> Thanks for the change @anson627
> 
> Can you please share some more details on the issue that leads to DRANET failing to move this interface to the pod's network namespace? Your linked issue includes this error message `link not found for interface enP63016s1` and I'm trying to understand why the "link could not be found".
> 
> The behaviour that we ideally want is that, for the devices that DRANET discovers and advertises, it should be able to ALL of them allocated to a pod.
> 
> 1. DRANET should NOT advertise devices which it knows for a fact it will fail to allocate to the pod.
> 2. There can be devices that DRANET successfully discovers and can even allocate to pod, but for various reasons we don't want them to be available to pods and thus wish to exclude them from ResourceSlice. The case of "default gateway interfaces" which is excluded already is a special one and falls under this category. We don't want "default gateway interfaces" to be dedicated to a pod because they are usually shared by all pods on the Node. And I'm hoping for cloud providers to be able to further extend this category through [Improving Device Discovery and Filtering Control in DRANET #152](https://github.com/kubernetes-sigs/dranet/issues/152) for their own specific needs.
> 
> I'm guessing you are saying the case of this PR falls under category number 1, and thus it'd be nice for us to verify why this failure is generic and why we can claim it will always fail in any platform.



> Thanks for the change @anson627
> 
> Can you please share some more details on the issue that leads to DRANET failing to move this interface to the pod's network namespace? Your linked issue includes this error message `link not found for interface enP63016s1` and I'm trying to understand why the "link could not be found".
> 
> The behaviour that we ideally want is that, for the devices that DRANET discovers and advertises, it should be able to ALL of them allocated to a pod.
> 
> 1. DRANET should NOT advertise devices which it knows for a fact it will fail to allocate to the pod.
> 2. There can be devices that DRANET successfully discovers and can even allocate to pod, but for various reasons we don't want them to be available to pods and thus wish to exclude them from ResourceSlice. The case of "default gateway interfaces" which is excluded already is a special one and falls under this category. We don't want "default gateway interfaces" to be dedicated to a pod because they are usually shared by all pods on the Node. And I'm hoping for cloud providers to be able to further extend this category through [Improving Device Discovery and Filtering Control in DRANET #152](https://github.com/kubernetes-sigs/dranet/issues/152) for their own specific needs.
> 
> I'm guessing you are saying the case of this PR falls under category number 1, and thus it'd be nice for us to verify why this failure is generic and why we can claim it will always fail in any platform.

Thanks for confirming on this! To answer your questions:

1. it's **category 1 (DRANET should not advertise what it cannot allocate)**, and the failure  isn't actually Azure-specific — AKS just happens to be where it surfaces because of how the mlx5 "accelerated networking" topology is laid out.

**What's happening on the failing node:**

- `eth0` is the default-gateway uplink (a synthetic netvsc device).
- `enP63016s1` is an [SR-IOV VF](https://learn.microsoft.com/en-us/windows-hardware/drivers/network/overview-of-single-root-i-o-virtualization--sr-iov-) enslaved to `eth0` — confirmed by ifconfig on the host
```
  enP63016s1 Link encap:Ethernet  HWaddr 7C:ED:8D:6D:BB:4C
            UP BROADCAST RUNNING SLAVE MULTICAST  MTU:1500  Metric:1
            RX packets:4407116 errors:0 dropped:2295 overruns:0 frame:0
            TX packets:4292578 errors:0 dropped:0 overruns:0 carrier:0
            collisions:0 txqueuelen:1000
            RX bytes:2798331313 (2.6 GiB)  TX bytes:2004276229 (1.8 GiB)

  eth0      Link encap:Ethernet  HWaddr 7C:ED:8D:6D:BB:4C
            inet addr:10.224.0.91  Bcast:10.224.255.255  Mask:255.255.0.0
            inet6 addr: fe80::7eed:8dff:fe6d:bb4c/64 Scope:Link
            UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
            RX packets:3633539 errors:0 dropped:0 overruns:0 frame:0
            TX packets:3471629 errors:0 dropped:0 overruns:0 carrier:0
            collisions:0 txqueuelen:1000
            RX bytes:2700172620 (2.5 GiB)  TX bytes:1949772777 (1.8 GiB)
```
Noticed `eth0` and `enP63016s1` have the same `HWaddr`

**Why I believe this generalizes beyond Azure:** anything built on the same "synthetic master + SR-IOV VF slave" pattern has the same problem — bonded uplinks with enslaved VFs, teamd/bond failover setups, and any `IFLA_MASTER`-linked child of an interface we already refuse to hand out. The signal we're filtering on is purely the Linux link hierarchy (`MasterIndex` / `IFLA_MASTER`), not an Azure heuristic, so a bonded baremetal node with a default route over `bond0` and VF child would be filtered the same way.

anson627:
While this PR only address the part of "drop netdev" for uplink child, I think the `ProcessDevices` extension point you sketched in #152 is exactly the right place to land the part of "keep RDMA" for uplink children.
The flow would be:

1. Discovery still picks up `enP63016s1` and its RDMA identity (`mlx5_8` via `/sys/class/net/enP63016s1/device/infiniband/mlx5_8`).
2. The generic `ProcessDevices` hook (or, in @dkennetzoracle's framing, the combined `AddAttributes` + `FilterDevices` contract) gets invoked with the device list.
3. DRANET's own default rule (or a cloud-provider implementation, e.g. Azure) recognizes "this device is a child of an excluded uplink AND `dra.net/rdma == true`" and rewrites the device in place: delete `dra.net/ifName`, keep `pciAddress`, `dra.net/rdma`, `dra.net/rdmaDevice`.
4. The resulting ResourceSlice entry is allocatable only via the existing RDMA-only / IB-only path, so `attachNetdevToNS()` is never invoked for it and `RunPodSandbox` can't fail on the "link not found" move.

For the scope of this PR, I'd still like to merge the netdev exclusion as-is (it's the safety fix that prevents users from falling into trap today), and then layer the RDMA-only publishing on top once #152's extension point is agreed on. Does that sequencing work for you?

pkg/inventory/net.go
anson627: make sense, updated to address the comment

gauravkghildiyal:
Thanks for the details @anson627 

So it seems like what we are saying is that whenever we have a parent-child relationship through a `MasterIndex`, it is _**NEVER**_ a good idea to move just the child interfaces to a different network namespace. In our case, this means for all default gateway interfaces acting as a parent it is never correct to move just their children (through `MasterIndex`) to a pod. I think that makes sense. @anson627 can you please update the comment in `getExcludedUplinkInterfaces()` to clearly state our rationale?

I will approve this but I'll request @aojea for another pair of eyes to make sure I didn't miss any obvious potential future use case. 

/approve
/hold 
/assign

anson627:
> Thanks for the details @anson627
> 
> So it seems like what we are saying is that whenever we have a parent-child relationship through a `MasterIndex`, it is _**NEVER**_ a good idea to move just the child interfaces to a different network namespace. In our case, this means for all default gateway interfaces acting as a parent it is never correct to move just their children (through `MasterIndex`) to a pod. I think that makes sense. @anson627 can you please update the comment in `getExcludedUplinkInterfaces()` to clearly state our rationale?
> 
> I will approve this but I'll request @aojea for another pair of eyes to make sure I didn't miss any obvious potential future use case.
> 
> /approve /hold /assign

comments updated, @aojea please help to take another look

aojea:
/assign @aojea 

I want to check carefully , complex area this one

pkg/inventory/net_test.go
```diff
@@ -254,3 +254,171 @@ func testGetDefaultGwInterfaces_Namespaced(t *testing.T) {
 		})
 	}
 }
+
+func TestGetExcludedUplinkInterfaces(t *testing.T) {
+	userns.Run(t, testGetExcludedUplinkInterfaces_Namespaced, syscall.CLONE_NEWNET)
+}
+
+func testGetExcludedUplinkInterfaces_Namespaced(t *testing.T) {
```
aojea, line 262: can we add a ipvlan or macvlan testcase to validate those are not excluded?


IIUIC there is the `MasterIndex` that tells us the device is `Slave`, we do not want enslaved interfaces from the gateways to be included in the resource slices

There is the `ParentIndex` that tell us what is the `Upper device`, we do want the macvlan and ipvlan like interfaces from the gateway to be included in the resource slices

aojea:
learn a lot with this PR and the context, thanks for the good description and congrats to the azure folks documentation, really well explained the technical details of the VF and network acceleration.

Just one comment to avoid in the future possible regressions with macvlan and ipvlan interfaces https://github.com/kubernetes-sigs/dranet/pull/176/changes#r3139442015 the rest LGTM

pkg/inventory/net_test.go
anson627: added unit test to verify that we exclude only the MasterIndex-linked descendants of the default-gateway uplink, while ParentIndex-linked children remain allocatable, i.e. macvlan/ipvlan interfaces backed by the gateway uplink are still included in the resource slices.

anson627:
> learn a lot with this PR and the context, thanks for the good description and congrats to the azure folks documentation, really well explained the technical details of the VF and network acceleration.
> 
> Just one comment to avoid in the future possible regressions with macvlan and ipvlan interfaces https://github.com/kubernetes-sigs/dranet/pull/176/changes#r3139442015 the rest LGTM

Thanks for the thorough review! I’ve updated the code to address your comments. please take another look when you have a chance.

aojea:
/lgtm
/approve
/hold cancel

Thanks

anson627:
looks like github is degraded again and taking longer than usual: https://damrnelson.github.io/github-historical-uptime/

PR #180, opened 2026-04-28
Bug Fix: RDMA device is not restored to the host network namespace in rdma netns=exclusive mode

### LINKED ISSUE #177: RDMA device is not restored to the host network namespace in rdma netns=exclusive mode
# Description:

In rdma netns=exclusive mode, the RDMA device is not properly restored to the host network namespace after a pod is deleted.
While PF/VF netdevices are correctly returned to the host network namespace, the RDMA device is not.

# Environment:

* Kubernetes: v1.35.3
* containerd: v2.2.2
* runc: v1.4.2
* RDMA NIC: Mellanox ConnectX-6

kanlkan:
#### What type of PR is this?
/kind bug

* nsDetachRdmadev was querying the host RDMA subsystem instead of the container's

  * Pass unix.NETLINK_RDMA explicitly to nlwrap.NewHandleAt(containerNs, unix.NETLINK_RDMA)

* Reorder RDMA-before-netdev in stopPodSandbox

  * nsDetachNetdev's LinkSetUp on the host VF emits RTM_NEWLINK, which causes the inventory to rescan. If the RDMA device is still in the pod namespace at that point, it is not re-detected. Returning the RDMA device first avoids this race.

* File descriptor / netlink socket leak in RDMA attach/detach

  * Add defer containerNs.Close() in both nsAttachRdmadev and nsDetachRdmadev
  * Add defer nhNs.Close() in nsDetachRdmadev
  * One fd leaked per pod start/stop, accumulating over the daemon lifetime.

* Re-detect RDMA capability promptly after pod removal

  * There are paths where an RDMA device returns to init_net without an accompanying NEWLINK event (e.g. devices without an associated netdev, or kernel netns teardown after a failed nsDetachRdmadev). In those cases the device would not be re-detected until the next periodic scan (up to ~1 minute).
  * Add rescanCh (capacity-1 channel) and RequestRescan() on inventory.DB, plumb it through the inventoryDB interface, and call it from removePodSandbox so the inventory rescans immediately on pod teardown.

Fixes #177 

None

```release-note
NOTE
```


pkg/driver/nri_hooks.go
```diff
@@ -417,6 +422,15 @@ func (np *NetworkDriver) RemovePodSandbox(ctx context.Context, pod *api.PodSandb
...
 	np.netdb.RemovePodNetNs(podKey(pod))
+	// The inventory only subscribes to NETLINK_ROUTE link events, not
+	// NETLINK_RDMA. For SR-IOV VFs the netdev's NEWLINK (emitted by
+	// nsDetachNetdev or, on failure, by kernel netns cleanup) drives the
+	// rescan that re-detects RDMA via sysfs. For IB-only RDMA devices there
+	// is no associated netdev, so no NEWLINK is ever emitted on return to
+	// init_net — neither on success nor on failure of nsDetachRdmadev.
+	// Request a rescan so RDMA capability is re-detected immediately rather
+	// than waiting up to maxPollInterval (~1 minute).
+	np.netdb.RequestRescan()
```
anson627: The `RequestRescan()` implementation looks good. My concern is the call site: `RemovePodSandbox()` returns early when `podConfigStore.GetPodConfig()` is missing, so `removePodSandbox()` and `RequestRescan()` are skipped.

If `UnprepareResourceClaims()`/`DeleteClaim()` runs before the NRI remove hook, a pod that previously had a RDMA device can have its config removed before `RemovePodSandbox()` runs. In that ordering there may be no `NEWLINK` event and no explicit `RequestRescan()`, so rediscovery still waits for the periodic poll.

Can we either request the rescan before the early return, or confirm that this teardown ordering cannot happen?

tamilmani1989: if rescan an expensive operation,  should we do only for IB only rdma device?

dkennetzoracle: I'd recommend doing before the early return, and conditionally, right? No reason to always rescan.

kanlkan:
@anson627 @dkennetzoracle @tamilmani1989 
Thanks for the feedbacks. 

Reworked the change as follows:

* Moved the rescan call from `RemovePodSandbox` into `stopPodSandbox,` where `PodConfig` is still available and we know per-device detach outcomes. `RemovePodSandbox` is back to its previous shape.

* Gated the rescan with a pure helper `needsRescanAfterDetach(rdmaDetached, netdevDetached) bool`. The condition is `rdmaDetached && !netdevDetached` — fire only when RDMA was returned to init_net but no netdev NEWLINK was emitted to act as an implicit rescan trigger. That covers IB-only devices (no netdev to detach) and SR-IOV pods whose `nsDetachNetdev` failed before LinkSetUp could emit NEWLINK; successful SR-IOV pods continue to rely on the existing NEWLINK-driven rescan path.

* Trade-off: when `nsDetachRdmadev` itself fails, the device is still in the pod namespace, so a rescan at that moment would not observe any state change. Recovery via kernel netns cleanup is picked up by the inventory's periodic scan — same behavior as before this PR.

* Tests:

  * `TestNeedsRescanAfterDetach` -  pure-function unit test covering the four boolean combinations.
  * `TestStopPodSandboxRescanGating` -  integration test verifying that no rescan fires in the negative cases (no PodConfig, host-network pod, shared RDMA mode, detach failure against fake netns). The positive path cannot be exercised against synthetic netns paths, which is why the gating logic was extracted into a pure function.

Let me know if you'd prefer different coverage or a different placement.

kanlkan:
Sorry, Bats e2e tests failed. I will check it.

pkg/driver/nri_hooks.go
kanlkan: Thanks for the review. Please take a look at my comment below.
https://github.com/kubernetes-sigs/dranet/pull/180#issuecomment-4350242382

kanlkan: Thanks for the review. Please take a look at my comment below.
https://github.com/kubernetes-sigs/dranet/pull/180#issuecomment-4350242382

kanlkan:
@anson627 @MikeZappa87 
> Sorry, Bats e2e tests failed. I will check it.

This appears to be CI flakiness in kubectl wait racing the deployment pod creation; failing tests are unrelated to RDMA paths. Could you please re-run this test?

anson627:
thanks for addressing the comments!

/ok-to-test

kanlkan:
I created an empty commit for trigger to test.

anson627:
/lgtm

I will run some test of creating/deleting pods to verify

pkg/inventory/db.go
```diff
@@ -212,6 +214,7 @@ func (db *DB) Run(ctx context.Context) error {
 			for len(nlChannel) > 0 {
 				<-nlChannel
 			}
+		case <-db.rescanCh:
```
gauravkghildiyal, line 217: nit: Can we add a log line here, something like:

```
klog.V(3).Infof("Triggering inventory rescan due to manual request")
```

pkg/driver/nri_hooks.go
```diff
@@ -378,23 +378,59 @@ func (np *NetworkDriver) stopPodSandbox(_ context.Context, pod *api.PodSandbox,
...
 	for deviceName, config := range podConfig.DeviceConfigs {
+		// Move the RDMA device back to the host namespace BEFORE the netdev.
+		// nsDetachNetdev calls LinkSetUp on the VF in the host namespace, which
+		// triggers a NEWLINK event causing the inventory to rescan. If the RDMA
+		// device is still in the pod namespace at that point it will not be
+		// detected, so it must be returned first.
+		rdmaDetached := false
+		if !np.rdmaSharedMode && config.RDMADevice.LinkDev != "" {
+			if err := nsDetachRdmadev(ns, config.RDMADevice.LinkDev); err != nil {
+				klog.Infof("fail to return rdma device %s : %v", deviceName, err)
```
gauravkghildiyal: nit: I know this log was at Infof level even before but since we're making a change, could you please switch this to Error level to correctly reflect an error?

gauravkghildiyal (commented):
Thanks @kanlkan. This looks great.

Just two non blocking requests if you can find the time.

pkg/inventory/db.go
kanlkan: I fixed it in commit 7f7b601. Could you please review it again?

pkg/driver/nri_hooks.go
kanlkan: I fixed it in commit 7f7b601. Could you please review it again?

gauravkghildiyal:
Thanks @kanlkan 

/lgtm
/approve

PR #185, opened 2026-05-01
Update release process doc and improve release image tagging

gauravkghildiyal:
#### What type of PR is this?

/kind cleanup

* Cleanup the release process doc.
* When releasing images, make sure they use proper tag (e.g. 1.23) rather than _always_ using the dated format.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,6 @@ image-push: ensure-buildx
 		--push
 
 helm-package: ensure-helm
-	mkdir -p $(OUT_DIR)
 	helm package deployments/helm/dranet \
 		--version "$(CHART_VERSION)" \
 		--app-version "$(TAG)" \
--- a/Makefile
+++ b/Makefile
@@ -58,18 +58,14 @@ ensure-helm:
 		go install helm.sh/helm/v3/cmd/helm@$(HELM_VERSION_SHA); \
 	fi
 
-# get image name from directory we're building
 IMAGE_NAME=dranet
-# docker image registry, default to upstream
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -96,11 +92,11 @@ image-push: ensure-buildx
 		--tag="${IMAGE}" \
 		--push
 
-helm-package:
-	@test -n "$(HELM_TAG)" || (echo "ERROR: not on an exact git tag, cannot package helm chart"; exit 1)
+helm-package: ensure-helm
+	mkdir -p $(OUT_DIR)
```
fmuyassarov: I think `--destination $(OUT_DIR)` will take care of this if the `OUT_DIR` directory doesn't exit

fmuyassarov (commented):
Thanks @gauravkghildiyal . I have a single comment otherwise LGTM

Makefile
gauravkghildiyal: Sounds valid. Removed mkdir. Thanks @fmuyassarov

fmuyassarov: No strong opinion on commit history policy here happy to defer to whatever is preferred. That said, would it make sense to squash the latest commit into the https://github.com/kubernetes-sigs/dranet/pull/185/changes/e907b2adb40c4d0b7241f56f1056dca335a35da3?  Please feel free to unhold and merge as-is if you'd rather keep the history as-is.
/lgtm
/hold

gauravkghildiyal: That's okay, we can just squash before merging. Thanks for the review!

/unhold

PR #187, opened 2026-05-01
Mark NICs with shared IOMMU groups or default routes as vfio-unsafe

### LINKED ISSUE #186: VFIO passthrough fails for NICs with shared IOMMU groups or default gateway
**What happened:**

When using VFIO mode for VM passthrough, dranet publishes all discovered NICs in the ResourceSlice without indicating which ones are safe for VFIO. The scheduler may allocate a NIC that fails at VFIO bind time or causes host connectivity loss.

Two failure modes:

1. **Shared IOMMU group** — QEMU reports: `vfio 0000:02:00.1: group 75 is not viable. Please ensure all devices within the iommu_group are bound to their vfio bus driver.`

2. **Default gateway interface** — unbinding the NIC that carries the host's default route kills SSH, kubectl, and all host management connectivity.

**Feature request:**

Add a `dra.net/vfioUnsafe` device attribute so claims can exclude NICs that are unsafe for VFIO passthrough, while still allowing them for non-VFIO use cases (bridge, macvlan).

**Environment:**
- Dell XE8640 with Broadcom `0000:02:00.1` sharing IOMMU group 75 with management port `0000:02:00.0` (tg3 driver)

/kind feature

johnahull:
#### What type of PR is this?

/kind feature

Adds `dra.net/vfioUnsafe` and `dra.net/vfioUnsafeReason` device attributes to NICs that cannot safely be used for VFIO passthrough:

- **Shared IOMMU group** — a sibling PCI device in the same IOMMU group is bound to a different driver. VFIO requires all devices in the group to be on vfio-pci, otherwise the kernel rejects the bind ("group is not viable").
- **Default gateway interface** — unbinding the NIC carrying the host's default route kills host connectivity.

Devices are annotated rather than removed so non-VFIO use cases (bridge, macvlan) still work. VM claims can exclude unsafe devices with a CEL selector:

```
!(has(device.attributes["dra.net"].vfioUnsafe) && device.attributes["dra.net"].vfioUnsafe)
```

Reads `/proc/1/net/route` for default gateway detection (accesses host PID 1's network namespace when running inside a container).

Fixes https://github.com/kubernetes-sigs/dranet/issues/186

Observed on Dell XE8640: Broadcom `0000:02:00.1` shares IOMMU group 75 with management port `0000:02:00.0` (tg3 driver). Without this filter, the scheduler allocates the NIC, dranet binds it to vfio-pci, and QEMU fails with "vfio 0000:02:00.1: group 75 is not viable."

The annotation approach was chosen over filtering (removing devices) because the same NIC may be perfectly usable for non-VFIO modes like bridge or macvlan.

```release-note
Add dra.net/vfioUnsafe attribute to mark NICs with shared IOMMU groups or default gateway routes as unsafe for VFIO passthrough.
```

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -60,6 +60,7 @@ func (np *NetworkDriver) PublishResources(ctx context.Context) {
 		case devices := <-np.netdb.GetResources(ctx):
 			klog.V(3).Infof("Got %d devices from inventory: %s", len(devices), formatDeviceNames(devices, 15))
 			devices = filter.FilterDevices(np.celProgram, devices)
+			devices = filter.MarkVFIOUnsafe(devices)
 			klog.V(3).Infof("After filtering, publishing %d devices in ResourceSlice(s): %s", len(devices), formatDeviceNames(devices, 15))
 
 			np.publishResourcesPrometheusMetrics(devices)
--- a/pkg/filter/vfio_safety.go
+++ b/pkg/filter/vfio_safety.go
@@ -1,0 +1,171 @@
+/*
+Copyright The Kubernetes Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


johnahull:
Closing — this filter is only relevant with VFIO passthrough mode, which is not being proposed upstream at this time.

PR #189, opened 2026-05-01
Fix postsubmit failure caused by Bats tests for metric server

ngcxy:
#### What type of PR is this?

/kind cleanup

The [periodic postsubmit test](https://github.com/kubernetes-sigs/dranet/actions/workflows/periodics.yaml) has been failing for the Bats test [test metric server is up and operating on host](https://github.com/kubernetes-sigs/dranet/blob/main/tests/e2e.bats#L181). It is caused by a race condition where kubectl run -i terminates before streaming the output. This PR fixes it by replacing `kubectl run -i` with wait and log.

N/A

### DIFF
--- a/tests/e2e.bats
+++ b/tests/e2e.bats
@@ -179,14 +179,17 @@ setup_tcx_filter() {
 }
 
 @test "test metric server is up and operating on host" {
-  output=$(kubectl \
-    run -i test-metrics \
+  # Run a temporary pod to access metrics
+  kubectl run test-metrics \
     --image registry.k8s.io/e2e-test-images/agnhost:2.54 \
     --overrides='{"spec": {"hostNetwork": true}}' \
     --restart=Never \
     --command \
-    -- sh -c "curl --silent localhost:9177/metrics | grep process_start_time_seconds >/dev/null && echo ok || echo fail")
-  assert_equal "$output" "ok"
+    -- sh -c "curl --silent localhost:9177/metrics | grep process_start_time_seconds >/dev/null && echo ok || echo fail"
+
+  # Wait for completion and verify output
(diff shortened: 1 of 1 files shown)


tests/e2e.bats
```diff
@@ -179,14 +179,17 @@ setup_tcx_filter() {
...
-    run -i test-metrics \
+  # Run a temporary pod to access metrics
+  kubectl run test-metrics \
     --image registry.k8s.io/e2e-test-images/agnhost:2.54 \
     --overrides='{"spec": {"hostNetwork": true}}' \
     --restart=Never \
     --command \
-    -- sh -c "curl --silent localhost:9177/metrics | grep process_start_time_seconds >/dev/null && echo ok || echo fail")
-  assert_equal "$output" "ok"
+    -- sh -c "curl --silent localhost:9177/metrics | grep process_start_time_seconds >/dev/null && echo ok || echo fail"
```
michaelasp, line 188: Once a pod goes into the completed state, can we rely on the logs to always be accessible? This is still much better than before but I'm curious why kubectl -i terminates before the output gets sent to the command. I'm not normally pro sleeps, but I wonder if adding one here may give us a buffer to receive the logs.

ngcxy: Thank you for the insight @michaelasp ! I tried adding the sleep buffer, but unfortunately `kubectl run -i` still returned empty to the output. It seems not to be transmission latency issue, but the execution limitation of kubectl run.
I believe we can reply on the log after pod goes into completion, as long as we retrieve it right after the execution and the pod is not deleted.

michaelasp: Ack, lgtm then!

PR #191, opened 2026-05-05
Consolidate pod-level state into podConfigStore

purvavj:
#### What type of PR is this?
/kind cleanup

Consolidates pod-level state by moving the network namespace (netns) storage from the inventory database (pkg/inventory/db.go) into PodConfigStore (pkg/driver/pod_device_config.go).
This ensures all pod-level metadata is managed in a single unified store and decouples the inventory DB to focus solely on host device discovery.

N/A

Consolidated pod network namespace (netns) tracking completely into PodConfigStore.
Optimized Synchronize to only store netns for pods with active device configurations.
Added NetNs unit tests in pod_device_config_test.go and a new Synchronize behavior test in nri_hooks_test.go to verify the new state logic.

```release-note
NONE
```

### DIFF
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -56,9 +56,7 @@ func (np *NetworkDriver) Synchronize(_ context.Context, pods []*api.PodSandbox, 
 		ns, isLive := livePodNetNs[storedUID]
 
 		if isLive {
-			if ns != "" {
-				np.podConfigStore.SetPodNetNs(storedUID, ns)
-			}
+			np.podConfigStore.SetPodNetNs(storedUID, ns)
 		} else {
 			klog.Infof("Synchronize: pruning stale config for pod %s", storedUID)
 			np.podConfigStore.DeletePod(storedUID)
@@ -370,18 +368,16 @@ func (np *NetworkDriver) stopPodSandbox(_ context.Context, pod *api.PodSandbox, 
 		// we workaround it using the local copy we have in the db to associate interfaces with Pods via
 		// the network namespace id.
 		storedNs, ok := np.podConfigStore.GetPodNetNs(types.UID(pod.GetUid()))
-		if ok {
-			if storedNs != "" {
-				ns = storedNs
-			} else {
-				// Pod is not configured for DRAnet (host network or other driver)
-				klog.Infof("StopPodSandbox pod %s/%s using host network ... skipping", pod.Namespace, pod.Name)
-				return nil
-			}
-		} else {
+		if !ok {
 			klog.Warningf("StopPodSandbox: pod %s/%s (UID %s) not found in podConfigStore when fetching fallback NetNS", pod.Namespace, pod.Name, pod.Uid)
 			return nil
 		}
+		if storedNs == "" {
+			// Pod is not configured for DRAnet (host network or other driver)
+			klog.Infof("StopPodSandbox pod %s/%s using host network ... skipping", pod.Namespace, pod.Name)
+			return nil
+		}
+		ns = storedNs
 	}
 	needsRescan := false
 	for deviceName, config := range podConfig.DeviceConfigs {
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -367,17 +367,11 @@ func (np *NetworkDriver) stopPodSandbox(_ context.Context, pod *api.PodSandbox, 
 		// some version of containerd does not send the network namespace information on this hook so
 		// we workaround it using the local copy we have in the db to associate interfaces with Pods via
 		// the network namespace id.
-		storedNs, ok := np.podConfigStore.GetPodNetNs(types.UID(pod.GetUid()))
-		if !ok {
-			klog.Warningf("StopPodSandbox: pod %s/%s (UID %s) not found in podConfigStore when fetching fallback NetNS", pod.Namespace, pod.Name, pod.Uid)
+		if podConfig.NetNS == "" {
+			klog.Warningf("StopPodSandbox: network namespace for DRANET pod %s/%s (UID %s) is unknown; skipping explicit device detach and relying on kernel netns teardown", pod.Namespace, pod.Name, pod.Uid)
 			return nil
 		}
-		if storedNs == "" {
-			// Pod is not configured for DRAnet (host network or other driver)
-			klog.Infof("StopPodSandbox pod %s/%s using host network ... skipping", pod.Namespace, pod.Name)
-			return nil
-		}
-		ns = storedNs
+		ns = podConfig.NetNS
 	}
 	needsRescan := false
 	for deviceName, config := range podConfig.DeviceConfigs {
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -42,24 +42,24 @@ func (np *NetworkDriver) Synchronize(_ context.Context, pods []*api.PodSandbox, 
 	klog.Infof("Synchronized state with the runtime (%d pods, %d containers)...",
 		len(pods), len(containers))
 
-	livePods := set.New[types.UID]()
+	// livePodNetNs map tracks live pods by UID and their network namespace paths.
+	livePodNetNs := make(map[types.UID]string)
 	for _, pod := range pods {
 		klog.Infof("Synchronize Pod %s/%s UID %s", pod.Namespace, pod.Name, pod.Uid)
 		klog.V(2).Infof("pod %s/%s: namespace=%s ips=%v", pod.GetNamespace(), pod.GetName(), getNetworkNamespace(pod), pod.GetIps())
-		livePods.Insert(types.UID(pod.Uid))
-		// get the pod network namespace
-		ns := getNetworkNamespace(pod)
-		// host network pods are skipped
-		if ns != "" {
-			// store the Pod metadata in the db
-			np.netdb.AddPodNetNs(podKey(pod), ns)
-		}
+		livePodNetNs[types.UID(pod.Uid)] = getNetworkNamespace(pod)
 	}
 
-	// Prune persisted configs for pods that no longer exist in the runtime.
-	// This handles the case where pods were deleted while the driver was down.
+	// Process stored pods: update NetNS for live pods, and prune configurations
+	// for pods that no longer exist in the runtime.
 	for _, storedUID := range np.podConfigStore.ListPods() {
-		if !livePods.Has(storedUID) {
+		ns, isLive := livePodNetNs[storedUID]
+
+		if isLive {
+			if ns != "" {
+				np.podConfigStore.SetPodNetNs(storedUID, ns)
+			}
+		} else {
 			klog.Infof("Synchronize: pruning stale config for pod %s", storedUID)
 			np.podConfigStore.DeletePod(storedUID)
 		}
@@ -152,8 +152,8 @@ func (np *NetworkDriver) runPodSandbox(_ context.Context, pod *api.PodSandbox, p
 	if ns == "" {
 		return fmt.Errorf("RunPodSandbox pod %s/%s using host network can not claim host devices", pod.Namespace, pod.Name)
 	}
-	// store the Pod metadata in the db
-	np.netdb.AddPodNetNs(podKey(pod), ns)
+	// store the Pod network namespace in the pod config store
+	np.podConfigStore.SetPodNetNs(types.UID(pod.GetUid()), ns)
 
 	// Track all the status updates needed for the resource claims of the pod.
 	statusUpdates := map[types.NamespacedName]*resourceapply.ResourceClaimStatusApplyConfiguration{}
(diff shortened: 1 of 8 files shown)


pkg/driver/nri_hooks.go
```diff
@@ -42,26 +42,23 @@ func (np *NetworkDriver) Synchronize(_ context.Context, pods []*api.PodSandbox,
...
+	// for pods that no longer exist in the runtime.
 	for _, storedUID := range np.podConfigStore.ListPods() {
-		if !livePods.Has(storedUID) {
+		ns, isLive := livePods[storedUID]
+
+		if !isLive {
 			klog.Infof("Synchronize: pruning stale config for pod %s", storedUID)
 			np.podConfigStore.DeletePod(storedUID)
+		} else if ns != "" {
+			np.podConfigStore.SetPodNetNs(storedUID, ns)
```
tamilmani1989: can there be a case where pod is alive but ns is nil? so SetPodNetNs  wont be called and ns wont be set in pod config store

purvavj: There can be cases (like host-network or transient pods), but they aren't relevant here. For any pod tracked in our store, it can never be alive and have ns == "".

pkg/driver/nri_hooks.go
```diff
@@ -42,26 +42,23 @@ func (np *NetworkDriver) Synchronize(_ context.Context, pods []*api.PodSandbox,
...
 		}
```
rbtr: for a single conditional, it's okay, but once there is branching I find leading with inverted checks hard to follow - 
```suggestion
		if isLive {
			if ns != "" {
				np.podConfigStore.SetPodNetNs(storedUID, ns)
			}
		} else {
			klog.Infof("Synchronize: pruning stale config for pod %s", storedUID)
			np.podConfigStore.DeletePod(storedUID)
		} 
```

pkg/driver/nri_hooks.go
```diff
@@ -42,26 +42,23 @@ func (np *NetworkDriver) Synchronize(_ context.Context, pods []*api.PodSandbox,
 	klog.Infof("Synchronized state with the runtime (%d pods, %d containers)...",
 		len(pods), len(containers))
 
-	livePods := set.New[types.UID]()
+	livePods := make(map[types.UID]string)
```
rbtr: not sure about this overloading of the `livePods` Set to create a pod:netns map. it at least needs to be renamed and documented, if not a separate new datastructure

pkg/driver/nri_hooks.go
```diff
@@ -363,16 +360,18 @@ func (np *NetworkDriver) StopPodSandbox(ctx context.Context, pod *api.PodSandbox
...
 		// some version of containerd does not send the network namespace information on this hook so
 		// we workaround it using the local copy we have in the db to associate interfaces with Pods via
 		// the network namespace id.
-		ns = np.netdb.GetPodNetNs(podKey(pod))
+		storedNs, ok := np.podConfigStore.GetPodNetNs(types.UID(pod.GetUid()))
+		if ok {
+			ns = storedNs
+		} else {
+			klog.Warningf("StopPodSandbox: pod %s/%s (UID %s) not found in podConfigStore when fetching fallback NetNS", pod.Namespace, pod.Name, pod.Uid)
+		}
```
rbtr: this seems partially redundant with the next few existing lines

pkg/driver/pod_device_config.go
```diff
@@ -253,9 +261,40 @@ func (s *PodConfigStore) GetPodConfig(podUID types.UID) (PodConfig, bool) {
...
+	defer s.mu.RUnlock()
+	podCfg, ok := s.configs[podUID]
+	if !ok {
+		return "", false
+	}
+	if podCfg.NetNS == "" {
+		return "", false
+	}
+	return podCfg.NetNS, true
+}
```
rbtr: why do I want to return `false` if the podCfg is `""`, even if it was found? Why isn't this just `return s.configs[podUID]`?

pkg/driver/nri_hooks.go
```diff
@@ -152,8 +149,8 @@ func (np *NetworkDriver) runPodSandbox(_ context.Context, pod *api.PodSandbox, p
 	if ns == "" {
 		return fmt.Errorf("RunPodSandbox pod %s/%s using host network can not claim host devices", pod.Namespace, pod.Name)
 	}
-	// store the Pod metadata in the db
-	np.netdb.AddPodNetNs(podKey(pod), ns)
+	// store the Pod network namespace in the pod config store
+	np.podConfigStore.SetPodNetNs(types.UID(pod.GetUid()), ns)
```
rbtr, line 154: this in-mem data doesn't survive restarts so this key change is okay?

purvavj: Yes, because Synchronize re-learns the in-mem state from the runtime. Since the UIDs are stable, we can easily match those live ns paths back to the config we have in the persistent store.

pkg/driver/pod_device_config.go
purvavj: It was originally kept to be sure that "" was handled safely, but I see we can simplify it here since we have other checks in place in the hooks.

pkg/driver/nri_hooks.go
purvavj: since the GetPodNs returned false for empty strings, which forced a second ns == "" check at the bottom. however, making GetPodNetNs to return the map's ok directly, we can clean things here.

pkg/driver/nri_hooks.go
```diff
@@ -363,18 +363,23 @@ func (np *NetworkDriver) StopPodSandbox(ctx context.Context, pod *api.PodSandbox
...
 	if ns == "" {
 		// some version of containerd does not send the network namespace information on this hook so
 		// we workaround it using the local copy we have in the db to associate interfaces with Pods via
 		// the network namespace id.
-		ns = np.netdb.GetPodNetNs(podKey(pod))
-		if ns == "" {
-			klog.Infof("StopPodSandbox pod %s/%s using host network ... skipping", pod.Namespace, pod.Name)
+		storedNs, ok := np.podConfigStore.GetPodNetNs(types.UID(pod.GetUid()))
+		if ok {
+			if storedNs != "" {
```
gauravkghildiyal: I think the code here could use some further simplification. Will recommend checking out https://maelvls.dev/go-happy-line-of-sight/ as a good practice

pkg/driver/pod_device_config.go
```diff
@@ -87,9 +91,11 @@ type LinuxDevice struct {
 // daemon restarts. Following the kubelet DRA checkpoint pattern
 // (pkg/kubelet/cm/dra/state), the in-memory PodConfigStore is the source
 // of truth and the Checkpointer is a write-through backend.
+// Note: Pod-level metadata like NetNS is not persisted (rebuilt on restart
+// via Synchronize() which queries the container runtime).
 type Checkpointer interface {
-	// GetOrCreate returns all persisted pod device configs, or an empty map
-	// if the checkpoint does not yet exist. Used at startup to restore state.
+	// GetOrCreate returns all persisted device configs. Used at startup to
```
gauravkghildiyal: Can you confirm that this behaviour isn't changing right? (If yes, then curious about the reason to update the comment)

pkg/driver/pod_device_config.go
```diff
@@ -253,9 +261,34 @@ func (s *PodConfigStore) GetPodConfig(podUID types.UID) (PodConfig, bool) {
...
+// SetPodNetNs stores the Pod's network namespace path in the pod-level config.
+// This is in-memory only; pod NetNS is rebuilt from the container runtime on
+// driver restart via Synchronize().
+func (s *PodConfigStore) SetPodNetNs(podUID types.UID, netns string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	podCfg, ok := s.configs[podUID]
+	if !ok {
+		klog.Warningf("SetPodNetNs: pod UID %s not found in store; skipping NetNS update", podUID)
```
gauravkghildiyal: This should be an Error, because we never expect this to happen

pkg/driver/nri_hooks.go
```diff
@@ -42,24 +42,24 @@ func (np *NetworkDriver) Synchronize(_ context.Context, pods []*api.PodSandbox,
...
-	// Prune persisted configs for pods that no longer exist in the runtime.
-	// This handles the case where pods were deleted while the driver was down.
+	// Process stored pods: update NetNS for live pods, and prune configurations
+	// for pods that no longer exist in the runtime.
 	for _, storedUID := range np.podConfigStore.ListPods() {
-		if !livePods.Has(storedUID) {
+		ns, isLive := livePodNetNs[storedUID]
+
+		if isLive {
+			if ns != "" {
```
gauravkghildiyal: I guess this check is not needed because we'll never have such a state?

gauravkghildiyal (commented):
Thanks @purvavj 

Seems like you need to rebase and fix the failing tests.

pkg/driver/pod_device_config.go
purvavj: Yes, no change in behavior just keeping it simple.

pkg/driver/nri_hooks.go
purvavj: Done. Removed the check and added a V(3) log in SetPodNetNs to aid in future debugging.

pkg/driver/nri_hooks.go
purvavj: yes, we can make the happy path left-aligned

pkg/driver/nri_hooks.go
```diff
@@ -363,20 +361,23 @@ func (np *NetworkDriver) StopPodSandbox(ctx context.Context, pod *api.PodSandbox
...
-	}()
 	// get the pod network namespace
 	ns := getNetworkNamespace(pod)
 	if ns == "" {
 		// some version of containerd does not send the network namespace information on this hook so
 		// we workaround it using the local copy we have in the db to associate interfaces with Pods via
 		// the network namespace id.
-		ns = np.netdb.GetPodNetNs(podKey(pod))
-		if ns == "" {
+		storedNs, ok := np.podConfigStore.GetPodNetNs(types.UID(pod.GetUid()))
```
gauravkghildiyal: I think you can directly get the network namespace from the functions podConfig argument, which is preferable over refetching the same information

pkg/driver/pod_device_config.go
```diff
@@ -35,6 +35,10 @@ type PodConfig struct {
 	// LastNRIActivity timestamp is updated whenever an NRI hook processes
 	// a container for this Pod. Used to track pod initialization progress.
 	LastNRIActivity time.Time
+
+	// NetNS is the path to the Pod's network namespace as observed by the
+	// container runtime.
+	NetNS string `json:"netns,omitempty"`
```
gauravkghildiyal: The json tag may not be useful given we are not persisting this.

pkg/driver/nri_hooks_test.go
```diff
@@ -565,7 +628,10 @@ func TestStopPodSandboxRescanGating(t *testing.T) {
 				np.podConfigStore.SetDeviceConfig(podUID, "eth0", tc.deviceConfig)
 			}
 			if tc.setupNetNs {
-				netdb.AddPodNetNs(podKey(pod), "/dummy/netns")
+				if !tc.setupDeviceConfig {
```
gauravkghildiyal: Thinking a bit about this, this feels a bit hacky to fix this test. 

Think about our actual implementation -- a netns is only set if the pod config exists. If the pod config doesn't exist, setnetns is never set. I think what would be better to do is to only setupNetNs when setupDeviceConfig is also true

```
			if tc.setupDeviceConfig && tc.setupNetNs {
				np.podConfigStore.SetPodNetNs(podUID, "/dummy/netns")
			}
```

pkg/driver/nri_hooks.go
```diff
@@ -363,20 +361,23 @@ func (np *NetworkDriver) StopPodSandbox(ctx context.Context, pod *api.PodSandbox
...
+		if !ok {
+			klog.Warningf("StopPodSandbox: pod %s/%s (UID %s) not found in podConfigStore when fetching fallback NetNS", pod.Namespace, pod.Name, pod.Uid)
+			return nil
+		}
+		if storedNs == "" {
+			// Pod is not configured for DRAnet (host network or other driver)
```
gauravkghildiyal: It seems like the existing log statement is misleading. Thinking about this, if we never stored a host network pod in the pod config store, we will never run into this situation. 

Even then if we run into this situation is where the namespace is empty for a pod in our store, that means this is a breakage in our invariant and we should log it with a Warning:

```
klog.Warningf("StopPodSandbox: network namespace for DRANET pod %s/%s (UID %s) is unknown; skipping explicit device detach and relying on kernel netns teardown", pod.Namespace, pod.Name, pod.Uid)
```

gauravkghildiyal (commented):
Thanks for the patience @purvavj. Just a few minor suggestions but looking good now.

pkg/driver/nri_hooks.go
purvavj: True, since any pod in our store must have devices (and is therefore a DRANET pod), an empty namespace here is indeed an anomaly rather than a normal host-network pod.

pkg/driver/nri_hooks_test.go
purvavj: yes, setting up the namespace when setupDeviceConfig is true will remove the dummy config creation and aligns the test with the actual driver behavior.

gauravkghildiyal:
Thanks!

/lgtm
/approve

PR #193, opened 2026-05-06
fix: Skip PCI network devices whose kernel driver is unbound

wevans-ant:
#### What type of PR is this?

/kind bug

**Symptom**

If a privileged workload unbinds a secondary NIC from its kernel driver while the device is allocated to it, dranet enters a permanent pod-trap on that node:

* `discoverPCIDevices()` still sees the PCI device (it never left the bus) and publishes it in the ResourceSlice with PCI-only attributes.
* The scheduler keeps allocating the device to new pods.
* `NodePrepareResources` for those pods fails every time with `device pci-… has no interface name in local store`, because there is no netdev for `GetNetInterfaceName()` to return.
* Pods sit forever in `Init` with `FailedPrepareDynamicResources`. Only a node reboot recovers.

**Minimal repro**

1. Schedule a privileged pod that holds a `dra.net` / `netdev.google.com` device claim.
2. Inside that pod: `echo 0000:00:09.0 > /sys/bus/pci/drivers/gve/unbind` (substitute the claimed device's BDF).
3. Delete the pod.
4. Schedule another pod with the same device class.

The new pod is permanently stuck:

```
Warning  FailedPrepareDynamicResources  ...  claim default/...: device pci-0000-00-09-0 has no interface name in local store
```

while the ResourceSlice still lists `pci-0000-00-09-0`:

```
dra_hooks.go:60] Got 4 devices from inventory: ..., pci-0000-00-09-0, pci-0000-00-0a-0
dra_hooks.go:62] After filtering, publishing 2 devices in ResourceSlice(s): pci-0000-00-09-0, pci-0000-00-0a-0
```

**Fix**

`ghw` already reports the bound kernel driver for each PCI device (via `readlink /sys/bus/pci/devices/$BDF/driver`). In `discoverPCIDevices()`, skip devices whose driver is either unbound (`""`) or one of the well-known non-netdev drivers (`vfio-pci`, `uio_pci_generic`, `igb_uio`, `pci-stub`). Those devices have no kernel netdev or RDMA link anywhere and can never be prepared, so they should not be advertised.

This changes the failure mode from "pod trap" to "device unschedulable": the bad device drops out of the ResourceSlice, the scheduler routes new claims to remaining devices on this or other nodes, and the node degrades by one NIC instead of becoming a black hole.

A device that has merely been moved into a pod's network namespace is **not** affected: its original NIC driver (`gve`, `mlx5_core`, …) stays bound, so it continues to be published with PCI-only attributes exactly as today. The Cluster Autoscaler template behaviour from https://github.com/google/dranet/issues/178 / https://github.com/google/dranet/pull/194 is preserved. IB-only RDMA devices likewise keep their NIC driver bound and are unaffected.

**Manual recovery (workaround for current releases)**

For anyone hitting this on an existing release, a privileged `hostNetwork` pod can rebind the driver without a reboot:

```sh
BDF=0000:00:09.0
echo gve > /sys/bus/pci/devices/$BDF/driver_override
echo $BDF > /sys/bus/pci/drivers_probe
echo ""  > /sys/bus/pci/devices/$BDF/driver_override
```

dranet's next inventory scan then republishes the device with its `ifName` attribute and pods schedule normally.

Observed on GKE image `dranet:v1.0.1-gke.0`, git `bb7e8b7af0c571b9e474981064e26ad9dc85c21f` from the now-archived `google/dranet`; the relevant code is unchanged on `kubernetes-sigs/dranet@main`.

Related design discussion: https://github.com/google/dranet/issues/97, https://github.com/google/dranet/issues/178

Things deliberately left out of this PR as follow-ups:

* **Self-heal in `GetNetInterfaceName`**: when a BDF is found with the wrong/no driver, write `driver_override` + `drivers_probe` to recover automatically before failing the prepare. Left out because it writes to host sysfs and is a policy decision (a workload may have unbound the driver intentionally for the lifetime of the node).
* **Kubernetes Event on lost device**: dranet has no `EventRecorder` plumbed today; wiring one is a separate change.
* **Re-key the local store by BDF instead of `ifName`**: after a `gve` rebind the kernel can hand out a different interface name (`eth1` ↔ `eth2` swap observed), so even a workload that correctly rebinds on exit can confuse the interface-name-keyed cleanup. Larger refactor; noted here as a known limitation.

**Downstream**: the affected build (`v1.0.1-gke.0`) is a GKE-shipped image built from the archived `google/dranet`. If GKE builds from a separate branch, this fix will need backporting there in addition to `main` here.

**Testing**: `GOOS=linux go build ./...`, `go vet ./...`, and `golangci-lint run ./...` (v2) pass. Added `TestIsAllocatableNetworkDevice` for the new helper. The remaining `inventory` tests require Linux to execute (netlink); deferring to CI.

```release-note
PCI network devices whose kernel driver has been unbound or replaced with a userspace driver (vfio-pci, uio_pci_generic, igb_uio, pci-stub) are no longer published in the ResourceSlice. Previously such devices were published but every NodePrepareResources call for them failed, trapping pods in FailedPrepareDynamicResources.
```

### DIFF
--- a/pkg/inventory/db.go
+++ b/pkg/inventory/db.go
@@ -59,6 +59,12 @@ var (
 	// ignoredInterfaceNames is a set of network interface names that are typically
 	// created by CNI plugins or are otherwise not relevant for DRA resource exposure.
 	ignoredInterfaceNames = sets.New("cilium_net", "cilium_host", "docker0")
+
+	// nonNetdevDrivers is the set of well-known kernel drivers that bind
+	// to PCI network devices without creating a kernel netdev or RDMA link
+	// (userspace I/O and passthrough drivers). See
+	// isAllocatableNetworkDevice for why this list is used.
+	nonNetdevDrivers = sets.New("vfio-pci", "uio_pci_generic", "igb_uio", "pci-stub")
 )
 
 type DB struct {
@@ -280,6 +286,10 @@ func (db *DB) discoverPCIDevices() []resourceapi.Device {
 
 	for _, pciDev := range pci.Devices {
 		if !isNetworkDevice(pciDev) {
+			continue
+		}
+		if !isAllocatableNetworkDevice(pciDev) {
+			klog.Warningf("PCI network device %s is bound to driver %q which does not provide a netdev; not publishing it", pciDev.Address, pciDev.Driver)
 			continue
 		}
 		device := resourceapi.Device{
@@ -605,4 +615,32 @@ func (db *DB) GetRDMADeviceName(deviceName string) (string, error) {
 // https://pcisig.com/sites/default/files/files/PCI_Code-ID_r_1_11__v24_Jan_2019.pdf
 func isNetworkDevice(dev *ghw.PCIDevice) bool {
 	return dev.Class.ID == "02"
+}
+
+// isAllocatableNetworkDevice reports whether dranet can ever prepare this
+// PCI network device for a pod. A device that is allocatable is either
+// providing a kernel netdev or RDMA link in some namespace; a device that is
+// not has nothing dranet can move and would trap any pod allocated to it in
+// FailedPrepareDynamicResources.
+//
+// The natural test would be "does /sys/bus/pci/devices/$BDF/net have any
+// entries", but those entries are netns-tagged and only visible from the
+// namespace the netdev currently lives in (see the kernel's
+// Documentation/networking/sysfs-tagging.rst). A device whose netdev has
+// already been moved into a pod's namespace therefore looks identical from
+// the host to a device whose driver creates no netdev at all, and we must
+// keep publishing the former so Cluster Autoscaler can size new nodes from
+// the running ResourceSlice.
+//
+// The kernel driver bound at /sys/bus/pci/devices/$BDF/driver is the one
+// signal that distinguishes the two: it is not namespaced and stays bound
+// while the netdev is in another namespace. We treat the device as not
+// allocatable if no driver is bound, or if the bound driver is one of the
+// well-known userspace/passthrough drivers that never create a netdev or
+// RDMA link.
+func isAllocatableNetworkDevice(dev *ghw.PCIDevice) bool {
+	if dev.Driver == "" {
+		return false
+	}
+	return !nonNetdevDrivers.Has(dev.Driver)
 }
--- a/pkg/inventory/db_test.go
+++ b/pkg/inventory/db_test.go
@@ -1,0 +1,48 @@
+/*
+Copyright The Kubernetes Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package inventory
+
+import (
+	"testing"
+
+	"github.com/jaypipes/ghw"
+)
+
+func TestIsAllocatableNetworkDevice(t *testing.T) {
+	cases := []struct {
+		name   string
+		driver string
+		want   bool
+	}{
+		{name: "unbound", driver: "", want: false},
+		{name: "vfio passthrough", driver: "vfio-pci", want: false},
+		{name: "uio generic", driver: "uio_pci_generic", want: false},
+		{name: "dpdk uio", driver: "igb_uio", want: false},
+		{name: "pci stub", driver: "pci-stub", want: false},
+		{name: "gve", driver: "gve", want: true},
+		{name: "mlx5", driver: "mlx5_core", want: true},
+		{name: "virtio", driver: "virtio_net", want: true},
+	}
+	for _, tc := range cases {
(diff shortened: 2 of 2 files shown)


gauravkghildiyal:
Thanks for contributing @wevans-ant 

My initial thoughts are that this generally looks reasonable and falls under the category of "not publishing devices which we know we can't allocate" (similar to what we discussed at https://github.com/kubernetes-sigs/dranet/pull/176#issuecomment-4299953695). I'll wait for others to chime in as well.

/ok-to-test
/assign

wevans-ant:
Appreciate the quick feedback @gauravkghildiyal! Would be interested to hear if you have thoughts on how a recovery path could or should work -- DRANET rebinding the original kernel driver to the device on pod shutdown seems reasonable to me but not confident enough to prescribe that behavior.

gauravkghildiyal:
> Appreciate the quick feedback @gauravkghildiyal! Would be interested to hear if you have thoughts on how a recovery path could or should work -- DRANET rebinding the original kernel driver to the device on pod shutdown seems reasonable to me but not confident enough to prescribe that behavior.

I share your hesitation regarding automatic rebinding, as DRANET currently lacks the necessary context to handle it safely. To put it in other words, DRANET is NOT original 'creator' of these devices but rather an 'allocator' of existing devices. So we don't really know the _correct_ driver to rebind a device to (if at all).

Overall, my current stance is that it makes perfect sense to not advertise devices we know we'll fail to "allocate" (which is what this PR proposes), but I would lean towards not trying to rebind right now and see if other related use cases show up and then figure out how to handle this well.

NOTE: The situation would be different if DRANET were responsible for the initial rebind, for example, if we supported DPDK or vfio-pci. In such a case, DRANET would have the 'before' state and could safely 'undo' its own changes upon returning the device to the host. However, as is the case here, since the driver was unbound externally (not by DRANET), DRANET cannot safely assume how to recover it.

aojea:
/assign @aojea 

@wevans-ant can you please sign the CLA ?

Some thoughts:

>> skip devices whose driver is either unbound ("") or one of the well-known non-netdev drivers (vfio-pci, uio_pci_generic, igb_uio, pci-stub). Those devices have no kernel netdev or RDMA link anywhere and can never be prepared, so they should not be advertised.

As correctled linked the problem here is well explained: we are advertising devices that are not allocatable.

However, for the solution, I want to explore a different approach that does not rely on a hardcoded list of drivers, can we do a helper function that returns a boolean indicating "isNetworkDeviceReady(pciDev)" or "isNetworkDeviceAllocatable(pciDev)" or similar ? and then we define there what are the conditions we expect the device to have. I like more if we can implement this logic "Those devices have no kernel netdev or RDMA link anywhere and can never be prepared, so they should not be advertised." than using the list of kernel drivers  


> Kubernetes Event on lost device: dranet has no EventRecorder plumbed today; wiring one is a separate change.

We need to add that, opened a new issue so anybody is free to take it https://github.com/kubernetes-sigs/dranet/issues/194

aojea:
I have to travel today and I do not have time to repro until next week but also I'm interested in knowing more about this

> discoverPCIDevices() still sees the PCI device (it never left the bus) and publishes it in the ResourceSlice with PCI-only attributes.
> The scheduler keeps allocating the device to new pods.


@gauravkghildiyal I think this PR is pretty good but I also want us to think more about the long term, so I want to fully understand the intersection with the scheduling and how to avoid this duplicate allocation on the scheduler itself ... I will come back to this next week as soon as I come back so please bear with me a couple of days more

wevans-ant:
> @wevans-ant can you please sign the CLA ?

Will do, waiting on legal approval on my side.

wevans-ant:
Thanks both. CLA is sorted; rebased onto `main` and pushed.

> can we do a helper function that returns a boolean indicating "isNetworkDeviceReady(pciDev)" or "isNetworkDeviceAllocatable(pciDev)" or similar?

Done — pulled it out into `isAllocatableNetworkDevice(dev *ghw.PCIDevice) bool` next to the existing `isNetworkDevice`, with a table test.

> I like more if we can implement this logic "Those devices have no kernel netdev or RDMA link anywhere and can never be prepared" than using the list of kernel drivers

I started down that road and hit a wall I've documented in the function comment, but here's the short version:

The obvious positive check is "is `/sys/bus/pci/devices/$BDF/net/` non-empty (or is there an RDMA link)". The problem is that the netdev kobjects under that directory are netns-tagged (kernel `Documentation/networking/sysfs-tagging.rst`), so from the host that directory is empty *both* when the driver creates no netdev *and* when the netdev exists but has been moved into a pod's namespace. With a pure sysfs check, an in-use device and a vfio-bound device look identical, and we'd stop publishing in-use devices — which IIUC is exactly what google/dranet#178 / google/dranet#194 moved to PCI-first discovery to avoid.

The driver symlink is the one piece of state that distinguishes the two: it isn't namespaced and stays pointing at `gve`/`mlx5_core`/etc. while the netdev is off in a pod. So the helper keys on it: no driver bound → never allocatable; driver is one of the small, stable set that never creates a netdev (`vfio-pci`, `uio_pci_generic`, `igb_uio`, `pci-stub`) → never allocatable; anything else → allocatable (covers both "idle in host ns" and "in use in a pod ns").

Also dropped the `nri_hooks.go` log-level change from this PR — #180 already moved those to `klog.Errorf`, so this is now a single-file change in `pkg/inventory/db.go`.

aojea:
/approve

Thanks for the explanation, this LGTM @gauravkghildiyal  for final LGTM

gauravkghildiyal:
/lgtm

Thank you!

PR #198, opened 2026-05-12
Resourceslice attr length overflow

dkennetzoracle:
#### What type of PR is this?

/kind bug

When adding link attributes, it joins every global-scope ipv6 or ipv4 address on the interface with commas and stores the result as a single DRA string attribute. String attrs are capped at 64 bytes:

(`spec.devices[N].attributes[dra.net/ipv4].string: Too long: may not be more than 64 bytes)`

However, kube-ipvs0 (the kube-proxy IPVS dummy) carries every ClusterIP and the API server rejected the resourceslice.

It ends up looking something like:

```
10.96.0.1/32,10.96.0.10/32,10.96.1.42/32,10.96.2.17/32,10.96.5.88/32,...
```

And the cap is hit after roughly 5 entries. When a cluster has dozens of services (prometheus, grafana, slinky, mariadb, contour, cert-manager... etc) the real string was hundreds of bytes long.

Additionally, I added a constant and a check for maxBytes allowed so that people don't arbitrarily hit this in the future.

N/A

The 64-byte cap is enforced by the upstream DRA API server (spec.devices[N].attributes[<key>].string: Too long: may not be more than 64 bytes); this PR does not change the limit, only how DRANET reacts when an attribute it would have published would exceed it.

```release-note
 Fixed DRANET failing to publish any ResourceSlice on nodes where an interface had enough global-scope IP addresses for the joined `dra.net/ipv4` or `dra.net/ipv6` attribute to exceed DRA's 64-byte string attribute limit (commonly triggered by `kube-ipvs0` on clusters running kube-proxy in IPVS mode). The oversized IP attribute is now omitted on the offending device; all other attributes and devices in the slice are unaffected.
```


gauravkghildiyal:
Thanks for the fix @dkennetzoracle 

So it seems like we'll have a proper solution once List type attributes are available (https://github.com/kubernetes/enhancements/issues/5491). In the mean time, I'm wondering whether we should:

- Exclude the attribute completely (the current proposal), OR
- Truncate the list + add a some suffix, like "..." (so the combined chars are within 64 chars)

I'm somewhat leaning towards the second option, but still thinking if there are any drawbacks.

/ok-to-test

dkennetzoracle:
Hi @gauravkghildiyal - thanks for the review! I wasn't aware that this enhancement will be coming in, but this should provide a nice stop-gap until it does. I've hit this twice on new installs now.

> Exclude the attribute completely (the current proposal), OR
Truncate the list + add a some suffix, like "..." (so the combined chars are within 64 chars)

In the previous install, I went with something like the first proposal, but instead just picked the first item in the list, so both functionally work. I'm okay with either if you'd prefer the second.

dkennetzoracle:
@gauravkghildiyal before any code changes on my end, the only drawback is that something like this:

> Truncate the list + add a some suffix, like "..." (so the combined chars are within 64 chars)

Not hard just a decent amount of additional code to support a "safe" maxBytes truncation which is why I just skipped it this time.  

```go
func TruncateCommaSeparatedList(value string, maxBytes int) (string, bool) {
	const suffix = "..."

	if maxBytes <= 0 {
		return "", len(value) > 0
	}

	if len(value) <= maxBytes {
		return value, false
	}

	// If even the suffix does not fit, return as much of the suffix as possible.
	// This is byte-safe because suffix is ASCII.
	if len(suffix) > maxBytes {
		return suffix[:maxBytes], true
	}

	parts := strings.Split(value, ",")

	var kept []string
	keptBytes := 0

	for _, part := range parts {
		partBytes := len(part)

		// Bytes needed to add this part to the kept list.
		addBytes := partBytes
		if len(kept) > 0 {
			addBytes++ // comma before this part
		}

		candidateKeptBytes := keptBytes + addBytes

		// Final string will be:
		//   kept1,kept2,...
		//
		// So reserve one comma before the suffix if we keep any item.
		totalBytes := candidateKeptBytes + len(suffix)
		if candidateKeptBytes > 0 {
			totalBytes++ // comma before suffix
		}

		if totalBytes > maxBytes {
			break
		}

		kept = append(kept, part)
		keptBytes = candidateKeptBytes
	}

	if len(kept) == 0 {
		return suffix, true
	}

	return strings.Join(kept, ",") + "," + suffix, true
}
```

pkg/inventory/db.go
```diff
@@ -53,6 +53,11 @@ const (
 	// defaultPollBurst is the default number of polls that can be run in a
 	// burst.
 	defaultPollBurst = 5
+
+	// maxDeviceAttributeStringLen is the per-attribute string-value byte limit
+	// enforced by DRA (resource.k8s.io). Values larger than this cause the
+	// API server to reject the entire ResourceSlice.
```
tamilmani1989: do we have any link to doc/page claiming 64 bytes limit. if there, lets update link in comment

pkg/inventory/net_test.go
```diff
@@ -500,6 +505,251 @@ func addChildDummy(t *testing.T, name string, masterIndex int) netlink.Link {
...
+		{
+			name:      "many IPv6 overflow - attribute omitted",
+			ipv6:      manyV6,
+			wantV6Set: false,
+		},
+		{
+			// kube-ipvs0-style mix: lots of v4 ClusterIPs overflow the limit
+			// but a single v6 host address still fits — only AttrIPv4 should
+			// be dropped, AttrIPv6 must still be published.
+			name:        "v4 overflow does not drop v6",
```
tamilmani1989: case for v6 overflow does not drop v4?

pkg/inventory/db.go
aojea: good point, we can use the exported variable instead of having to define our own https://github.com/kubernetes/kubernetes/blob/f01363b4912ad545ff1a50b690e21bc148271a8c/staging/src/k8s.io/api/resource/v1/types.go#L777

```go
// DeviceAttributeMaxValueLength is the maximum length of a string or version attribute value.
const DeviceAttributeMaxValueLength = 64
```

aojea:
@dkennetzoracle let's truncate but just we evaluate how many IPs can we fit while building the attribute, instead of building the attribute and truncate later

```go
func buildIPList(ips []string, maxBytes int) string {
	if len(ips) == 0 {
		return ""
	}

	var builder strings.Builder
	currentLength := 0

	for i, ip := range ips {
		// Calculate how many bytes this IP (and its comma separator) would add
		addedLength := len(ip)
		if i > 0 {
			addedLength += 1 // For the comma
		}

		// If adding this IP exceeds our limit, we stop building right here
		if currentLength+addedLength > maxBytes {
			break
		}

		// Append the comma separator for subsequent items
		if i > 0 {
			builder.WriteString(",")
		}

		builder.WriteString(ip)
		currentLength += addedLength
	}

	return builder.String()
}
```

pkg/inventory/net_test.go
dkennetzoracle: Addressed

dkennetzoracle:
@aojea - addressed - truncating rather than skipping and importing the resources api constant directly so it should also work if that were to ever change. Good callout! cc: @gauravkghildiyal

pkg/inventory/db.go
```diff
@@ -405,6 +405,39 @@ func (db *DB) discoverNetworkInterfaces(pciDevices []resourceapi.Device) []resou
...
+	currentLength := 0
+	kept := 0
+
+	for i, ip := range ips {
+		addedLength := len(ip)
+		if i > 0 {
+			addedLength++ // comma separator
+		}
+
+		if currentLength+addedLength > maxBytes {
```
gauravkghildiyal: nit: Is currentLength variable needed given builder.Len() == currentLength?

pkg/inventory/net_test.go
```diff
@@ -500,6 +505,408 @@ func addChildDummy(t *testing.T, name string, masterIndex int) netlink.Link {
...
+// TestAddLinkAttributesIPLengthCap covers the per-attribute string-value
+// limit on AttrIPv4 / AttrIPv6 (see resourceapi.DeviceAttributeMaxValueLength).
+// The kube-proxy IPVS dummy interface (kube-ipvs0) accumulates every cluster
+// ServiceIP, and if dranet joins them all into a single comma-separated
+// attribute the resulting ResourceSlice is rejected by the API server with
+// "Too long: may not be more than 64 bytes". addLinkAttributes must truncate
+// the joined value so it fits within the cap while still publishing as many
+// addresses as possible, and the other family's attribute must not be
+// affected when only one family overflows.
+func TestAddLinkAttributesIPLengthCap(t *testing.T) {
```
gauravkghildiyal: Can we please move the tests to db_test.go to match the source files of the function under test as the idiomatic way?

gauravkghildiyal (commented):
Just a small request and we're ready.

pkg/inventory/net_test.go
dkennetzoracle: @gauravkghildiyal - done. Sorry these felt like net stuff, but realized the LinkAttrs are in db. Resolved!

gauravkghildiyal:
Thanks!

/lgtm

PR #201, opened 2026-05-17
chore: drop unnecessary helm installation during workflow

fmuyassarov:
#### What type of PR is this?
/kind cleanup

According to https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md, Helm is already pre baked into the image. Thus, drop unnecessary step of Helm binary installation.

"N/A".

```release-note
"NONE"
```

### DIFF
--- a/.github/workflows/helm-lint.yml
+++ b/.github/workflows/helm-lint.yml
@@ -4,15 +4,13 @@ on:
     branches: [ main ]
     paths:
       - 'deployments/helm/**'
+      - '.github/workflows/helm-lint.yml'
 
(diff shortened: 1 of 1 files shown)


fmuyassarov:
Example run of its own: https://github.com/kubernetes-sigs/dranet/actions/runs/26002654742/job/76428610169?pr=201

aojea:
/lgtm
/approve

Good, the less dependencies we use the less exposed we are to this lately wave of github actions attacks

PR #206, opened 2026-05-19
Add IPvlan L2 driver for Alibaba Cloud HPN nodes

Bowser1704:
#### What type of PR is this?

/kind feature

Introduce a configurable IPvlan driver that creates IPvlan slave
interfaces in the pod namespace while keeping the parent netdev on
the host. The implementation supports L2/L3/L3S modes, bridge/private/
vepa flags, and three addressing strategies (none, static,
parentIPv6PrefixPodIPv4).

Key changes:
- Expand IPVlanConfig API with mode, flag, addressing, and route/
  neighbor copy options
- Add IPvlan validation in the config validation pipeline
- Implement idempotent attach: detect existing target interface in
  pod netns on retry, tolerate EEXIST on address assignment
- Use deterministic temp names (parentName_iv) serialized by mutex
- Support user-configured routes with full Scope/Table/Source fields,
  sorted link-scope-first for gateway reachability
- Fail-fast on neighbor resolution errors instead of silently
  degrading connectivity
- Make route/neighbor copy from parent opt-in via config
- Decouple RDMA discovery from IPvlan config
- Add Alibaba Cloud provider that auto-detects HPN instances and
  generates the IPvlan preset for bond devices with global IPv6

N/A

```release-note
Add IPvlan L2 networking mode for Alibaba Cloud HPN nodes, allowing non-hostNetwork pods to use RoCE RDMA with derived IPv6 addresses. Activated automatically on HPN instances when using `--cloud-provider-hint=ALIBABA`.
```


gauravkghildiyal:
Happy to have you contributing @Bowser1704 

This is very interesting as we're also working on a somewhat generic approach towards having IPVLAN as part of https://github.com/kubernetes-sigs/dranet/issues/63. @ngcxy is working on a POC at the moment and we can align to make sure it covers your use case too (from the looks of it, I think it would).

Bowser1704:
Thanks @gauravkghildiyal! 

I've actually implemented a more generic IPvlan driver in this PR — it's not tightly coupled to Alibaba Cloud. The cloud provider hint just signals when to enable it. 

Would love to align with @ngcxy's POC from #63 — if we can converge on the IPvlan approach, that would let Alibaba Cloud users onboard to dranet right away. Happy to iterate on the design together.

gauravkghildiyal:
Thanks for the context.

We'd very much like to get things work with Alibaba Cloud. But the approach that I think we should take with IPVLAN ( or MACVLAN and others) is also having the flexibility to share the parent interface across multiple pods and doing this in a more DRA native way (using things like DRA Consumable Capacity and others)

This is something we are actively working towards and I'd highly encourage and request we wait for the work being done for https://github.com/kubernetes-sigs/dranet/issues/63. This is slightly opinionated, but it'd be awesome to get the IPVLAN support right from the get-go -- a modeling that allows for future extension like sharing between multiple pods. We'd appreciate your patience with this and would be happy to collaborate together so that all requirements are met. This PR is already a great demonstration of what exactly are the things we should have a solution for (in addition to  things like device sharing).

PR #214, opened 2026-06-02
Remove multiple pods sharing one claim in prepareResourceClaim

ngcxy:
#### What type of PR is this?

/kind cleanup

Previously `prepareResourceClaim` implements the iteration over pods reserved for a single claim. This is a preparation work for the scenario where a single claim could be shared among multiple pods, when they're claiming a shared device. Currently this logic is never used since shared device hasn't been introduced yet.

Our finalized design came out that even in shared device scenarios, each pod will get its own distinct ResourceClaim, and the driver handles subinterface creation per claim. Therefore this iteration over multiple pods becomes redundant.

This change removes this logic to simply the preparation logic given this design. 

```release-note
- Add 1:1 pod-to-claim enforcement
  - Previously the driver allows preparing resource for a claim reserved by multiple pods (in rare and anti-pattern case when they are created simultaneously with the same claim).
  - Now it fails the preparation if the reserved reference is larger than one.
- Non-pod fail-fast
  - Previously a claim containing single non-pod reference would succeed silently.
  - Now it fails resourceClaim preparation with the validation error.
```

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -144,19 +144,25 @@ func (np *NetworkDriver) prepareResourceClaim(ctx context.Context, claim *resour
 	defer func() {
 		klog.V(2).Infof("PrepareResourceClaim Claim %s/%s  took %v", claim.Namespace, claim.Name, time.Since(start))
 	}()
-	// TODO: shared devices may allocate the same device to multiple pods, i.e. macvlan, ipvlan, ...
-	podUIDs := []types.UID{}
-	for _, reserved := range claim.Status.ReservedFor {
-		if reserved.Resource != "pods" || reserved.APIGroup != "" {
-			klog.Infof("Driver only supports Pods, unsupported reference %#v", reserved)
-			continue
-		}
-		podUIDs = append(podUIDs, reserved.UID)
-	}
-	if len(podUIDs) == 0 {
+	if len(claim.Status.ReservedFor) == 0 {
 		klog.Infof("no pods allocated to claim %s/%s", claim.Namespace, claim.Name)
 		return kubeletplugin.PrepareResult{}
 	}
+	if len(claim.Status.ReservedFor) > 1 {
+		return kubeletplugin.PrepareResult{
+			Err: fmt.Errorf("driver only supports one pod per claim, got %d", len(claim.Status.ReservedFor)),
+		}
+	}
+
+	// One ResourceClaim is consumed by a single pod, regardless of whether the device is allocated in
+	// exclusive or shared (e.g., ipvlan, macvlan) way, so we process the first and only consumer in ReservedFor.
+	reserved := claim.Status.ReservedFor[0]
+	if reserved.Resource != "pods" || reserved.APIGroup != "" {
+		return kubeletplugin.PrepareResult{
+			Err: fmt.Errorf("driver only supports Pods, unsupported reference %#v", reserved),
+		}
+	}
+	podUID := reserved.UID
 
 	nlHandle, err := nlwrap.NewHandle()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/driver/dra_hooks.go
```diff
@@ -247,12 +248,10 @@ func (np *NetworkDriver) prepareResourceClaim(ctx context.Context, claim *resour
 				continue
 			}
 			deviceCfg.RDMADevice = buildRDMAConfig(rdmaDevName, charDevices)
-			for _, uid := range podUIDs {
```
gauravkghildiyal, line 250: I initially came to the conclusion changing this would break IB-only devices. IB-only devices do not have a NetDevice and only have an RDMADevice. As long as the RDMADevice is operating is shared mode, you can share it across containers and you are not restricted by the limitation of not being able to assigned the associated NetDevice across containers.

But on second inspection, I realized that PrepareResourceClaim is only called the very first time we're preparing the ResourceClaim. So if we have `pod-1` as the first pod using `claim-1`, we expect to see PrepareResourceClaim called, but for a second `pod-2` created in the future which also tries to use `claim-`, PrepareResourceClaim will not get called, which means that the pod will fail during the NRI setup phase as its config would not be found in the podConfigStore

The only rare (unverified) case when two pods `pod-1` and `pod-2` would be able to share the same `claim-1` is when we try to create both of these at the exact same time, and then they are batched together by the kubelet into a single PrepareResourceClaim request. But this is an inconsistent behaviour which is more likely to fail rather than succeed.

Based on this, I'm thinking that perhaps we're safe with there being no use case where multiple pods currently consistently use the same claim and hence this should be fine to refactor.

FYI @anson627 to make sure I'm not missing anything for the IB-only devices.

pkg/driver/dra_hooks.go
```diff
@@ -144,20 +144,21 @@ func (np *NetworkDriver) prepareResourceClaim(ctx context.Context, claim *resour
...
-	}
-	if len(podUIDs) == 0 {
+	if len(claim.Status.ReservedFor) == 0 {
 		klog.Infof("no pods allocated to claim %s/%s", claim.Namespace, claim.Name)
 		return kubeletplugin.PrepareResult{}
 	}
 
+	// One ResourceClaim is consumed by a single pod, regardless of whether the device is allocated in
+	// exclusive or shared (e.g., ipvlan, macvlan) way, so we process the first and only consumer in ReservedFor.
+	reserved := claim.Status.ReservedFor[0]
```
gauravkghildiyal, line 159: (Please read https://github.com/kubernetes-sigs/dranet/pull/214/changes#r3345439122 before this comment)

Can we confirm our hypothesis from the above comment to understand the situation where we will expect to see multiple ReservedFor pods in this list? Once we've confirmed that, let's actually document why we think it is sufficient for us to use just the first `ReservedFor[0]` pod. 

In the case that we encounter multiple pods in the `ReservedFor` list, perhaps its better to fail early as something we don't support.

gauravkghildiyal (commented):
Thanks @ngcxy 

/ok-to-test

pkg/driver/dra_hooks.go
anson627, line 159: The old code logged and continued when encountering a non-Pod reference in ReservedFor. The new code returns an error immediately. This is a behavior change worthwhile callout in release notes: a claim with a single non-Pod reserved reference previously succeeded silently (with an info log), and now it fails the entire claim preparation

ngcxy: Thank you Gaurav for the great insights!

Just a quick validation note for the two edge scenarios that you mentioned:

- When we create pod-1 with claim-1, and then create pod-2 with claim-1, it's validated that `PrepareResourceClaim` is skipped and no device configuration is copied for pod-2 into `podConfigStore`. Then NRI skips pod-2 as no-op without configuring the interface to pod-2.
- When we create pod-1 and pod-2 simultaneously using the same claim-1, in old code the driver copied device configuration for both pods, and whether NRI succeeds depends on if the device is shareable. But you're right that this pattern isn't standard. I have added the validation to fail early when this case occurs that `ReservedFor` has more than one references.

ngcxy: Thank you Anson for pointing this out! I have updated this note into the description.

gauravkghildiyal:
Thanks!

/lgtm
/approve

PR #220, opened 2026-06-04
chore: Update KIND and k8s versions across tests to facilitate package upgrades

gauravkghildiyal:
#### What type of PR is this?

/kind dependency
/kind cleanup

This update is motivated by test failures highlighted in https://github.com/kubernetes-sigs/dranet/pull/218

```release-note
NONE
```

### DIFF
--- a/.github/workflows/bats.yml
+++ b/.github/workflows/bats.yml
@@ -12,8 +12,8 @@ on:
 
 env:
   GO_VERSION: "1.26"
-  K8S_VERSION: "v1.34.0"
-  KIND_VERSION: "v0.30.0"
+  K8S_VERSION: "v1.36.1"
+  KIND_VERSION: "v0.32.0"
   IMAGE_NAME: registry.k8s.io/networking/dranet
   KIND_CLUSTER_NAME: kind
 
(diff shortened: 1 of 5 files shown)


.github/workflows/bats.yml
```diff
@@ -34,6 +34,20 @@ jobs:
...
+          curl -L "https://dl.k8s.io/${{ env.K8S_VERSION }}/bin/linux/amd64/kubectl" -o "${TMP_DIR}/kubectl"
+          # kind
+          curl -Lo "${TMP_DIR}/kind" "https://kind.sigs.k8s.io/dl/${{ env.KIND_VERSION }}/kind-linux-amd64"
+          # Install
+          sudo cp "${TMP_DIR}/kubectl" /usr/local/bin/kubectl
+          sudo cp "${TMP_DIR}/kind" /usr/local/bin/kind
+          sudo chmod +x /usr/local/bin/*
+          # Clean
+          sudo rm -rf "${TMP_DIR}"
+
```
fmuyassarov, line 50: according to https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md the runner image comes with the these binaries out of the box. If there is no hard requirement for specific versions, perhaps we could use what is already available?

fmuyassarov: so we have 
Kind 0.31.0 and Kubectl 1.36.1

aojea: pinned dependencies guarantees resilience to external floating dependencies, is not the first time some actions changes introduce flakiness and you start to spend days until realize that change was not visibile to you

aojea:
/lgtm
/approve

explicit > implicit

PR #221, opened 2026-06-05
feat: plumb EventRecorder and emit failure events

fmuyassarov:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Add eventRecorder to the driver and as a starting point we record and publish only the following failure events: ClaimPrepareFailed, NetworkDeviceAttachFailed, RDMADeviceAttachFailed.

One thing I wasn't fully convinced about was the benefit of propagating the event into the driver logs via ...
```go
eventBroadcaster.StartLogging(klog.Infof)
```

This basically replicates the event into the kubectl logs of the driver. In a way I don't see an issue propagating it to the logs and it may help to discovery the failures a bit faster but at the same time I'm thinking that publishing the failure in the events is enough. I don't have a strong opinion on this and would be happy to hear you think on this.
in practice in looks like this

```
I0605 09:29:20.122080       1 db.go:592] Device "dummy0" not found in local store, rescanning.
I0605 09:29:20.271155       1 dra_hooks.go:432] claim c8810e22-6955-4690-911b-3251b027bcd1 contain errors: failed to get network interface name for device dummy0: device dummy0 not found in store
I0605 09:29:20.271283       1 nonblockinggrpcserver.go:161] "handling request succeeded" logger="dra" requestID=4 method="/k8s.io.kubelet.pkg.apis.dra.v1.DRAPlugin/NodePrepareResources" response="claims:{key:\"c8810e22-6955-4690-911b-3251b027bcd1\" value:{error:\"claim c8810e22-6955-4690-911b-3251b027bcd1 contain errors: failed to get network interface name for device dummy0: device dummy0 not found in store\"}}"
I0605 09:29:20.271407       1 event.go:377] Event(v1.ObjectReference{Kind:"ResourceClaim", Namespace:"default", Name:"dummy-interface-static-ip", UID:"c8810e22-6955-4690-911b-3251b027bcd1", APIVersion:"resource.k8s.io/v1", ResourceVersion:"897", FieldPath:""}): type: 'Warning' reason: 'ClaimPrepareFailed' failed to get network interface name for device dummy0: device dummy0 not found in store
I0605 09:30:38.120584       1 db.go:592] Device "dummy0" not found in local store, rescanning.
I0605 09:30:38.271555       1 dra_hooks.go:432] claim c8810e22-6955-4690-911b-3251b027bcd1 contain errors: failed to get network interface name for device dummy0: device dummy0 not found in store
[...]

PR #222, opened 2026-06-05
chore: add startup probe to avoid false unhealthy events

fmuyassarov:
#### What type of PR is this?
/kind cleanup

Give the driver enough time to complete plugin registration before throwing unhealthy events, which may give wrong signal. While introducing events in https://github.com/kubernetes-sigs/dranet/pull/221, I noticed probe failure events on the driver Pod and I was a bit confused for a while until I realized that readinessProbe was faster than DRANET waiting timeout for the kubelet to complete plugin registration (30 sec). Thus, adding startup timeout should avoid getting false positives even if the registration timeout takes longer but no more than 30 sec which is set [here](https://github.com/kubernetes-sigs/dranet/blob/4fcedee6e72f1d285fc4d6602e1b9e8364b6071f/pkg/driver/driver.go#L170).

I've given in total 60s for the startup probe (2 times 30s [timeout](https://github.com/kubernetes-sigs/dranet/blob/4fcedee6e72f1d285fc4d6602e1b9e8364b6071f/pkg/driver/driver.go#L170) set for the kubelet plugin registration) and those extra seconds basically to cover image pulling and other stuff that are not directly about the driver itself.

before
```
Events:
  Type     Reason     Age   From               Message
  ----     ------     ----  ----               -------
  Normal   Scheduled  14s   default-scheduler  Successfully assigned kube-system/dranet-7sxql to dra-worker2
  Normal   Pulled     13s   kubelet            spec.containers{dranet}: Container image "registry.k8s.io/networking/dranet:v1.3.0" already present on machine and can be accessed by the pod
  Normal   Created    13s   kubelet            spec.containers{dranet}: Container created
  Normal   Started    13s   kubelet            spec.containers{dranet}: Container started
  Warning  Unhealthy  13s   kubelet            spec.containers{dranet}: Readiness probe failed: Get "http://172.18.0.3:9177/healthz": dial tcp 172.18.0.3:9177: connect: connection refused
  Warning  Unhealthy  12s   kubelet            spec.containers{dranet}: Readiness probe failed: HTTP probe failed with statuscode: 503
```

after
```
Events:
  Type    Reason     Age   From               Message
  ----    ------     ----  ----               -------
  Normal  Scheduled  29m   default-scheduler  Successfully assigned kube-system/dranet-m2gwf to dra-worker
  Normal  Pulling    29m   kubelet            spec.containers{dranet}: Pulling image "ttl.sh/dranet:3aa6a69"
  Normal  Pulled     29m   kubelet            spec.containers{dranet}: Successfully pulled image "ttl.sh/dranet:3aa6a69" in 4.939s (4.939s including waiting). Image size: 46028325 bytes.
  Normal  Created    29m   kubelet            spec.containers{dranet}: Container created
  Normal  Started    29m   kubelet            spec.containers{dranet}: Container started
```

"N/A".

```release-note
 "NONE"
```

### DIFF
--- a/deployments/helm/dranet/templates/daemonset.yaml
+++ b/deployments/helm/dranet/templates/daemonset.yaml
@@ -86,6 +86,12 @@ spec:
             {{- end }}
           securityContext:
             privileged: true
+          startupProbe:
+            httpGet:
+              path: {{ .Values.metricsPath }}
+              port: {{ .Values.metricsPort | default 9177 }}
+            failureThreshold: 12
+            periodSeconds: 5
           readinessProbe:
             httpGet:
               path: {{ .Values.metricsPath }}
--- a/examples/dranetctl-install.yaml
+++ b/examples/dranetctl-install.yaml
@@ -130,6 +130,12 @@ spec:
         securityContext:
           capabilities:
             add: ["NET_ADMIN", "SYS_ADMIN"]
+        startupProbe:
+          httpGet:
+            path: /healthz
+            port: 9177
+          failureThreshold: 12
+          periodSeconds: 5
         readinessProbe:
           httpGet:
             path: /healthz
--- a/install-containerd-1.7yaml
+++ b/install-containerd-1.7yaml
@@ -136,6 +136,12 @@ spec:
             memory: "50Mi"
         securityContext:
           privileged: true
+        startupProbe:
+          httpGet:
+            path: /healthz
+            port: 9177
+          failureThreshold: 12
+          periodSeconds: 5
         readinessProbe:
           httpGet:
             path: /healthz
--- a/install.yaml
+++ b/install.yaml
@@ -119,6 +119,12 @@ spec:
             memory: "50Mi"
         securityContext:
           privileged: true
+        startupProbe:
+          httpGet:
+            path: /healthz
+            port: 9177
+          failureThreshold: 12
+          periodSeconds: 5
         readinessProbe:
           httpGet:
             path: /healthz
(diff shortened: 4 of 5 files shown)


aojea:
I want to be very conservative here since a wrong value can cause restarts, and there are several dependencies:
- kubelet
- container runtime
- after https://github.com/kubernetes-sigs/dranet/pull/223 , webhooks

60 seconds seems to short, @gauravkghildiyal you have more insights on the production metrics than I do, do you feel comfortable defining a value here?

fmuyassarov:
> I want to be very conservative here since a wrong value can cause restarts, and there are several dependencies:
> 
>     * kubelet
> 
>     * container runtime
> 
>     * after [Bring Your Own DRANET Provider (BYODP) #223](https://github.com/kubernetes-sigs/dranet/pull/223) , webhooks
> 
> 
> 60 seconds seems to short, @gauravkghildiyal you have more insights on the production metrics than I do, do you feel comfortable defining a value here?

Thanks @aojea . Happy to adjust. I will admit, I don't run the driver on production (yet) thus I can't have a good reference to the timeout. But, I think having the incorrect timeout which causes false warning isn't also something we expect to happen. We could also improve this within Start() but it feels an overkill to do it there. 
To be safe, we could give longer timeout for the startupProbe since it is maximum window. Another alternative would be to add initialDelaySeconds to give little extra time to warm up the driver.

gauravkghildiyal:
60 seconds sounds good to me for now and for some time in the future.

60 is double than our kubelet registration timeout. We handle the NRI registration async. So the only remaining thing to think about remains BYODP, which we don't have data on at the moment.

/lgtm
/approve

PR #223, opened 2026-06-08
Bring Your Own DRANET Provider (BYODP)

### LINKED ISSUE #103: Add support for Whereabouts IPAM
## Problem

DRANET currently does not provide a built-in mechanism for managing IP address ranges for network devices.  
As a result, IP address allocation must be implemented separately or reimplemented.

## Proposal

Add support for the [Whereabouts](https://github.com/k8snetworkplumbingwg/whereabouts) IPAM project in DRANET.

Whereabouts is primarily designed as a CNI-based IPAM plugin. However, its core IP range management logic can also be reused as a library. By importing it as a library, DRANET could leverage its existing functionality for IP allocation without depending on the CNI plugin interface.

Whereabouts is already widely used in Kubernetes environments and provides a well-tested implementation for managing IP address ranges. Reusing it would avoid reimplementing similar functionality in DRANET and provide a familiar IPAM mechanism for users.

## Example

I have prepared a sample implementation demonstrating this approach.  
Some demonstration manifests are also included to show how the feature could be used.
https://github.com/kanlkan/dranet/blob/feature/whereabouts-ipam/temp/README.md

Feedback on this approach would be appreciated.  
If the direction looks reasonable, I can open a draft PR with the proposed implementation.

aojea:
#### What type of PR is this?
/kind feature
/kind documentation
This PR introduces the **Bring Your Own DRANET Provider (BYODP)** feature, allowing users to provide custom implementations for both hardware discovery and user intent via external webhooks. 

Instead of hardcoding bare-metal or CNI-specific logic directly into DRANET, we can now delegate these responsibilities to an external HTTP REST server or Unix domain socket. This ensures DRANET's core runtime remains solid, statically verifiable, and predictable, while still exposing flexibility for third-party extensions.

This PR specifically introduces:

1. **Cloud Providers (Cluster Provider Intent)**: Webhooks can dictate physical infrastructural (e.g., base MTU, hardware MAC, PCI boundaries).
2. **Profile Providers (User Intent)**: Webhooks can dictate logical network assignments like IPAM (e.g., allocating an IP address from an external IPAM plugin like `whereabouts` or from a Cloud Provider).
3. **Webhook Capabilities Discovery**: DRANET queries `/health` on startup to negotiate `cloudProvider` and `profileProvider` capability support.
4. **Reference Implementation**: Adds a fully functional reference webhook server (`cmd/webhook-whereabouts`) that wraps the standard CNI `whereabouts` plugin to deliver dynamic IPAM profiles.

Fixes: #103

To make the review process this PR has 3 commits:

1. **Refactor**: Decouples the existing cloud hardware discovery logic out of `pkg/inventory` into a distinct `pkg/cloudprovider/discovery` package.
2. **Feature & Docs**: Introduces the generic webhook client, `ProfileProvider` interface integration in the inventory DB, and comprehensive architectural documentation (`site/content/docs/contributing/webhook-providers.md`).
3. **Example & Tests**: Adds the `webhook-whereabouts` reference module and the associated E2E bats tests verifying the dynamic whereabouts CNI integration.
4
```release-note
Add support for Custom Webhook Providers (BYODP) via the `--cloud-provider-hint=webhook`, `--profile-provider=webhook`, and `--webhook-url` flags. This enables users to delegate physical hardware discovery and logical network IPAM assignments to external HTTP servers or Unix domain sockets.
```


pkg/cloudprovider/discovery/discovery.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+type CloudProviderHint string
+
+const (
+	CloudProviderHintGCE     CloudProviderHint = "GCE"
+	CloudProviderHintAWS     CloudProviderHint = "AWS"
+	CloudProviderHintAzure   CloudProviderHint = "AZURE"
+	CloudProviderHintOKE     CloudProviderHint = "OKE"
+	CloudProviderHintWebhook CloudProviderHint = "webhook"
```
aojea, line 39: This leaked into the first commit when tryied to reorganize for simplifying the review

pkg/cloudprovider/discovery/discovery_test.go
```diff
@@ -0,0 +1,102 @@
...
+*/
+
+package discovery
+
+import (
+	"context"
+	"testing"
+)
+
+func TestGetInstanceProperties(t *testing.T) {
```
aojea: this test is wrong

pkg/inventory/db.go
```diff
@@ -555,6 +586,62 @@ func (db *DB) GetDevice(deviceName string) (resourceapi.Device, bool) {
...
 
+func (db *DB) getProfileProvider() cloudprovider.ProfileProvider {
+	return db.profProv
+}
+
+// GetProfileConfig resolves a dynamic profile by querying the underlying cloud provider.
+func (db *DB) GetProfileConfig(deviceName, profile string, claimUID types.UID) (*apis.NetworkConfig, error) {
+	p := db.getProfileProvider()
+	if p == nil {
+		return nil, fmt.Errorf("current cloud provider does not support dynamic profiles")
```
aojea, line 597: This means that the user hasn't configured a dynamic profile provider (e.g. they didn't start the dranet daemon with --profile-provider=webhook or the clooud provider does not implement it, however, if this function is being called, it strictly means that a Pod is attempting to start with a specific Profile explicitly requested in its DRANET Network Config

cmd/dranet/app.go
```diff
@@ -171,12 +178,67 @@ func main() {
 		}
 		opts = append(opts, driver.WithFilter(prg))
 	}
-	db := inventory.New(
+	var cloudInst cloudprovider.CloudInstance
+	var profProv cloudprovider.ProfileProvider
+
+	var hint discovery.CloudProviderHint
+	// Auto-discover cloud provider if not explicitly set
+	if cloudProviderHint == "" {
```
anson627: I understand the intent, the concern is that the code currently makes explicit cloud-provider failures indistinguishable from no underlay provider requested

I would suggest sth like this:
```
explicitCloudProvider := cloudProviderHint != ""
cloudInst, err = discovery.GetInstanceProperties(ctx, hint, webhookURL)
if err != nil {
    if explicitCloudProvider {
        klog.Fatalf("failed to initialize cloud provider %q: %v", hint, err)
    }
    klog.Infof("failed to initialize auto-discovered cloud provider %q: %v", hint, err)
    cloudInst = nil
}
```

pkg/cloudprovider/webhook/webhook.go
```diff
@@ -0,0 +1,248 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqURL, nil)
+	if err == nil {
+		if resp, err := p.client.Do(req); err == nil {
+			defer resp.Body.Close()
+			if resp.StatusCode == http.StatusOK {
+				var caps Capabilities
+				if err := json.NewDecoder(resp.Body).Decode(&caps); err == nil {
+					p.caps = caps
+				} else {
+					p.caps = Capabilities{CloudProvider: true, ProfileProvider: true}
```
anson627: This returns a provider even when /health failed, because all errors above are swallowed

For explicit webhook use, startup should fail if the capability contract cannot be validated, otherwise the dranet pod can become ready with a misconfigured webhook and fail later during claim preparation

pkg/driver/dra_hooks.go
```diff
@@ -479,6 +503,22 @@ func (np *NetworkDriver) unprepareResourceClaims(ctx context.Context, claims []k
...
+	for _, podUID := range np.podConfigStore.ListPods() {
+		podCfg, ok := np.podConfigStore.GetPodConfig(podUID)
+		if !ok {
+			continue
+		}
+		for deviceName, devCfg := range podCfg.DeviceConfigs {
+			if devCfg.Claim.Namespace == claim.Namespace && devCfg.Claim.Name == claim.Name {
+				if devCfg.NetworkInterfaceConfigInPod.Profile != "" {
+					if err := np.netdb.ReleaseProfileConfig(deviceName, devCfg.NetworkInterfaceConfigInPod.Profile, claim.UID); err != nil {
+						klog.Errorf("failed to release profile config for claim %v: %v", claim.NamespacedName, err)
```
anson627, line 504: it is better to return the ReleaseProfileConfig error instead of only logging it, otherwise unprepare reports success, deletes the claim state, and kubelet will not retry leaked profile/IPAM cleanup

aojea: Everything on pod teardown use to get swallowed because a retry does not imply a retry of the operation and the risk of causing deadlock situations on exit.

This is also how [CNI works and is how is also specified](https://www.cni.dev/docs/spec/#del-remove-container-from-network-or-un-apply-modifications)

> Plugins should generally complete a DEL action without error even if some resources are missing

and how NRI works too, we are trying to document the NRI contract too https://github.com/containerd/nri/pull/286

I understand your concern but is like a `defer x.Close()` , is on the provider to ensure there is no leaks

cmd/dranet/app.go
aojea: This keeps the same behavior as before, we do not want to make a breaking change because this will cause that dranets that worked before to start to fail ... also we explicitily indicated in the flag is just a `hint` because of this

pkg/cloudprovider/webhook/webhook.go
aojea: good catch, fixed, added new commits for simplifying reviews

pkg/driver/dra_hooks.go
anson627: make sense

cmd/dranet/app.go
anson627: sounds good

aojea:
/hold

latest e2e test is not working correctly, will unhold once I'm understanding better the problem

tests/python_webhook.bats
```diff
@@ -54,9 +54,24 @@ teardown_file() {
   kubectl --context kind-dranet-test-cluster apply -f "$BATS_TEST_DIRNAME"/../tests/manifests/deviceclass.yaml
 
   # Wait for the interface to be discovered and validate the custom attribute from webhook
-  sleep 5
-  run kubectl --context kind-dranet-test-cluster get resourceslices --field-selector spec.nodeName="$NODE_NAME" -o jsonpath='{.items[0].spec.devices[?(@.name=="dummy1")].attributes.dra\.net\/webhook-attr.string}'
```
aojea: the hyphen fails the apiserver validation `webhook-attr` and the ResourceSlice was never updated, I can see how this can happen in prod

anson627:
these may fix e2e issues:
1. Update tests/manifests/python_webhook_pod.yaml to use webhook_attr and avoid has on map indexing
2. Run go mod tidy and include go.mod/go.sum changes for cmd/webhook-whereabouts

aojea:
/hold cancel

tests are passing, covering custom webhooks and wherabouts ipam

site/content/docs/contributing/webhook-providers.md
```diff
@@ -0,0 +1,121 @@
...
+  * **Network Failures**: If DRANET cannot reach the webhook due to a network timeout or connection refused, it also fails the `NodePrepareResources` call, and kubelet will continually retry.
+
+  **Trade-offs & Downsides (Node-level vs. API-level Validation)**:
+  Because this validation happens at the *node level* during DRA `NodePrepareResources` (rather than at the API server via standard Admission Webhooks), there are important trade-offs to consider:
+  * **Late Feedback**: If a configuration is invalid, the API server will still accept the Pod. The failure happens asynchronously when the pod is scheduled and the kubelet attempts to prepare resources. Users won't see an immediate error on `kubectl apply`; the pod will remain in a `Pending` state, and they must inspect Pod events to see the `NodePrepareResources` failure.
+  * **Kubelet Retry Loops**: Standard Kubernetes behavior is to retry failed resource preparations. A persistent denial (like a 400 Bad Request) will cause the Kubelet to continuously retry `NodePrepareResources`, which can generate unnecessary load on the node and webhook server compared to an upfront API rejection.
+
```
kanlkan, line 110: I would like to add one more sentence here.

```
  * **Idempotency**: The kubelet may retry `NodePrepareResources`, so DRANET can call this more than once for the same `(device, claimUID)`. It must return an equivalent result without allocating additional resources (e.g. key the allocation by `claimUID`, as `whereabouts` does via `CNI_CONTAINERID`).
```

site/content/docs/contributing/webhook-providers.md
```diff
@@ -0,0 +1,121 @@
...
+* `POST /ReleaseProfileConfig`: Frees stateful resources (e.g., releasing an IP address). Also receives the full `NetworkConfig`. Should return `200 OK` on success or if the resource was already released (idempotency).
+
```
kanlkan, line 113: I would like to add one more sentence here.

```
  * **Best-effort teardown**: A failed `ReleaseProfileConfig` is logged but not retried by DRANET (teardown must not block pod deletion). The provider therefore owns leak reclamation and must be able to garbage-collect orphaned allocations on its own, otherwise resources leak permanently.
```

kanlkan (commented):
Thank you for opening this PR. I added some comments for the document.

site/content/docs/contributing/webhook-providers.md
aojea: added

site/content/docs/contributing/webhook-providers.md
aojea: added

kanlkan (approved):
Thank you!
/lgtm

PR #230, opened 2026-06-18
feat(driver): persist allocated device attributes in ResourceSlice 

### LINKED ISSUE #99: [question] Clarification on ResourceSlice behavior discussed in google/dranet#97
Hi,

While looking into DRANET’s ResourceSlice handling, I came across the discussion in the previous repository:

https://github.com/google/dranet/issues/97

From that thread, my understanding is that DRANET currently publishes only devices visible in the root network namespace. When a device is allocated, the interface is moved into the pod's namespace, which effectively removes it from the ResourceSlice.

The issue also mentioned a possible alternative approach where all devices would remain listed in the ResourceSlice, with allocation state determined via ResourceClaims instead.

Since the project has moved to kubernetes-sigs/dranet and a year has passed, I wanted to check whether this behaviour is still the intended design, or if the thinking around this has evolved.

Thanks for any clarification!

/kind question

purvavj:
#### What type of PR is this?

/kind feature

When network devices are allocated to a pod, they move to the pod's network namespace and disappear from host scans, causing them to vanish from the published ResourceSlice.
This PR resolves the issue by snapshotting allocated device attributes in the local BoltDB database during allocation (prepareResourceClaim) and merging them back into the published ResourceSlice list during the background publishing loop (PublishResources).
Fixes kubernetes-sigs/dranet#99

```release-note
Added the `PersistentResourceSliceAttributes` alpha feature gate (disabled by default) to preserve device attributes in ResourceSlices across daemon restarts. To enable this feature, pass `--feature-gates=PersistentResourceSliceAttributes=true` as an argument to the driver container.
```


gauravkghildiyal:
Thanks for the change @purvavj 

Haven't reviewed the code, but couple of requests:
- Please add e2e tests.
- Please update release note in PR description. 

/ok-to-test

tamilmani1989:
how this work for virtual devices.. virtual devices will get deleted with deletion of pod.. in that case, will it never get removed from db?

purvavj:
> how this work for virtual devices.. virtual devices will get deleted with deletion of pod.. in that case, will it never get removed from db?

They actually do get cleaned up! During pod deletion, kubelet calls unprepareResourceClaim. This moves the virtual device back into the host namespace (via nsDetachNetdev) and deletes the pod configuration from BoltDB (via DeleteClaim). 
Since the device is no longer in BoltDB and is back on the host namespace, the background scan detects it and publishes it back to the ResourceSlice as a clean, available device again.

pkg/driver/dra_hooks.go
```diff
@@ -433,6 +442,13 @@ func (np *NetworkDriver) prepareResourceClaim(ctx context.Context, claim *resour
 				klog.Infof("error unpinning ebpf programs for %s : %v", ifName, err)
 			}
 		}
+		// Query the local discovery database (netdb) for the card's clean attributes
+		device, ok := np.netdb.GetDevice(result.Device)
+		if ok {
+			deviceCfg.Device = &device
```
gauravkghildiyal: We should setup the Device within deviceCfg as early as possible in the function. For example, at the moment, there's a place earlier in the function which also stores the config in the podConfigStore which would be missing the Device struct.

pkg/driver/dra_hooks.go
```diff
@@ -61,13 +62,21 @@ func (np *NetworkDriver) PublishResources(ctx context.Context) {
...
 			devices = filter.FilterDevices(np.celProgram, devices)
-			klog.V(3).Infof("After filtering, publishing %d devices in ResourceSlice(s): %s", len(devices), formatDeviceNames(devices, 15))
 
-			np.publishResourcesPrometheusMetrics(devices)
+			// Fetch allocated devices from BoltDB store and merge
+			var allocated []resourceapi.Device
+			if np.podConfigStore != nil {
+				allocated = np.podConfigStore.GetAllocatedDevices()
+			}
+			merged := mergeDevices(devices, allocated)
```
gauravkghildiyal: Merging of the devices should happen before the Filtering happens. Filtering should happen on the _almost_ final set of devices.

pkg/driver/dra_hooks.go
```diff
@@ -670,3 +686,58 @@ func (np *NetworkDriver) getDeviceNetworkConfig(device string, claimUID types.UI
 	}
 	return mergedConf, nil
 }
+
+func mergeDevices(available, allocated []resourceapi.Device) []resourceapi.Device {
```
gauravkghildiyal: I'm not sure if we really need to differentiate things like a physical device or related complexity at this point.

Can we instead have a more thorough _merging_ (and not overwriting) behaviour of the various attributes/capacities (or anything within the Device struct), whereby the only criteria for merging is that precedence is given to data from `available` when a choice needs to be made?

gauravkghildiyal (commented):
Thanks @purvavj 

Sharing some initial thoughts here.

In addition, can we give some thought into any possible race conditions here during the pod deletion lifecycle? Things like can there be a point where there's a stale device in the ResourceSlice which may no longer exist.

pkg/driver/pod_device_config.go
```diff
@@ -47,6 +48,10 @@ type PodConfig struct {
 type DeviceConfig struct {
 	Claim types.NamespacedName `json:"claim"`
 
+	// Device contains the original discovered ResourceSlice Device structure,
+	// which includes the device's identifying attributes and capacity.
+	Device *resourceapi.Device `json:"device,omitempty"`
```
gauravkghildiyal: Can we rename to something more descriptive like `DeviceSnapshot`, alternatives welcomed :)

purvavj: I agree, DeviceSnapshot is much clearer.

pkg/driver/dra_hooks.go
purvavj: Yes, I have simplified this completely by removing the physical/virtual device checks. We now do a generic map merge of attributes and capacities, using the database snapshot as a base and overwriting it with any live scan data. I've updated the unit tests to match.

pkg/driver/dra_hooks.go
purvavj: makes sense; swapped the order so we now run mergeDevices first and then apply FilterDevices on the merged list before publishing.

tests/e2e.bats
```diff
@@ -576,3 +576,83 @@ EOF
...
+  docker exec "$NODE_NAME" bash -c "ip link add $DUMMY_IFACE type dummy"
+  docker exec "$NODE_NAME" bash -c "ip link set up dev $DUMMY_IFACE"
+
+  # 2. Wait for it to be discovered and published to the ResourceSlice
+  sleep 5
+  run kubectl get resourceslices -o jsonpath='{.items[*].spec.devices[*].name}'
+  assert_success
+  assert_output --partial "$DUMMY_IFACE"
+
+  # Retrieve its original attributes before allocation
```
gauravkghildiyal, line 679: Checking individual attributes won't be future-proof. We should compare the whole set of attributes.

tests/e2e.bats
```diff
@@ -576,3 +576,83 @@ EOF
   run kubectl exec -it $POD_1 -- ping -I eth1 -c 1 10.10.20.1
   assert_success
 }
+
+@test "allocated device attributes are persisted in ResourceSlice across daemon restart" {
```
gauravkghildiyal: In particular what we care about is that even for allocated devices, the ResourceSlice remains the same. Across restarts is a natural extension to that. Let's rename accordingly.

pkg/driver/pod_device_config.go
```diff
@@ -309,3 +314,20 @@ func (s *PodConfigStore) DeleteClaim(claim types.NamespacedName) []types.UID {
 	}
 	return podsToDelete
 }
+
+// GetAllocatedDevices returns all devices currently allocated to active pods
+// that have a valid device attributes snapshot stored in BoltDB.
+func (s *PodConfigStore) GetAllocatedDevices() []resourceapi.Device {
```
gauravkghildiyal: nit: Consider naming GetAllocatedDeviceSnapshots to make the intent clear.

pkg/driver/dra_hooks_test.go
```diff
@@ -853,3 +853,104 @@ func TestGetDeviceNetworkConfigWithWebhook(t *testing.T) {
...
+	}
+
+	pciDev := resourcev1.Device{
+		Name: "0000:c0:14.0",
+		Attributes: map[resourcev1.QualifiedName]resourcev1.DeviceAttribute{
+			resourcev1.QualifiedName(apis.AttrPCIAddress): mockAttr("0000:c0:14.0"),
+		},
+	}
+
+	pciDevDegraded := resourcev1.Device{
```
gauravkghildiyal: Is there a difference between pciDevDegraded and pciDev?

pkg/driver/dra_hooks_test.go
```diff
@@ -853,3 +853,104 @@ func TestGetDeviceNetworkConfigWithWebhook(t *testing.T) {
 		})
 	}
 }
+
+func TestMergeDevices(t *testing.T) {
+	mockAttr := func(val string) resourcev1.DeviceAttribute {
```
gauravkghildiyal: nit: Consider naming `stringAttr` to represent that this constructs a String attribute.

pkg/driver/dra_hooks_test.go
```diff
@@ -853,3 +853,104 @@ func TestGetDeviceNetworkConfigWithWebhook(t *testing.T) {
...
+			allocated: []resourcev1.Device{pciDevSnapshot},
+			expected:  []resourcev1.Device{pciDevSnapshot},
+		},
+		{
+			name:      "Virtual device preserve",
+			available: nil,
+			allocated: []resourcev1.Device{virtualDevSnapshot},
+			expected:  []resourcev1.Device{virtualDevSnapshot},
+		},
+		{
```
gauravkghildiyal, line 900: For the mergeDevices() function, I don't think it treats the two cases `Virtual device preserve` and `Physical device missing from host (Ghost device)` any differently right? If yes, let's avoid the duplication and avoid any distinction of physical / virtual.

pkg/driver/dra_hooks.go
```diff
@@ -670,3 +690,59 @@ func (np *NetworkDriver) getDeviceNetworkConfig(device string, claimUID types.UI
...
+	for _, dev := range merged {
+		result = append(result, dev)
+	}
+	sort.Slice(result, func(i, j int) bool {
+		return result[i].Name < result[j].Name
+	})
+	return result
+}
+
+func mergeDeviceStructs(live, snap resourceapi.Device) resourceapi.Device {
```
gauravkghildiyal, line 731: 1. I think this modifies the state of live and snap itself which we'd like to avoid doing.

2. Let's also use the live as base and then layer snapshot on top of it. That should help preserve any other fields of the Live object, and also serves as the mental model we're aiming for (i.e. use Live, and then rely on attributes where somethings missing)

pkg/driver/pod_device_config_test.go
```diff
@@ -407,3 +408,44 @@ func TestPodConfigStore_NoDuplicateDevices(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to set device config: %v", err)
+	}
+
+	// Verify we get exactly the snapshot device
+	allocated = store.GetAllocatedDevices()
+	if len(allocated) != 1 {
+		t.Fatalf("expected 1 allocated device, got %d", len(allocated))
+	}
+	if allocated[0].Name != "0000:c0:14.0" {
```
gauravkghildiyal: Prefer comparing the complete struct, rather than just the name.

(You can use cmp.Diff to perform the comparison, you can find examples in this repo)

pkg/driver/pod_device_config_test.go
```diff
@@ -407,3 +408,44 @@ func TestPodConfigStore_NoDuplicateDevices(t *testing.T) {
...
+
+	// Verify it returns nothing if no device snapshots are stored
+	allocated := store.GetAllocatedDevices()
+	if len(allocated) != 0 {
+		t.Errorf("expected 0 allocated devices, got %d", len(allocated))
+	}
+
+	// Set config with device snapshot
+	snapDev := resourceapi.Device{
+		Name: "0000:c0:14.0",
```
gauravkghildiyal, line 436: Lets populate the attributes and capacities fields too.

pkg/driver/dra_hooks_test.go
purvavj: Yeah, they are handled exactly the same way. Unified them into a single "Allocated device missing from host" case.

pkg/driver/dra_hooks_test.go
purvavj: No difference, they were identical. Originally pciDev was meant to have more fields, but after some refactoring they both ended up with just the PCI address. Cleaned it up.

pkg/driver/dra_hooks_test.go
```diff
@@ -853,3 +853,83 @@ func TestGetDeviceNetworkConfigWithWebhook(t *testing.T) {
...
+			expected:  []resourcev1.Device{pciDev},
+		},
+		{
+			name:      "Device merge",
+			available: []resourcev1.Device{pciDev},
+			allocated: []resourcev1.Device{pciDevSnapshot},
+			expected:  []resourcev1.Device{pciDevSnapshot},
+		},
+		{
+			name:      "Allocated device missing from host (Ghost device)",
```
gauravkghildiyal: nit: Please remove "(Ghost device)" (Doesn't add value and can be confusing)

pkg/driver/dra_hooks_test.go
```diff
@@ -853,3 +853,83 @@ func TestGetDeviceNetworkConfigWithWebhook(t *testing.T) {
...
+			available: nil,
+			allocated: []resourcev1.Device{pciDevSnapshot},
+			expected:  []resourcev1.Device{pciDevSnapshot},
+		},
+		{
+			name: "Device back on host (Zombie override)",
```
gauravkghildiyal: It's unclear to me what this test is trying to do and how it is different from Device merge case. Can you please double check the intent please?

Also, a minor nitpick on the naming: it would be helpful to avoid describing the test cases in terms of the "host" or "zombies". Think of it this way that the mergeDevices function is simply a pure data merging function that takes two slices (available and allocated) and resolves them with a specific precedence. The test names should ideally reflect the data merging behavior (e.g., "Available device attribute takes precedence over snapshot"). Comment applicable to other tests names too.

pkg/driver/dra_hooks_test.go
```diff
@@ -853,3 +853,83 @@ func TestGetDeviceNetworkConfigWithWebhook(t *testing.T) {
...
+				Name: "0000:c0:14.0",
+				Attributes: map[resourcev1.QualifiedName]resourcev1.DeviceAttribute{
+					resourcev1.QualifiedName(apis.AttrPCIAddress):    stringAttr("0000:c0:14.0"),
+					resourcev1.QualifiedName(apis.AttrInterfaceName): stringAttr("eth1"), // back on host!
+				},
+			}},
+			allocated: []resourcev1.Device{pciDevSnapshot},
+			expected:  []resourcev1.Device{pciDevSnapshot},
+		},
+	}
```
gauravkghildiyal, line 946: Please add test cases to verify correct precedence is given during merging to liveDevices versus snapshots.

pkg/driver/dra_hooks.go
```diff
@@ -670,3 +690,68 @@ func (np *NetworkDriver) getDeviceNetworkConfig(device string, claimUID types.UI
 	}
 	return mergedConf, nil
 }
+
+func mergeDevices(available, allocated []resourceapi.Device) []resourceapi.Device {
```
gauravkghildiyal: Please consider naming the parameters `live` and `snapshots` for better differentiation. That also keep things consistent with what they are called in merge struct.

pkg/driver/dra_hooks_test.go
```diff
@@ -853,3 +853,83 @@ func TestGetDeviceNetworkConfigWithWebhook(t *testing.T) {
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			result := mergeDevices(tc.available, tc.allocated)
+			if len(result) != len(tc.expected) {
+				t.Fatalf("expected %d devices, got %d", len(tc.expected), len(result))
+			}
+			for i, dev := range result {
+				if dev.Name != tc.expected[i].Name {
```
gauravkghildiyal: Similar to the other test, let's just use cmp.Diff to compare the whole thing?

pkg/driver/dra_hooks.go
gauravkghildiyal: You can make use of DeepCopy and simplify the code.

```
live := live.DeepCopy()
snap := snap.DeepCopy()
```

gauravkghildiyal (commented):
Took another pass. Thanks @purvavj we're almost there. 

Signal boosting my comment from last time regarding proneness of a race condition in such cases (e.g. Device is still in ResourceSlice when it was supposed to have been removed), I think what we should do is feature gate this initially so we have better control over being able to disable things in case of issues. I'll send a PR for some basic feature gate scaffolding and you can use that in this PR.

pkg/driver/dra_hooks_test.go
purvavj: You're right, treating it as a pure data merge is much clearer. I've dropped the redundant test and renamed the others to focus on live vs. snapshot behavior. I also added a test for precedence and switched to cmp.Diff.

purvavj:
> Took another pass. Thanks @purvavj we're almost there.
> 
> Signal boosting my comment from last time regarding proneness of a race condition in such cases (e.g. Device is still in ResourceSlice when it was supposed to have been removed), I think what we should do is feature gate this initially so we have better control over being able to disable things in case of issues. I'll send a PR for some basic feature gate scaffolding and you can use that in this PR.

Great call on the feature gate and thanks for the code pointer! I've added the feature gate PersistentResourceSliceAttributes (disabled by default) and wrapped the device merging logic behind it. I also set up the --feature-gates command-line flag in app.go so it can be parsed on startup.

pkg/features/features.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"k8s.io/component-base/featuregate"
+)
+
+const (
+	// PersistentResourceSliceAttributes gates the persistence of network
+	// attributes (like MAC, MTU, etc.) in the ResourceSlice across daemon restarts.
+	// owner: @purvavj
+	// alpha: v0.1.0
```
gauravkghildiyal: The version here indicates which DRANET release version was this made alpha in. This would be the next release

pkg/features/features.go
```diff
@@ -0,0 +1,33 @@
+package features
+
+import (
+	"k8s.io/component-base/featuregate"
+)
+
+const (
+	// PersistentResourceSliceAttributes gates the persistence of network
```
gauravkghildiyal: nit: 'network' seems redundant?

pkg/driver/dra_hooks.go
```diff
@@ -58,16 +60,30 @@ func (np *NetworkDriver) PublishResources(ctx context.Context) {
...
+		case live := <-np.netdb.GetResources(ctx):
+			klog.V(3).Infof("Got %d devices from inventory: %s", len(live), formatDeviceNames(live, 15))
+
+			// Fetch allocated devices from BoltDB store and merge
+			var snapshots []resourceapi.Device
+			if np.podConfigStore != nil {
+				snapshots = np.podConfigStore.GetAllocatedDeviceSnapshots()
+			}
+			merged := live
+			if features.DefaultFeatureGate.Enabled(features.PersistentResourceSliceAttributes) {
```
gauravkghildiyal, line 69: Can we simplify things to:

```golang
			merged := live
			if features.DefaultFeatureGate.Enabled(features.PersistentResourceSliceAttributes) {
				snapshots := np.podConfigStore.GetAllocatedDeviceSnapshots()
				merged = mergeDevices(live, snapshots)
			}

```

pkg/driver/dra_hooks.go
```diff
@@ -58,16 +60,30 @@ func (np *NetworkDriver) PublishResources(ctx context.Context) {
...
-			klog.V(3).Infof("Got %d devices from inventory: %s", len(devices), formatDeviceNames(devices, 15))
-			devices = filter.FilterDevices(np.celProgram, devices)
-			klog.V(3).Infof("After filtering, publishing %d devices in ResourceSlice(s): %s", len(devices), formatDeviceNames(devices, 15))
+		// Wait for updates from the host-discovered (live available) device inventory
+		case live := <-np.netdb.GetResources(ctx):
+			klog.V(3).Infof("Got %d devices from inventory: %s", len(live), formatDeviceNames(live, 15))
+
+			// Fetch allocated devices from BoltDB store and merge
+			var snapshots []resourceapi.Device
+			if np.podConfigStore != nil {
```
gauravkghildiyal: Checking podConfigStore is nil is unnecessary and odd.

You do expect podConfigStore to NOT be nil. If it's nil, something's seriously wrong with the code. Generally you would do nil check podConfigStore was an optional thing.

pkg/driver/dra_hooks_test.go
```diff
@@ -853,3 +855,102 @@ func TestGetDeviceNetworkConfigWithWebhook(t *testing.T) {
...
+	}
+
+	tests := []struct {
+		name     string
+		live     []resourcev1.Device
+		snapshot []resourcev1.Device
+		expected []resourcev1.Device
+	}{
+		{
+			name:     "Only available devices returned",
```
gauravkghildiyal: Let's update the test names as well with consistent 'live' and 'snapshot' terminology, instead of 'available' and 'allocated'

pkg/driver/dra_hooks.go
purvavj: oh right! NewPodConfigStore always returns a valid store, so this check makes no sense.

pkg/features/features.go
purvavj: yea, changing that to 'device' would make more sense.

pkg/driver/dra_hooks.go
purvavj: Thanks for pointing this out! I've removed the nil check and simplified the flow.

gauravkghildiyal:
Thanks!

/lgtm
/approve

PR #231, opened 2026-06-18
fix invalid ifname and add a test for webhook-whereabouts

### LINKED ISSUE #225: webhook-whereabouts: CNI_IFNAME uses PCI-derived device name, breaking IPAM for SR-IOV VFs
# Summary
webhook-whereabouts passes req.Device.Name verbatim as CNI_IFNAME when invoking the whereabouts IPAM plugin:

```
// cmd/webhook-whereabouts/main.go (ADD and DEL paths)
"CNI_IFNAME=" + req.Device.Name,
```

For PCI devices such as SR-IOV VFs, Device.Name is PCI-derived (pkg/names.NormalizePCIAddress: 0000:27:00.2 → pci-0000-27-00-2, 16 chars). This exceeds the CNI interface-name limit (IFNAMSIZ - 1 = 15), so CNI's interface-name validation rejects it and the IPAM ADD fails — IP allocation does not work for these devices.

# Root cause

`CNI_IFNAME` must satisfy CNI's `ValidateInterfaceName` (1–15 bytes; not `./..`; no `/`, `:`, or whitespace). A 16-char PCI-derived name violates the length rule.

# Proposed fix
Derive `CNI_IFNAME` deterministically from the stable device identifier (`Device.Name`, the inventory store key, which is the immutable PCI-derived identity) instead of the mutable kernel ifName. This makes (2) hold by construction, independent of rescans or namespace moves.

The 16-char overflow for PCI devices is caused solely by the constant `pci-` prefix; the bare address (`0000-27-00-2`, ≤12 chars for a standard `DDDD:BB:DD.F` address) is already valid. So we keep the value human-readable wherever possible — important so a leaked reservation can still be traced back to its device during manual pool cleanup:

* `Device.Name` already valid (short, non-PCI names) → used verbatim;
* PCI name → strip the `pci-` prefix (`pci-0000-27-00-2` → `0000-27-00-2`), re-validating the result so we never emit an over-length name on the assumption it fit;
* anything still invalid after the above (e.g. base32-encoded non-DNS-1123 names, already unreadable upstream) → a deterministic hash as a last resort.

# Sample implementation

https://github.com/kanlkan/dranet/tree/fix/webhook-ifname

# Environment
dranet: build with main branch (commit [66ad100](https://github.com/kubernetes-sigs/dranet/tree/66ad100f484d31269d1880dce0270761cd942807))
whereabouts: v0.9.4
Device type: SR-IOV VF (PCI-derived device name)

kanlkan:
#### What type of PR is this?

/kind bug

webhook-whereabouts passes req.Device.Name verbatim as CNI_IFNAME when invoking the whereabouts IPAM plugin.
For PCI devices such as SR-IOV VFs, Device.Name is PCI-derived (pkg/names.NormalizePCIAddress: 0000:27:00.2 → pci-0000-27-00-2, 16 chars). This exceeds the CNI interface-name limit (IFNAMSIZ - 1 = 15), so CNI's interface-name validation rejects it and the IPAM ADD fails — IP allocation does not work for these devices.

This PR fixes the bug using the following approach:
* Device.Name already valid (short, non-PCI names) → used verbatim;
* PCI name → strip the pci- prefix (pci-0000-27-00-2 → 0000-27-00-2), re-validating the result so we never emit an over-length name on the assumption it fit;
* anything still invalid after the above (e.g. base32-encoded non-DNS-1123 names, already unreadable upstream) → a deterministic hash as a last resort.

Fixes #225 

None.

```release-note
NOTE
```

### DIFF
--- a/cmd/webhook-whereabouts/main.go
+++ b/cmd/webhook-whereabouts/main.go
@@ -58,11 +58,6 @@ func (s *Server) GetDeviceConfig(w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte("{}"))
 }
 
-// maxCNIIfnameLen is the maximum interface-name length accepted by the CNI
-// runtime (IFNAMSIZ-1). Names longer than this are rejected by CNI's
-// interface-name validation.
-const maxCNIIfnameLen = 15
-
 // pciNamePrefix is the prefix dranet prepends to a normalized PCI address
 // (pkg/names.NormalizePCIAddress: "0000:8a:00.0" -> "pci-0000-8a-00-0").
 const pciNamePrefix = "pci-"
@@ -103,7 +98,7 @@ func cniIfname(req webhook.ProfileRequest) string {
 // interface-name validation. It mirrors github.com/containernetworking/cni
 // pkg/utils.ValidateInterfaceName.
 func isValidCNIIfname(s string) bool {
-	if len(s) == 0 || len(s) > maxCNIIfnameLen {
+	if len(s) == 0 || len(s) > apis.MaxInterfaceNameLen {
 		return false
 	}
 	if s == "." || s == ".." {
--- a/cmd/webhook-whereabouts/main.go
+++ b/cmd/webhook-whereabouts/main.go
@@ -21,11 +21,14 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"hash/fnv"
 	"log"
 	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
+	"unicode"
 
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	types100 "github.com/containernetworking/cni/pkg/types/100"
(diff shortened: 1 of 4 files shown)


cmd/webhook-whereabouts/main.go
```diff
@@ -55,6 +58,65 @@ func (s *Server) GetDeviceConfig(w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte("{}"))
 }
 
+// maxCNIIfnameLen is the maximum interface-name length accepted by the CNI
+// runtime (IFNAMSIZ-1). Names longer than this are rejected by CNI's
+// interface-name validation.
+const maxCNIIfnameLen = 15
```
tamilmani1989: can we use `apis.MaxInterfaceNameLen` instead of defining this again?

kanlkan: Thank you for your review, reused apis.MaxInterfaceNameLen, dropped the duplicate constant. [e693aee](https://github.com/kubernetes-sigs/dranet/pull/231/commits/e693aee78fe672b4d87949b9553f6035ade77258)

tamilmani1989 (approved):

aojea:
/lgtm
/approve

> // whereabouts matches a reservation for release on (CNI_CONTAINERID, CNI_IFNAME)

not familiar with wherabouts but if that is the key the risk of collision is minimal, since all ifnames that can collide need to belong to the same containerid

PR #233, opened 2026-06-19
remove pruning from Synchronize

### LINKED ISSUE #232: [Reliability] Race condition in Synchronize prunes pod network configs aggressively during startup
### Behavior
When a pod is scheduled immediately after the `dranet` driver starts up, the pod starts successfully but is missing its allocated device configuration.

### Race Condition
There is a race condition when the NRI `Synchronize` function runs and prunes the pod's device configuration, before the driver's `RunPodSandbox` NRI hook is called.

- `PrepareResourceClaims` is first called when the new pod is scheduled, and the driver allocates the configuration and writes it to the bbolt persistent storage.
- Because the driver just started, containerd registers the driver's NRI plugin and triggers `Synchronize` to align the database state with the runtime.
- By now`RunPodSandbox` is not yet called for the new pod, the pod is missing from the runtime's sandbox list, and the driver eager-prunes the configuration from the database: [nri_hooks.go](https://github.com/kubernetes-sigs/dranet/blob/main/pkg/driver/nri_hooks.go#L42C1-L64C4)
- When containerd finally executes the NRI `RunPodSandbox`, the configuration is missing from the database, and interface injection is skipped.

### Version affected
Starting from dranet v1.2.0

ngcxy:
#### What type of PR is this?

/kind bug

A race condition occurs when right after driver startup, the NRI `Synchronize` function prunes a pod's device configuration which is just stored for a new pod, before the `RunPodSandbox` hook is actually executed. This causes the network device injection to be skipped.

This change mitigates the issue by removing the pruning step from `Synchronize` NRI hook, because its functionality can be fully handled by Kubelet DRA manager calling to the `UnprepareResourceClaims`.

Fixes #232 

Key changes:
- Remove startup stale configuration pruning from NRI Synchronize.
- Rely on native Kubelet call to UnprepareResourceClaims for pod device cleanup.
- Add E2E lifecycle test verifying cleanup on driver restart after pod deletion during downtime.

```release-note
This PR fixes a race condition appears in dranet v1.2.0+ that NRI `Synchronize` function prunes the config store for pending pods during driver startup, causing pod interface injection being skipped (#232).
```

### DIFF
--- a/tests/e2e.bats
+++ b/tests/e2e.bats
@@ -424,6 +424,85 @@ EOF
   kubectl rollout status ds/dranet --namespace=kube-system
 }
 
+@test "unprepare claims on driver restart after pod deletion during driver downtime" {
+  local NODE_NAME="$CLUSTER_NAME"-worker
+  local DUMMY_IFACE="dummy-downtime"
+
+  docker exec "$NODE_NAME" bash -c "ip link add $DUMMY_IFACE type dummy"
+  docker exec "$NODE_NAME" bash -c "ip link set up dev $DUMMY_IFACE"
+
+  kubectl apply -f "$BATS_TEST_DIRNAME"/../tests/manifests/deviceclass.yaml
+  kubectl apply -f "$BATS_TEST_DIRNAME"/../tests/manifests/resourceclaim.yaml
+  
+  kubectl wait --timeout=30s --for=condition=ready pods -l app=pod
+
+  local POD_NAME
+  POD_NAME=$(kubectl get pods -l app=pod -o name)
+  local POD_UID
+  POD_UID=$(kubectl get "$POD_NAME" -o jsonpath='{.metadata.uid}')
+
+  # Turn down the dranet driver on NODE_NAME by scheduling it off
+  kubectl label node "$NODE_NAME" e2e-test-do-not-schedule=true
+  kubectl patch daemonset dranet -n kube-system --type='merge' --patch-file=<(cat <<EOF
+spec:
+  template:
+    spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: e2e-test-do-not-schedule
+                operator: DoesNotExist
+EOF
+)
+  kubectl rollout status ds/dranet --namespace=kube-system
+
+  # Delete the pod forcefully. Verify the pod is immediately removed from the API server.
+  kubectl delete "$POD_NAME" --force
+  run kubectl get "$POD_NAME"
+  assert_failure
+
+  # Inspect the bbolt database for the deleted pod configs. The pod configs should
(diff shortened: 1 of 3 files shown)


gauravkghildiyal:
/ok-to-test

Thanks for the fix @ngcxy. I'm actually wondering do we even need pruning during startup? Isn't the UnprepareResourceClaims sufficient? Needs some thought.

ngcxy:
Thanks for the insight! My understanding based on PR https://github.com/kubernetes-sigs/dranet/pull/115 is that pruning helps removing stale config store if a pod is deleted when driver is down, and cleanup checkpoint for stale config store during startup.

While UnprepareResourceClaims handles a similar cleanup task for pod config stores, I'm not entirely sure if it is triggered during startup to fully cover these edge cases (seems like it's not?).

gauravkghildiyal:
> pruning helps removing stale config store if a pod is deleted when driver is down

This is the exactly the premise that I'm trying to question the likelihood of. I believe while DRANET is down, the Kubelet DRA manager will continue retrying the UnprepareResourceClaims call until it gets delivered. Even if the pod is force deleted, the UnprepareResourceClaims call likely is expected to be received by DRANET upon its restart. **Can we experiment with this a bit to see how the system behaves when DRANET is down?**

* We generally want to be robust, but in this case, it seems like it is highly unlikely for the podConfig to NOT automatically get cleaned up through UnpreapreResourceClaim. 
* At the same time, even if we do leak the pod within the podConfig in some rare occurrence, it very likely isn't going to impact things in a negative way. 

So based on the results of the experiment, I would lean towards removing the pruning from Synchronize. That helps eliminate the race itself, while keeping the system simple and not dependent on another heuristic.

ngcxy:
I just ran the experiment, and it behaves exactly as you described! Even with a pod deleted during downtime, Kubelet retries UnprepareResourceClaims until DRANET restarts, which then automatically cleans up the podConfig with no need to explicitly remove the store.

I totally agree with keeping the system simple. I'll go ahead and remove the pruning logic from Synchronize :)

tests/e2e.bats
```diff
@@ -424,6 +424,71 @@ EOF
   kubectl rollout status ds/dranet --namespace=kube-system
 }
 
+@test "unprepare claims on driver restart after pod deletion during driver downtime" {
```
gauravkghildiyal, line 427: Great job adding a test for this!

Couple of suggestions:
- Instead of normal pod delete, can we try to instead verify the more edge case involving a force delete?
- It would be more appropriate if the test could directly try to read the bbolt db to verify its state change as a result of a successful UnprepareResourceClaim. That way you can also stop relying on trying to parse this through the logs.

ngcxy: Thanks for the suggestions! I have made the modifications: updating the pod deletion to be force delete, and changing the test verification to rely on bbolt database inspect.

ngcxy: Also refactored the test to inspect the database during driver downtime.

Issue #228, opened 2026-06-16
Add a method to discover RDMA devices missed by PCI class filter


Bowser1704:
### Problem

dranet discovers network devices by scanning PCI devices with class `0x02` (Network controller) in `discoverPCIDevices()`. RDMA devices whose PCI class is not `0x02` are invisible to this scan.

For example, Alibaba Cloud eRDMA devices use PCI class `0x00ff` (Unclassified):

```
$ lspci -nn -d 1ded:107f
1b:00.0 Unclassified device [00ff]: Alibaba (China) Co., Ltd. Elastic RDMA Adapter [1ded:107f]
9d:00.0 Unclassified device [00ff]: Alibaba (China) Co., Ltd. Elastic RDMA Adapter [1ded:107f]
```

These are fully functional RDMA devices registered with the kernel's ib_core subsystem:

```
$ ls -la /sys/class/infiniband/
erdma_0 -> ../../devices/pci0000:16/.../0000:1b:00.0/infiniband/erdma_0
erdma_1 -> ../../devices/pci0000:96/.../0000:9d:00.0/infiniband/erdma_1

$ rdma dev show
4: erdma_0: node_type ca protocol roce ...
5: erdma_1: node_type ca protocol roce ...
```

### Background

`/sys/class/infiniband/` is populated by the kernel's ib_core module whenever any RDMA driver calls `ib_register_device()`. Every entry is a symlink into `/sys/devices/` with a `device` back-link to the underlying PCI device. This works regardless of PCI class code — ConnectX (`0x02`), EFA (`0x02`), eRDMA (`0x00ff`) all appear here.

### Proposal

**No changes to existing discovery logic.** Add a supplementary scan step after `discoverRDMADevices()` that:

1. Lists `/sys/class/infiniband/` entries
2. Resolves each entry's `device` symlink to get the PCI BDF address
3. **Skips any device whose PCI address is already known** from the existing PCI class `0x02` scan (e.g. ConnectX, EFA)
4. For remaining devices (the ones the PCI class filter missed), publishes them as IB-only RDMA devices with `pciAddress`, `rdma`, `rdmaDevice`, and `pcieRoot` attributes

This is a pure additive change — existing device discovery is untouched, the new step only picks up devices that would otherwise be invisible.

gauravkghildiyal:
Thanks for reporting this @Bowser1704. This would certainly be a welcomed improvement. Would you be interested in taking this?

[...]

PR #238, opened 2026-06-23
inventory: discover standalone RDMA devices via /sys/class/infiniband

Bowser1704:
#### What type of PR is this?
/kind feature

discover standalone RDMA devices via /sys/class/infiniband

#228 

```release-note
discover standalone RDMA devices via /sys/class/infiniband
```

### DIFF
--- a/pkg/inventory/db.go
+++ b/pkg/inventory/db.go
@@ -196,8 +196,9 @@ func (db *DB) Run(ctx context.Context) error {
 // filters out default interfaces, and updates the device store.
 func (db *DB) scan() []resourceapi.Device {
 	devices := db.discoverPCIDevices()
+	devices = db.discoverStandaloneRDMADevices(devices)
 	devices = db.discoverNetworkInterfaces(devices)
-	devices = db.discoverRDMADevices(devices)
+	devices = db.addRDMAAttributes(devices)
 	devices = db.addCloudAttributes(devices)
 
 	// Remove default interface.
@@ -322,7 +323,7 @@ func (db *DB) discoverNetworkInterfaces(pciDevices []resourceapi.Device) []resou
 
 		// When moveIBInterfaces is false, skip IPoIB interfaces.
 		// The underlying PCI device will be discovered as an IB-only RDMA
-		// device (no netdev) via discoverRDMADevices. Associating the IPoIB
+		// device (no netdev) via addRDMAAttributes. Associating the IPoIB
 		// netdev with the PCI device would mask the IB-only nature of the
 		// device and prevent correct RDMA char device injection into pods.
 		// When moveIBInterfaces is true (default), IPoIB interfaces
@@ -489,7 +490,7 @@ func addLinkAttributes(device *resourceapi.Device, link netlink.Link) {
 	}
 }
 
-func (db *DB) discoverRDMADevices(devices []resourceapi.Device) []resourceapi.Device {
+func (db *DB) addRDMAAttributes(devices []resourceapi.Device) []resourceapi.Device {
 	for i := range devices {
 		isRDMA := false
 		if ifName := devices[i].Attributes[apis.AttrInterfaceName].StringValue; ifName != nil && *ifName != "" {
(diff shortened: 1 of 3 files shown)


pkg/inventory/db.go
```diff
@@ -198,6 +198,7 @@ func (db *DB) scan() []resourceapi.Device {
 	devices := db.discoverPCIDevices()
 	devices = db.discoverNetworkInterfaces(devices)
 	devices = db.discoverRDMADevices(devices)
+	devices = db.discoverStandaloneRDMADevices(devices)
```
tamilmani1989, line 199: should this be part of discoverRDMADevices itself instead of separate call? don't have strong opinion, just asking if we can avoid redundancy of discoverRDMADevices

Bowser1704: Good call — you're right that keeping them separate added redundancy. 

`discoverRDMADevices` was misnamed anyway since it only annotates existing devices rather than discovering new ones.

Reorganized the pipeline so the three discovery passes run first, then a single attribute pass covers everything uniformly:

```
discoverPCIDevices
discoverNetworkInterfaces
discoverStandaloneRDMADevices   ← moved up
addRDMAAttributes               ← renamed; now handles standalone devices too
addCloudAttributes
```

Standalone devices now just create a bare entry with the PCI address; `addRDMAAttributes` picks them up via the PCI-address branch the same way it handles IB-only PCI devices.

Bowser1704:
@gauravkghildiyal @tamilmani1989 Could you help review this when you have a chance? This implements #228.

pkg/inventory/sysfs.go
```diff
@@ -293,3 +293,35 @@ func pciAddressForNetInterface(ifName string) (*pciAddress, error) {
 	}
 	return address, nil
 }
+
+const sysInfinibandPath = "/sys/class/infiniband/"
+
+// listRDMADevicesFromSysfs returns all RDMA device names from /sys/class/infiniband/.
+func listRDMADevicesFromSysfs(basePath string) []string {
```
gauravkghildiyal: Let's use [rdmamap.GetRdmaDeviceList()](https://github.com/Mellanox/rdmamap/blob/a32040148d75cba14f350fdcca9362a7634eba9b/rdma_map.go#L60) instead of defining ours? (We already are using rdmamap in several other places so it's not a matter of introducing a new dependency)

pkg/inventory/db.go
```diff
@@ -744,4 +745,54 @@ func isAllocatableNetworkDevice(dev *ghw.PCIDevice) bool {
...
+		if pciAttr, ok := device.Attributes[apis.AttrPCIAddress]; ok && pciAttr.StringValue != nil {
+			knownPCIAddresses.Insert(names.NormalizePCIAddress(*pciAttr.StringValue))
+		}
+	}
+
+	rdmaDevNames := listRDMADevicesFromSysfs(sysInfinibandPath)
+	for _, rdmaDevName := range rdmaDevNames {
+		pciAddr, err := pciAddressForRDMADevice(sysInfinibandPath, rdmaDevName)
+		if err != nil {
+			klog.V(4).Infof("Skipping RDMA device %s: %v", rdmaDevName, err)
```
gauravkghildiyal: Can we surface this as an Warning with verbosity 3 (the default)? I suppose this is something we'd never expect to happen.

pkg/inventory/db.go
gauravkghildiyal: Seems like there may still be some form of duplication with things like `AttrRDMADevice` being configured in multiple places. I think it's okay for now and would need to be part of a larger refactoring.

gauravkghildiyal (commented):
Thanks @Bowser1704. Just a few request but looking good.

pkg/inventory/sysfs.go
Bowser1704: fixed

pkg/inventory/db.go
Bowser1704: fixed

pkg/inventory/db.go
Bowser1704: Agreed — the current pipeline mixes discovery and attribute enrichment in the same functions, which leads to attributes like AttrPCIAddress and PCIe root being set in multiple places. 

Worth cleaning up in a future refactoring, perhaps alongside #99.

PR #247, opened 2026-07-03
Require CIDR for route Destination in validation

arpitjain099:
#### What type of PR is this?

/kind bug

A route Destination that is a bare IP (like `192.168.1.1` or `::1`) passes `validateRoutes` today but then silently fails to apply, because `applyRoutingConfig` parses the Destination with `net.ParseCIDR` while validation falls back to `net.ParseIP`. So the config is reported valid and then errors out at Prepare time.

Since `RouteConfig.Destination` is documented as CIDR format (and the examples all use CIDRs, including host routes as `/32` and `/128`), this tightens validation to require a CIDR so it matches what the apply path actually accepts. It also lines routes up with the rest of the package: `validateRules` already requires CIDR for `rule.Destination` with no bare-IP fallback.

Per review, the error message now points the user at the fix: an unparseable Destination reports `invalid CIDR format '...' (host routes use /32 or /128)`.

N/A

If you would rather support bare IPs as host routes instead, I am happy to flip it the other way and fix the apply path rather than validation, just let me know. I added table cases to `TestValidateRoutes` that fail before and pass after, and `GOOS=linux go test ./pkg/apis` passes.

```release-note
A network route `Destination` that is a bare IP address is now rejected during validation with a clear error. Use CIDR notation, including `/32` for a single IPv4 host or `/128` for a single IPv6 host.
```

### DIFF
--- a/pkg/apis/validation.go
+++ b/pkg/apis/validation.go
@@ -211,7 +211,7 @@ func validateRoutes(routes []RouteConfig, fieldPath string) (allErrors []error) 
 			allErrors = append(allErrors, fmt.Errorf("%s.destination: cannot be empty", currentFieldPath))
 		} else {
 			if _, _, err := net.ParseCIDR(route.Destination); err != nil {
-				allErrors = append(allErrors, fmt.Errorf("%s.destination: invalid CIDR format '%s'", currentFieldPath, route.Destination))
+				allErrors = append(allErrors, fmt.Errorf("%s.destination: invalid CIDR format '%s' (host routes use /32 or /128)", currentFieldPath, route.Destination))
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


thc1006:
Read this against the apply path and the rest of validation. The change is correct: `applyRoutingConfig` (pkg/driver/netnamespace.go) parses `route.Destination` with `net.ParseCIDR`, so a bare-IP destination passed validation under the old `ParseIP` fallback and then failed when the route was applied, which is the silent gap this closes.

It also lines routes up with the rest of the package: `validateRules` already requires CIDR for `rule.Destination` with no bare-IP fallback, and `RouteConfig.Destination` is documented as CIDR. So routes were the outlier, and this makes validation, the apply path, and the doc agree. Test coverage looks good too (valid /32, bare IPv4 and IPv6 both rejected).

One small optional thought: since a bare host IP is the likely way someone trips this, the error could nudge them toward the fix, e.g. `invalid CIDR format '192.168.1.1' (host routes use /32 or /128)`. Not important, the current message is fine.

Nothing blocking from me.

arpitjain099:
Good call, thanks @thc1006. I took the suggestion: when the Destination fails to parse the error now reads `invalid CIDR format '...' (host routes use /32 or /128)` so a bare host IP points the user straight at the fix. Logic is unchanged and the test assertion was updated to match. Appreciate the thorough review.

gauravkghildiyal:
Thanks @arpitjain099 and @thc1006 

@arpitjain099 Can you please update the PR description in the expected format?

/ok-to-test
/lgtm
/approve

PR #249, opened 2026-07-08
helm: allow global values in chart schema

JoZie:
#### What type of PR is this?
/kind bug

When dranet is used as a subchart, Helm automatically passes the parent chart's
`global` values into it. The chart schema had `additionalProperties: false` without
declaring `global`, causing schema validation to fail for any parent chart that
uses `global` values.

N/A

`global` is a Helm-reserved key automatically propagated to all subcharts. No
functional change — schema validation only.

```release-note
Fixed Helm chart schema validation failure when dranet is used as a subchart and
the parent chart defines `global` values.
```

### DIFF
--- a/deployments/helm/dranet/values.schema.json
+++ b/deployments/helm/dranet/values.schema.json
@@ -107,6 +107,9 @@
         }
       }
     },
+    "global": {
+      "type": "object"
+    },
     "serviceAccount": {
       "type": "object",
       "additionalProperties": false,


aojea:
@fmuyassarov you are the expert / maintainer here :)

fmuyassarov:
> @fmuyassarov you are the expert / maintainer here :)

Thanks for the ping @aojea . I will review it in a couple of hours.

aojea:
> Thanks for the ping @aojea . I will review it in a couple of hours.

no rush or obligation, is also fair to decline assignments

fmuyassarov (commented):
Thanks for the change @JoZie. It is indeed something we should allow for users
that are using DRANET as a subchart inside a parent umbrella chart.

/ok-to-test
/approve
/lgtm

fmuyassarov:
@aojea I would approve this patch (but I don't have rights for that). It is indeed something we should allow as I mentioned above. If it is good for you too, let's get it merged.

aojea:
/lgtm
/approve

@fmuyassarov @gauravkghildiyal we need to put an OWNERS file in https://github.com/JoZie/dranet/tree/main/deployments/helm/dranet  so we can start delegating subsystems like this

fmuyassarov:
> /lgtm /approve
> 
> @fmuyassarov @gauravkghildiyal we need to put an OWNERS file in https://github.com/JoZie/dranet/tree/main/deployments/helm/dranet so we can start delegating subsystems like this

Good idea. Filed it at https://github.com/kubernetes-sigs/dranet/pull/252

PR #251, opened 2026-07-08
feat: make kubelet registrar and plugins directories configurable

### LINKED ISSUE #250: dranet fails to register on clusters with a relocated kubelet --root-dir (hardcoded /var/lib/kubelet plugin dirs)
### What happened

On a node whose kubelet runs with a non-default `--root-dir` (e.g. `/mnt/fast/k8s/kubelet`), the `dranet` DaemonSet crash-loops and never registers:

```
draplugin.go: "Starting"
nonblockinggrpcserver.go: "GRPC server started" logger="registrar" endpoint="/var/lib/kubelet/plugins_registry/dra.net-reg.sock"
app.go:169] driver failed to start: context deadline exceeded
```

No `dra.net` ResourceSlice is published. (This is distinct from #19, which was a kubelet gRPC idle-connection deadlock fixed in kubernetes/kubernetes#133934; reproduced here fresh on `dranet:stable` + kubelet v1.36.1, which already carries that fix.)

### Root cause

The kubelet plugin-watcher only watches `<kubelet-root-dir>/plugins_registry/`. dranet creates its registrar socket under a hardcoded `/var/lib/kubelet/plugins_registry/` (the `kubeletplugin` default), and `install.yaml` mounts that path as a literal `hostPath`. With a relocated `--root-dir`, the socket lands in a directory the kubelet never watches -> registration times out (~30s) -> `context deadline exceeded`.

Confirmed on `main`:
- `install.yaml` hardcodes `/var/lib/kubelet/plugins` and `/var/lib/kubelet/plugins_registry`.
- `cmd/dranet` defines no kubelet-path flag (args are `--v` / `--hostname-override` only).

### Reproduction + fix (both verified)

- **Repro:** `kubectl apply -f install.yaml` on a cluster with `--root-dir=/mnt/fast/k8s/kubelet` -> CrashLoop; `dra.net-reg.sock` lands in `/var/lib/kubelet/plugins_registry` (which the kubelet does not watch), while the real `<root-dir>/plugins_registry` has the other drivers' sockets but not dranet's.
- **Fix (confirmed working):** repoint the `plugin-registry` hostPath at the real root-dir:
  ```
  kubectl patch ds dranet -n kube-system --type=strategic -p \
    '{"spec":{"template":{"spec":{"volumes":[{"name":"plugin-registry","hostPath":{"path":"/mnt/fast/k8s/kubelet/plugins_registry"}}]}}}}'
  ```
  After the patch: the pod is `Running`, a `dra.net` ResourceSlice is published, and the registrar socket is in the watched dir.

### Environment

- Kubernetes v1.36.1 (kubeadm), kubelet `--root-dir=/mnt/fast/k8s/kubelet`
- dranet `registry.k8s.io/networking/dranet:stable` (`main` `install.yaml`)
- The same class of issue exists in `kubernetes-sigs/dra-driver-cpu` (filed as dra-driver-cpu#231).

### Proposed fix

Make the kubelet registrar + plugins directories configurable, as `NVIDIA/k8s-dra-driver-gpu` does, via `--kubelet-registrar-directory-path` / `--kubelet-plugins-directory-path` (env `KUBELET_REGISTRAR_DIRECTORY_PATH` / `KUBELET_PLUGINS_DIRECTORY_PATH`, defaulting to the `kubeletplugin` constants so current behavior is unchanged), and let operators set them plus the matching hostPaths. Happy to open a PR.

thc1006:
#### What type of PR is this?

/kind feature

On a node whose kubelet runs with a non-default `--root-dir`, `dranet` crash-loops and never registers: the kubelet plugin-watcher only watches `<root-dir>/plugins_registry`, but the driver creates its registrar socket under the hardcoded `/var/lib/kubelet/plugins_registry` (the `kubeletplugin` default), so registration times out with `context deadline exceeded`.

This adds a single `--kubelet-root-dir` flag (default `/var/lib/kubelet`) and a matching `kubeletRootDir` Helm value. The driver derives its registration directory (`<root>/plugins_registry`) and plugin data directory (`<root>/plugins/<driver>`) from it and passes them to the `kubeletplugin` `RegistrarDirectoryPath` / `PluginDataDirectoryPath` options. The chart uses the same value for the flag and for the plugin `hostPath` and `mountPath`, so the path is identical inside and outside the container, which the `kubeletplugin` options require: "This path must be the same inside and outside of the driver's container." At the default, rendered output and behavior are unchanged.

Verified on a live cluster (kubelet `--root-dir=/mnt/fast/k8s/kubelet`, v1.36.1): without the flag the DaemonSet CrashLoops with `context deadline exceeded`; with `--kubelet-root-dir=/mnt/fast/k8s/kubelet` the driver registers within a few seconds and publishes its `dra.net` ResourceSlice, with the registration socket under `<root>/plugins_registry` and `dra.sock` under `<root>/plugins/dra.net`. `go build ./...`, `CGO_ENABLED=1 go test -race` on the touched packages, `golangci-lint run` (v2.9.0), and `helm lint --strict` all pass; `helm template` at the default renders the current `/var/lib/kubelet` paths unchanged.

#### Which issue(s) this PR fixes

Fixes #250

#### Special notes for your reviewer

Per your suggestion this is a single `--kubelet-root-dir` flag rather than two separate directory flags, matching the Helm value and the kubelet's own `--root-dir`. The static `install.yaml` variants still target the default layout; I am happy to regenerate them to match if you would like.

```release-note
dranet: add a `--kubelet-root-dir` flag and a `kubeletRootDir` Helm value so the driver can run on clusters where the kubelet uses a non-default `--root-dir`.
```

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -46,10 +46,6 @@ import (
 )
 
 const (
-	kubeletPluginPath = "/var/lib/kubelet/plugins"
-)
-
-const (
 	// maxAttempts indicates the number of times the driver will try to recover itself before failing
 	maxAttempts = 5
 )
@@ -96,6 +92,16 @@ func WithDBPath(path string) Option {
 	}
 }
 
+// WithKubeletRootDir sets the kubelet data directory (its --root-dir). The
+// driver's registration socket lives under <dir>/plugins_registry and its
+// dra.sock under <dir>/plugins. Set this when the kubelet runs with a
+// non-default --root-dir.
+func WithKubeletRootDir(dir string) Option {
+	return func(o *NetworkDriver) {
+		o.kubeletRootDir = dir
+	}
+}
+
 type NetworkDriver struct {
 	draPlugin     pluginHelper
 	driverName    string
@@ -112,6 +118,10 @@ type NetworkDriver struct {
 	rdmaSharedMode bool
 	podConfigStore *PodConfigStore
 	dbPath         string // path for persistent bbolt database; empty means in-memory
+
+	// kubeletRootDir is the kubelet data directory (its --root-dir). Set when the
+	// kubelet runs with a non-default --root-dir.
+	kubeletRootDir string
 
 	clock clock.WithTicker // Injectable clock for testing
 }
@@ -165,16 +175,22 @@ func Start(ctx context.Context, driverName string, kubeClient kubernetes.Interfa
 	}
 	plugin.podConfigStore = store
 
-	driverPluginPath := filepath.Join(kubeletPluginPath, driverName)
+	driverPluginPath := filepath.Join(plugin.kubeletRootDir, "plugins", driverName)
 	err = os.MkdirAll(driverPluginPath, 0750)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create plugin path %s: %v", driverPluginPath, err)
 	}
 
+	// Derive the registration and plugin data directories from the kubelet root
+	// dir so they are correct when the kubelet uses a non-default --root-dir. At
+	// the default this matches the kubeletplugin defaults, so existing deployments
+	// are unaffected.
 	kubeletOpts := []kubeletplugin.Option{
 		kubeletplugin.DriverName(driverName),
 		kubeletplugin.NodeName(nodeName),
 		kubeletplugin.KubeClient(kubeClient),
+		kubeletplugin.RegistrarDirectoryPath(filepath.Join(plugin.kubeletRootDir, "plugins_registry")),
+		kubeletplugin.PluginDataDirectoryPath(driverPluginPath),
 	}
 	d, err := kubeletplugin.Start(ctx, plugin, kubeletOpts...)
 	if err != nil {
--- a/cmd/dranet/app.go
+++ b/cmd/dranet/app.go
@@ -67,6 +67,8 @@ var (
 	profileProvider   string
 	webhookURL        string
 
+	kubeletRootDir string
+
 	ready atomic.Bool
 )
 
@@ -83,6 +85,7 @@ func init() {
 	flag.StringVar(&cloudProviderHint, "cloud-provider-hint", "", "Hint for the cloud provider that will be used to select the appropriate provider plugin. Supported values: (AWS, GCE, AZURE, OKE, ALIBABA, webhook, NONE). If left unset, the cloud provider is auto-detected.")
 	flag.StringVar(&profileProvider, "profile-provider", "cloud", "Provides user intent (cloud, webhook, none). 'cloud' falls back to the cloud-provider's native implementation.")
 	flag.StringVar(&webhookURL, "webhook-url", "", "URL for the webhook provider (required if using webhook for either provider)")
+	flag.StringVar(&kubeletRootDir, "kubelet-root-dir", "/var/lib/kubelet", "The kubelet data directory (its --root-dir). The driver's registration socket lives under <dir>/plugins_registry and its dra.sock under <dir>/plugins/<driver-name>. Set this to match the kubelet --root-dir on clusters that relocate it.")
 
 	flag.Usage = func() {
 		fmt.Fprint(os.Stderr, "Usage: dranet [options]\n\n")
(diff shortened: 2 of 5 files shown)


thc1006:
The `test (1.26.x)` and `Bats e2e tests` failures are both transient download flakes, not from this change:

- `test (1.26.x)`: `go: download go1.26.0: ... net/http: TLS handshake timeout` while fetching the toolchain, before any test ran.
- `Bats e2e tests`: `setup_file` failed building the webhook-whereabouts image on `go mod tidy` with `golang.org/x/sys@v0.46.0: ... stream error: INTERNAL_ERROR received from peer`.

`build`, `e2e (ipv4)`, `e2e (ipv6)`, and `helm-lint` passed. Locally `go build ./...`, `CGO_ENABLED=1 go test -race` on the touched packages, `golangci-lint run` (v2.9.0), and `helm lint --strict` all pass. A `/retest` should clear both.

pkg/driver/driver_test.go
```diff
@@ -200,3 +200,22 @@ func TestStop(t *testing.T) {
...
 }
+
+func TestPluginDataDir(t *testing.T) {
+	const driverName = "dra.net"
+	tests := []struct {
+		name       string
+		configured string
+		want       string
+	}{
+		{"empty uses the kubeletplugin default base", "", "/var/lib/kubelet/plugins/dra.net"},
```
gauravkghildiyal: nit: I suppose my other comment might make some of this obsolete, but please consider naming the struct fields for better readability.

gauravkghildiyal (commented):
Thanks for the great details @thc1006 !

I think it makes sense to keep things simple for the user (and us) and only introduce a `--kubelet-root-dir` flag, similar to the helm chart. 

/ok-to-test

thc1006:
Done, thanks @gauravkghildiyal. Replaced the two flags with a single `--kubelet-root-dir` (default `/var/lib/kubelet`); the driver derives `<root>/plugins_registry` and `<root>/plugins/<driver>` from it, and the chart passes the same `kubeletRootDir` value to the flag and to both the `hostPath` and `mountPath`, so the path is identical inside and outside the container. Also named the struct fields in the test as you suggested.

Verified the single flag end-to-end on the relocated-root-dir cluster: it registers and publishes its `dra.net` slice, with the registration socket under `<root>/plugins_registry` and `dra.sock` under `<root>/plugins/dra.net`. `golangci-lint` (v2.9.0), `go test -race`, and `helm lint --strict` pass.

pkg/driver/driver_test.go
thc1006: Done, named the struct fields (`name` / `kubeletRootDir` / `want`). The switch to `--kubelet-root-dir` also simplified `pluginDataDir`, which this test covers.

pkg/driver/driver.go
```diff
@@ -46,7 +46,8 @@ import (
 )
 
 const (
-	kubeletPluginPath = "/var/lib/kubelet/plugins"
+	// kubeletRootDirDefault is the kubelet's default data directory (--root-dir).
+	kubeletRootDirDefault = "/var/lib/kubelet"
```
gauravkghildiyal: Should we get rid of this const, given the flag already has a default value?

thc1006: Done, removed it. The `--kubelet-root-dir` flag default (`/var/lib/kubelet`) is the single source now; `Start` is only called from `cmd/dranet`, where the flag always provides a value.

gauravkghildiyal: Thanks! Sorry to ask but just this one last thing, can you also cleanup the comments which indicate that "if empty, /var/lib/kubelet is used". Also let's get rid of the `pluginDataDir()`, since at this point it is just a wrapper over filepath.Join and not providing much value?

thc1006: Done, thanks. Cleaned up those comments and inlined the `filepath.Join`, so `pluginDataDir` and its test are gone now.

gauravkghildiyal:
Thanks a lot @thc1006 

/lgtm
/approve

gauravkghildiyal:
(Rerun manually, /retest doesn't work)

thc1006:
Heads up: the `Bats e2e tests` failure is a flake in the `whereabouts-webhook` test setup (`kubectl wait ... -l app=whereabouts-webhook` -> `no matching resources found`), not from this change, which only touches the dranet driver and its chart. That job passed on the previous two pushes of this PR and on `main`. `pull-dranet-nftables` (+ ipv6), `test (1.26.x)`, `build`, `e2e`, and `helm-lint` are all green here. Re-running the Bats job should clear it.

PR #252, opened 2026-07-08
chore: add OWNERS for helm chart and OWNERS_ALIASES

fmuyassarov:
#### What type of PR is this?
/kind cleanup

Add OWNERS file for deployments/helm/dranet/ so PRs touching only the Helm chart can be approved by its owners independently of the root OWNERS file.

N/A

```release-note
NONE
```

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,0 +1,8 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+aliases:
+  # -----------------------------------------------------------
+  # OWNERS_ALIASES for deployments/helm/dranet
+  # -----------------------------------------------------------
+  dranet-helm-approvers:
+    - fmuyassarov
--- a/deployments/helm/dranet/OWNERS
+++ b/deployments/helm/dranet/OWNERS
@@ -1,0 +1,5 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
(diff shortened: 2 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -0,0 +1,11 @@
...
+
+aliases:
+  # -----------------------------------------------------------
+  # OWNERS_ALIASES for deployments/helm/dranet
+  # -----------------------------------------------------------
+  dranet-helm-approvers:
+    - aojea
+    - gauravkghildiyal
+    - michaelasp
+    - fmuyassarov
```
fmuyassarov, line 8: this is purely copy paste from https://github.com/kubernetes-sigs/dranet/blob/main/deployments/helm/dranet/Chart.yaml. Let me know if I need to update the list.

aojea: I think we already inherit those permissions, so it should be just you IIRC

fmuyassarov: fixed.

fmuyassarov:
interesting why netlify got stuck. 
@aojea do you happen to know if there is a way to re-trigger it or shall I force push rather?

gauravkghildiyal:
Netlify Github plugins seem to have had some problem recently.

fmuyassarov:
thanks @gauravkghildiyal

PR #254, opened 2026-07-09
apis: reject routes and rules that mix IPv4 and IPv6 in validation

### LINKED ISSUE #253: Route/rule validation accepts mixed IPv4/IPv6 configs that the kernel rejects at apply time
### What happened

A `NetworkConfig` whose route mixes IP families (an IPv4 destination with an IPv6 gateway, or vice versa), or whose rule has a source and destination in different families, passes `ValidateConfig` but then fails when the driver applies it, because the kernel rejects the mismatch. The config is reported valid and then fails at Prepare time.

### Root cause

`validateRoutes` checks the gateway and source with `net.ParseIP`, which accepts any valid IP regardless of the destination's family, and `validateRules` checks source and destination independently with `net.ParseCIDR`. Neither checks that the families are consistent. The apply path uses `vishvananda/netlink`, whose `RouteAdd` / `RuleAdd` reject a family mismatch unconditionally before the syscall (`route_linux.go`, `rule_linux.go`: "gateway, source, and destination ip are not the same IP family" / "source and destination ip are not the same IP family").

### Reproduction

Applying the following against a real interface in a netns (dummy `eth0` with both a v4 and a v6 address) shows validation passing and apply failing:

| config | ValidateConfig | apply (netlink) |
|---|---|---|
| route `dst 10.1.0.0/24`, `gw 2001:db8::1` | accepted | `gateway, source, and destination ip are not the same IP family` |
| route `dst 2001:db8:2::/64`, `gw 10.0.0.1` | accepted | same |
| rule `src 10.0.0.0/8`, `dst 2001:db8::/32` | accepted | `source and destination ip are not the same IP family` |

### Impact

Dual-stack setups are where this bites: a mistyped or copy-pasted gateway in the wrong family is accepted by the API, and the failure only surfaces later as a Prepare error, with no hint that the family mismatch is the cause.

### Proposed fix

Check IP-family consistency in `validateRoutes` (gateway and source vs destination) and `validateRules` (source vs destination), so validation rejects what the kernel would reject. This is the same "make validation match the apply path" direction as #247, which tightened route destinations. I have a change with unit tests ready and am happy to open a PR.

thc1006:
#### What type of PR is this?

/kind bug

`validateRoutes` checks the gateway and source with `net.ParseIP` (which accepts any family regardless of the destination), and `validateRules` checks source and destination independently. Neither checks that the IP families are consistent, but the apply path (netlink `RouteAdd` / `RuleAdd`) rejects a family mismatch unconditionally. So a route with an IPv4 destination and an IPv6 gateway, or a rule whose source and destination are different families, passes validation and then fails at Prepare time with no hint that the family is the cause.

This checks IP-family consistency in `validateRoutes` (gateway and source vs destination) and `validateRules` (source vs destination), so validation rejects what the kernel would reject. It follows the same "make validation match the apply path" direction as #247, which tightened route destinations.

Verified with a differential harness (apply each config against a real interface in a netns): the three mismatch cases below passed `ValidateConfig` and failed netlink before this change, and are rejected by validation after it. Unit tests cover both the rejected mismatches and the valid same-family cases.

#### Which issue(s) this PR is related to

Fixes #253

#### Special notes for reviewers

`sameIPFamily` uses `To4()`, matching how `netlink` determines family (an IPv4-mapped IPv6 address is treated as IPv4 by both), so validation and apply stay in agreement. The route destination stays CIDR-only per #247; the change only retains the parsed destination IP to compare families.

#### Release note

```release-note
Route and rule network configs that mix IPv4 and IPv6 (e.g. an IPv4 destination with an IPv6 gateway) are now rejected during validation instead of failing later when the configuration is applied.
```

### DIFF
--- a/pkg/apis/validation_test.go
+++ b/pkg/apis/validation_test.go
@@ -433,6 +433,39 @@ func TestValidateRoutes(t *testing.T) {
 			expectErr: true,
 			errCount:  1,
 		},
+		{
+			name:      "valid IPv6 route with IPv6 gateway",
+			routes:    []RouteConfig{{Destination: "2001:db8:1::/64", Gateway: "2001:db8::1", Scope: scopeUniverse}},
+			fieldPath: "routes",
+			expectErr: false,
+		},
+		{
+			name:      "valid route with matching gateway and source",
+			routes:    []RouteConfig{{Destination: "10.0.0.0/8", Gateway: "192.168.1.1", Source: "192.168.1.2", Scope: scopeUniverse}},
+			fieldPath: "routes",
+			expectErr: false,
+		},
+		{
+			name:      "gateway family mismatch (IPv4 destination, IPv6 gateway)",
+			routes:    []RouteConfig{{Destination: "10.0.0.0/8", Gateway: "2001:db8::1", Scope: scopeUniverse}},
+			fieldPath: "routes",
+			expectErr: true,
+			errCount:  1,
+		},
+		{
+			name:      "gateway family mismatch (IPv6 destination, IPv4 gateway)",
+			routes:    []RouteConfig{{Destination: "2001:db8::/64", Gateway: "192.168.1.1", Scope: scopeUniverse}},
+			fieldPath: "routes",
+			expectErr: true,
+			errCount:  1,
+		},
+		{
+			name:      "source family mismatch (IPv4 destination, IPv6 source)",
+			routes:    []RouteConfig{{Destination: "10.0.0.0/8", Gateway: "192.168.1.1", Source: "2001:db8::2"}},
+			fieldPath: "routes",
+			expectErr: true,
+			errCount:  1,
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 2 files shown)


pkg/apis/validation_test.go
```diff
@@ -433,6 +433,39 @@ func TestValidateRoutes(t *testing.T) {
...
 			errCount:  1,
 		},
+		{
+			name:      "valid IPv6 route with IPv6 gateway",
+			routes:    []RouteConfig{{Destination: "2001:db8:1::/64", Gateway: "2001:db8::1", Scope: scopeUniverse}},
+			fieldPath: "routes",
+			expectErr: false,
+		},
+		{
+			name:      "valid route with matching gateway and source",
```
gauravkghildiyal: nit: In the context of this PR, 'matching' makes perfect sense, but so it makes sense later on, can we please explicitly clarify we're looking for matching ipfamily. 

(suggestion for any other places where that may be missing)

gauravkghildiyal (commented):
Looks great @thc1006, thanks for improving.

Left one minor request and it's ready.

thc1006:
Thanks for the review, @gauravkghildiyal. Clarifying the names by IP family does read better outside this PR's context. I applied it to both flagged cases and the sibling IPv6 route case so they stay consistent.

PR #256, opened 2026-07-09
apis: deduplicate merged routes by destination and table

### LINKED ISSUE #255: MergeNetworkConfig drops routes that share a destination but use different tables
### What happened

When a cloud provider (or webhook) config is present, `MergeNetworkConfig` deduplicates the merged routes by destination only. A config that defines the same destination in two routing tables (policy routing) loses one of the routes after the merge.

### Root cause

`deduplicateRoutes` keys the `seen` set on `route.Destination` alone, ignoring `Table`. Two routes with the same destination in different tables are distinct entries, but the dedup treats them as duplicates and keeps only the last one.

### Reproduction

User config with the same destination in two tables, merged with a cloud config that has a default route:

| input | routes |
|---|---|
| user | `10.0.0.0/8 table 100`, `10.0.0.0/8 table 200` |
| cloud | `0.0.0.0/0` |
| merged (actual) | `0.0.0.0/0`, `10.0.0.0/8 table 200` |
| merged (expected) | `0.0.0.0/0`, `10.0.0.0/8 table 100`, `10.0.0.0/8 table 200` |

The `table 100` route is silently dropped.

### Impact

Silent route loss for policy-routing / multi-table setups combined with a cloud provider config. The merge reports success and the missing route only surfaces later as a routing misconfiguration. This path is only taken when a cloud or webhook config is merged; the user-only path returns early without deduplication.

### Proposed fix

Key the deduplication on destination and table, so distinct-table routes are all kept while a route that collides on both destination and table still resolves in favor of the user config. I have a change with unit tests ready and am happy to open a PR.

thc1006:
#### What type of PR is this?

/kind bug

`MergeNetworkConfig` deduplicated the merged routes by destination only, so a config with the same destination in two routing tables (policy routing) lost one route after being merged with a cloud provider config. `deduplicateRoutes` keyed the `seen` set on `route.Destination` alone, ignoring `Table`.

This keys the deduplication on destination and table, so routes to the same destination in different tables are all kept, while a route that collides on both destination and table still resolves in favor of the user config (the existing "user wins" behavior).

Only the cloud/webhook merge path is affected; the user-only path returns early without deduplication.

#### Which issue(s) this PR is related to

Fixes #255

#### Special notes for reviewers

Neighbors are left unchanged: a neighbor maps an IP to a single MAC, so deduplicating by destination is correct there. Tests cover both the multi-table case (all kept) and the same-destination-and-table case (user still wins).

#### Release note

```release-note
Merging a network config with a cloud provider config no longer drops routes that share a destination but use different routing tables.
```

### DIFF
--- a/pkg/apis/config_merge_test.go
+++ b/pkg/apis/config_merge_test.go
@@ -131,7 +131,7 @@ func TestMergeNetworkConfig(t *testing.T) {
 			},
 		},
 		{
-			name: "same destination in different tables is kept (policy routing)",
+			name: "same destination in different tables is kept across user and cloud (policy routing)",
 			user: &NetworkConfig{
 				Routes: []RouteConfig{
 					{Destination: "10.0.0.0/8", Gateway: "192.168.1.1", Table: 100},
@@ -141,11 +141,15 @@ func TestMergeNetworkConfig(t *testing.T) {
 			cloud: &NetworkConfig{
 				Routes: []RouteConfig{
 					{Destination: "0.0.0.0/0", Gateway: "192.168.1.254"},
+					// Same destination as the user routes but a distinct table:
+					// this cloud route must be preserved, not deduplicated away.
+					{Destination: "10.0.0.0/8", Gateway: "192.168.1.254", Table: 300},
 				},
 			},
 			want: &NetworkConfig{
 				Routes: []RouteConfig{
 					{Destination: "0.0.0.0/0", Gateway: "192.168.1.254"},
+					{Destination: "10.0.0.0/8", Gateway: "192.168.1.254", Table: 300},
 					{Destination: "10.0.0.0/8", Gateway: "192.168.1.1", Table: 100},
 					{Destination: "10.0.0.0/8", Gateway: "192.168.1.1", Table: 200},
 				},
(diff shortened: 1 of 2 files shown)


pkg/apis/config_merge_test.go
```diff
@@ -130,6 +130,45 @@ func TestMergeNetworkConfig(t *testing.T) {
...
+			name: "same destination in different tables is kept (policy routing)",
+			user: &NetworkConfig{
+				Routes: []RouteConfig{
+					{Destination: "10.0.0.0/8", Gateway: "192.168.1.1", Table: 100},
+					{Destination: "10.0.0.0/8", Gateway: "192.168.1.1", Table: 200},
+				},
+			},
+			cloud: &NetworkConfig{
+				Routes: []RouteConfig{
+					{Destination: "0.0.0.0/0", Gateway: "192.168.1.254"},
```
gauravkghildiyal, line 143: I wonder if we should change the Destination=10.0.0.0/8 and table=300 so this also verifies merging across user and cloud also obeys the correct deduplication strategy and that this route would be preserved?

gauravkghildiyal (commented):
Looks mostly ready. Thanks @thc1006

thc1006:
Thanks for the review and the suggestion, @gauravkghildiyal. The cross-source case was a good one to add: the test now keeps a cloud `10.0.0.0/8` route on table 300 next to the user's tables, and the same-`(destination, table)` collision resolving to the user stays covered by the adjacent case.

gauravkghildiyal:
Thanks. Seems like my approval had a typo

/approve

PR #258, opened 2026-07-09
inventory: separate discovery from attribute enrichment

Bowser1704:
#### What this PR does / why we need it:
/kind feature

Separate device discovery from attribute enrichment.

#### Which issue(s) this PR is related to:

Fixes #246 

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #260, opened 2026-07-16
feat: publish rdmaDevice for netdev-linked RoCE/IB VFs

cabrinha:
#### What type of PR is this?

/kind feature

RDMA-capable SR-IOV VFs with an active netdev (e.g. Broadcom `bnxt_re*`
backing RoCE VFs) were advertised on the ResourceSlice with `dra.net/rdma=true`
but **without** `dra.net/rdmaDevice` — only the IB-only / netdev-absent PCI
branch set that attribute. As a result the kernel RDMA link name was not
discoverable from the inventory for the most common RoCE case.

This PR publishes `dra.net/rdmaDevice` (the kernel RDMA link name, e.g.
`bnxt_re12`, `mlx5_0`) for netdev-linked RoCE/IB devices too.

Per review feedback, the resolution stays in `addRDMAAttributes`: its `netdev`
branch trusts an existing `rdmaDevice` value (same pattern as the PCI branch),
otherwise resolves the kernel RDMA link name via `GetRdmaDevice(ifName)` and
fills the attribute. The `rdma=true/false` bool is derived from a successful
resolution, which consolidates RDMA detection into a single path and removes
the now-unused `isRdmaDeviceInSysfs` helper (its sysfs fallback already lives
in `GetRdmaDevice`).

Adds unit coverage for the netdev-linked attribute path (preserve an existing
value, and the non-RDMA case). Sysfs resolution itself is already covered by
`TestGetRdmaDeviceFromSysfs`.

> **Note:** an earlier revision of this PR also added an opt-in
> `--inject-rdma-env` flag that injected `UCX_NET_DEVICES` into containers. That
> has been dropped per review — env injection is application-specific (NCCL uses
> `NCCL_IB_HCA_NAME`, etc.), doesn't scale to every RDMA library, and can be
> silently overridden at runtime. Publishing `rdmaDevice` on the ResourceSlice
> lets applications set whatever env their stack needs.

N/A

```release-note
Broadcom/Mellanox RoCE SR-IOV VFs with an active netdev now publish the `dra.net/rdmaDevice` attribute (the kernel RDMA link name, e.g. `bnxt_re12`) on their ResourceSlice, previously set only for IB-only devices.
```

### DIFF
--- a/pkg/inventory/db.go
+++ b/pkg/inventory/db.go
@@ -483,15 +483,13 @@ func (db *DB) addRDMAAttributes(devices []resourceapi.Device) []resourceapi.Devi
 	for i := range devices {
 		isRDMA := false
 		if ifName := devices[i].Attributes[apis.AttrInterfaceName].StringValue; ifName != nil && *ifName != "" {
-			// Try rdmamap library first
-			isRDMA = rdmamap.IsRDmaDeviceForNetdevice(*ifName)
-
-			// Fallback to sysfs check if rdmamap fails. This is particularly
-			// needed for InfiniBand interfaces where rdmamap has a bug comparing
-			// against node GUID instead of port GUID:
-			// https://github.com/Mellanox/rdmamap/issues/15
-			if !isRDMA {
-				isRDMA = isRdmaDeviceInSysfs(*ifName)
+			// Netdev-linked device (RoCE or IPoIB): if AttrRDMADevice was already
+			// set, trust it; otherwise resolve and publish the kernel RDMA link name.
+			if rdmaDevAttr, ok := devices[i].Attributes[apis.AttrRDMADevice]; ok && rdmaDevAttr.StringValue != nil && *rdmaDevAttr.StringValue != "" {
+				isRDMA = true
+			} else if rdmaDevName, err := GetRdmaDevice(*ifName); err == nil && rdmaDevName != "" {
+				isRDMA = true
+				devices[i].Attributes[apis.AttrRDMADevice] = resourceapi.DeviceAttribute{StringValue: ptr.To(rdmaDevName)}
 			}
 		} else if pciAddr := devices[i].Attributes[apis.AttrPCIAddress].StringValue; pciAddr != nil && *pciAddr != "" {
 			// IB-only device: has RDMA capability but no netdev interface.
--- a/pkg/inventory/sysfs.go
+++ b/pkg/inventory/sysfs.go
@@ -185,27 +185,6 @@ func getRdmaDeviceFromSysfs(basePath, ifName string) (string, error) {
 	return "", fmt.Errorf("no RDMA device found for %s", ifName)
 }
 
-// isRdmaDeviceInSysfs checks if a network interface has RDMA capability by
-// examining the sysfs infiniband directory. This serves as a workaround for
-// cases where the rdmamap library fails to detect RDMA devices, particularly
-// for InfiniBand interfaces where the library incorrectly compares against the
-// node GUID instead of the port GUID.
-//
-// The function checks /sys/class/net/{ifname}/device/infiniband/ for any RDMA
-// device entries. If the directory exists and contains at least one entry, the
-// interface is considered RDMA-capable.
-func isRdmaDeviceInSysfs(ifName string) bool {
-	// Check if the infiniband directory exists under the device
-	rdmaName, err := getRdmaDeviceFromSysfs(sysnetPath, ifName)
-	if err != nil {
-		klog.V(4).Infof("No RDMA device found for interface %s via sysfs: %v", ifName, err)
-		return false
-	}
-
-	klog.V(4).Infof("Interface %s is RDMA-capable with device %s", ifName, rdmaName)
-	return true
-}
-
 // pciAddress BDF Notation
 // [domain:]bus:device.function
 // https://wiki.xenproject.org/wiki/Bus:Device.Function_(BDF)_Notation
--- a/pkg/inventory/db_test.go
+++ b/pkg/inventory/db_test.go
@@ -676,4 +676,51 @@ func TestAddRDMAAttributesStandalone(t *testing.T) {
 			t.Errorf("AttrRDMADevice = %v, want erdma_0 (should not be overwritten)", v)
 		}
 	})
+
+	t.Run("netdev-linked device uses existing AttrRDMADevice without lookup", func(t *testing.T) {
+		// A value already resolved elsewhere must be preserved: add*Attributes
+		// only fills missing attributes.
+		ifName := "dranettest1"
+		devices := []resourceapi.Device{
+			{
+				Name: ifName,
+				Attributes: map[resourceapi.QualifiedName]resourceapi.DeviceAttribute{
+					apis.AttrInterfaceName: {StringValue: ptr.To(ifName)},
+					apis.AttrRDMADevice:    {StringValue: ptr.To("bnxt_re12")},
+				},
+			},
+		}
+		db := &DB{}
+		result := db.addRDMAAttributes(devices)
+
+		if v, ok := result[0].Attributes[apis.AttrRDMA]; !ok || v.BoolValue == nil || !*v.BoolValue {
+			t.Errorf("AttrRDMA = %v, want true", v)
+		}
+		if v, ok := result[0].Attributes[apis.AttrRDMADevice]; !ok || v.StringValue == nil || *v.StringValue != "bnxt_re12" {
+			t.Errorf("AttrRDMADevice = %v, want bnxt_re12 (should not be overwritten)", v)
+		}
+	})
+
+	t.Run("netdev without RDMA device is rdma=false", func(t *testing.T) {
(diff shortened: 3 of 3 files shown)


Bowser1704:
@cabrinha Thanks for the contribution. Two notes:

**Part 1 — `rdmaDevice` publish fix:**

We have an in-flight inventory refactor (#238) that separates discovery (identity attributes) from enrichment (derived attributes). `rdmaDevice` is an identity attribute and should be resolved during discovery, not inside `addRDMAAttributes`. Please take a look at that branch so we can land this fix consistently.

**Part 2 — `--inject-rdma-env`:**

1. **NCCL is not covered, and can't scale.** The feature only injects `UCX_NET_DEVICES` for UCX-based transports (vLLM/NIXL). NCCL, the most common RDMA library for GPU training, uses `NCCL_IB_HCA_NAME`. DRANET can't reasonably maintain env vars for every RDMA communication library — this is application-specific, not a driver's job.

2. **Env injection can be silently overridden.** Entry points, init scripts, or user code inside the container can override these vars at runtime, so the injected value can't be trusted.

cabrinha:
Part 1 (rdmaDevice): Agreed it's an identity attribute. Moved resolution into the discovery phase (addLinkAttributes → GetRdmaDevice) to match #258, so netdev-linked RoCE/IB VFs publish rdmaDevice alongside the IB-only case. addRDMAAttributes now just derives the rdma bool from the resolved name, which let me drop the now-redundant isRdmaDeviceInSysfs helper. I've stacked this on #258 — the branch currently includes that commit and will be rebased onto main once #258 lands; please review the second commit only. 

Part 2 (--inject-rdma-env): Fair points, I've dropped the flag and env injection entirely. Publishing rdmaDevice on the ResourceSlice covers the underlying need — I'll figure out how to best get this stuff into the pod at runtime and leave it open for others to do the same.

aojea:
/assign @Bowser1704 

it seems there is overlap with https://github.com/kubernetes-sigs/dranet/pull/258

cabrinha:
> /assign @Bowser1704
> 
> it seems there is overlap with #258

This needs to be merged after #258

gauravkghildiyal:
Hey @cabrinha, let's rebase now

Bowser1704:
Thanks @cabrinha. Following the existing discovery/enrichment design, we should resolve it in the `netdev` branch of addRDMAAttributes using `GetRdmaDevice(ifName)` and fill the attribute only when missing. 

Keeping the change in addRDMAAttributes would be smaller and consistent with the responsibility of `add*Attributes` to fill missing attributes without overwriting existing values.

cabrinha:
### Bats e2e failures look like a pre-existing flake in `tests/e2e.bats`, not this change

The two failing Bats runs failed on **different** tests each time, both with the identical error:

```
error: no matching resources found
```

emitted by `kubectl wait --timeout=30s --for=condition=ready pods -l <label>` immediately after `kubectl apply` of a Deployment:

- Run 1 → test 3 *"dummy interface with IP addresses ResourceClaimTemplate"* (`tests/e2e.bats:149`, `-l app=MyApp`)
- Run 2 → test 12 *"reapply pod with dummy resource claim"* (`tests/e2e.bats:367`, `-l app=reapplyApp`)

`kubectl wait` does not wait for objects to be *created* — if the ReplicaSet hasn't produced the Pod yet, it exits non-zero immediately with `no matching resources found` (as opposed to a `condition ready` timeout, which is what a real scheduling/device failure would produce). This `apply`-then-immediately-`wait -l` pattern appears 18 times in the suite, so under CI load it can trip on any test.

Evidence that it's flaky and not caused by this PR:

| commit | vs. previous | Bats |
|---|---|---|
| earlier revision | current RDMA logic + a comment-only edit to `pkg/apis/attributes.go` | ✅ pass |
| current head | **same runtime code**, only the `attributes.go` comments reverted | ❌ fail |
| re-trigger | byte-identical no-op amend of the above | ❌ fail |

Runtime-identical code passed once and failed twice, on different tests — and `main` is consistently green with the same racy harness. This PR only touches `pkg/inventory` and cannot affect whether a Pod object gets created.

Could a maintainer re-run the Bats workflow? Longer term the flake would go away by replacing

```sh
kubectl wait --timeout=30s --for=condition=ready pods -l <label>
```

with a create-safe wait, e.g. `kubectl rollout status --timeout=60s deployment/<name>` (or prepending `kubectl wait --for=create ...`). Happy to send that as a separate PR if it's wanted.

gauravkghildiyal:
@cabrinha Thanks for investigating the test flake. I retriggered the test. If you find the opportunity, it'd be great if you can open up an issue (bonus if you can send the fix too :) )

pkg/inventory/sysfs.go
```diff
@@ -29,9 +29,12 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// sysnetPath is the sysfs class-net directory.
+// https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-net
+// It is a var rather than a const so tests can point it at a fake sysfs tree.
+var sysnetPath = "/sys/class/net/"
```
Bowser1704: These changes seem unrelated

aojea: it got me too, but check the comment, is mutated during the test in `pkg/inventory/db_test.go`

cabrinha: Let me know if this needs to be removed and the test re-written.

pkg/inventory/db.go
```diff
@@ -483,15 +483,17 @@ func (db *DB) addRDMAAttributes(devices []resourceapi.Device) []resourceapi.Devi
 	for i := range devices {
 		isRDMA := false
 		if ifName := devices[i].Attributes[apis.AttrInterfaceName].StringValue; ifName != nil && *ifName != "" {
-			// Try rdmamap library first
-			isRDMA = rdmamap.IsRDmaDeviceForNetdevice(*ifName)
```
aojea, line 487: this is

```
// Returns true if rdma device exist for netdevice, else false
func IsRDmaDeviceForNetdevice(netdevName string) bool {
	rdma, _ := GetRdmaDeviceForNetdevice(netdevName)

	return (rdma != "")
}
```

pkg/inventory/db.go
```diff
@@ -483,15 +483,17 @@ func (db *DB) addRDMAAttributes(devices []resourceapi.Device) []resourceapi.Devi
...
-			// needed for InfiniBand interfaces where rdmamap has a bug comparing
-			// against node GUID instead of port GUID:
-			// https://github.com/Mellanox/rdmamap/issues/15
-			if !isRDMA {
-				isRDMA = isRdmaDeviceInSysfs(*ifName)
+			// Netdev-linked device (RoCE or IPoIB): resolve the kernel RDMA
+			// link name. GetRdmaDevice applies the rdmamap lookup with a sysfs
+			// fallback for InfiniBand NICs where rdmamap compares the node GUID
+			// instead of the port GUID (https://github.com/Mellanox/rdmamap/issues/15).
+			if rdmaDevName, err := GetRdmaDevice(*ifName); err == nil && rdmaDevName != "" {
```
aojea: this is 

```
func GetRdmaDevice(ifName string) (string, error) {
	if rdmaDev, _ := rdmamap.GetRdmaDeviceForNetdevice(ifName); rdmaDev != "" {
		return rdmaDev, nil
	}

	// Fallback to sysfs check if rdmamap fails. This is particularly related to a known
	// issue to detect RDMA devices for certain Mellanox NICs
	// https://github.com/Mellanox/rdmamap/issues/15

	rdmaDev, err := getRdmaDeviceFromSysfs(sysnetPath, ifName)
	if err != nil {
		return "", fmt.Errorf("no RDMA device found for %s: %w", ifName, err)
	}

	return rdmaDev, nil
}
```

pkg/inventory/sysfs.go
cabrinha: Removed — `sysnetPath` is a `const` again. The netdev tests no longer mutate it; they cover the trust-existing-`AttrRDMADevice` path and the non-RDMA case. Sysfs resolution remains covered by `TestGetRdmaDeviceFromSysfs`.

pkg/inventory/db.go
cabrinha: Exactly — that bool helper already had the name and threw it away. Updated the netdev branch to mirror the PCI path: trust an existing `AttrRDMADevice`, otherwise call `GetRdmaDevice` and publish the name.

pkg/inventory/db.go
cabrinha: Ack — using `GetRdmaDevice` so we keep both the rdmamap result and the sysfs fallback name in one place, instead of going through `IsRDmaDeviceForNetdevice` + `isRdmaDeviceInSysfs` (which discarded it).

aojea: oh, sorry, I was leaving notes to do the review, because the code was calling the same functions

aojea:
/approve

@Bowser1704 can you do the final lgtm since it seems you refactored this code recently?

Issue #229, opened 2026-06-17
Add context logging for better log tracing


gauravkghildiyal:
We should migrate our logging to contextual logs so it becomes easier to associate logs to a specific pod, claim, device, etc. 

https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/3077-contextual-logging

gauravkghildiyal:
@ianchakeres Said he'll be helping out with this. Thank you!

fmuyassarov:
This is a good move!

PR #262, opened 2026-07-17
Migrate nri_hooks.go to contextual logging

ianchakeres:
#### What type of PR is this?

/kind cleanup

Migrates `pkg/driver/nri_hooks.go` from printf-style `klog` calls to contextual logging ([KEP-3077](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/3077-contextual-logging)):

- Each NRI hook (`Synchronize`, `CreateContainer`, `RunPodSandbox`, `StopPodSandbox`, `RemovePodSandbox`, `Shutdown`) derives its logger from the incoming context via `klog.FromContext` and attaches identifying key-value pairs (`pod` as `klog.KRef`, `podUID`, `container`) once with `klog.LoggerWithValues`.
- The enriched logger is propagated to helpers through `klog.NewContext`, so every log line emitted during a hook invocation carries the pod/claim/device it belongs to — making it possible to trace logs for a specific pod, claim, or device as the issue asks.
- All log calls are now structured (`logger.Info` / `logger.V(n).Info` / `logger.Error` with key-value pairs); original verbosity levels are preserved.
- `attachRdmaToNS` and `attachNetdevToNS` now accept a `context.Context`; the `pod` parameter of `attachNetdevToNS` was dropped as the logger now carries pod identity. The async ResourceClaim status update inherits a claim-scoped logger in its background context.

This is the first file migrated for #229; follow-up PRs can cover the remaining packages.

Part of #229

- Error-path logs that were previously `klog.Infof(...)` remain at Info level (with an `err` key) rather than being promoted to `logger.Error`, to keep log-level behavior identical. Happy to promote them if preferred.
- Verified with `GOOS=linux go build`, `go vet`, and `gofmt`.

```release-note
NONE
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/driver/nri_hooks.go
+++ b/pkg/driver/nri_hooks.go
@@ -39,15 +39,16 @@ import (
 // the information necessary should passed to the NRI hooks via the np.podConfigStore so it can be executed
 // quickly.
 
-func (np *NetworkDriver) Synchronize(_ context.Context, pods []*api.PodSandbox, containers []*api.Container) ([]*api.ContainerUpdate, error) {
-	klog.Infof("Synchronized state with the runtime (%d pods, %d containers)...",
-		len(pods), len(containers))
+func (np *NetworkDriver) Synchronize(ctx context.Context, pods []*api.PodSandbox, containers []*api.Container) ([]*api.ContainerUpdate, error) {
+	logger := klog.FromContext(ctx)
+	logger.Info("Synchronized state with the runtime", "pods", len(pods), "containers", len(containers))
 
 	// livePodNetNs map tracks live pods by UID and their network namespace paths.
 	livePodNetNs := make(map[types.UID]string)
 	for _, pod := range pods {
-		klog.Infof("Synchronize Pod %s/%s UID %s", pod.Namespace, pod.Name, pod.Uid)
-		klog.V(2).Infof("pod %s/%s: namespace=%s ips=%v", pod.GetNamespace(), pod.GetName(), getNetworkNamespace(pod), pod.GetIps())
+		podLogger := klog.LoggerWithValues(logger, "pod", klog.KRef(pod.Namespace, pod.Name), "podUID", pod.Uid)
+		podLogger.Info("Synchronize Pod")
+		podLogger.V(2).Info("Pod network details", "netns", getNetworkNamespace(pod), "ips", pod.GetIps())
 		livePodNetNs[types.UID(pod.Uid)] = getNetworkNamespace(pod)
 	}
 
@@ -63,7 +64,9 @@ func (np *NetworkDriver) Synchronize(_ context.Context, pods []*api.PodSandbox, 
 
 // CreateContainer handles container creation requests.
 func (np *NetworkDriver) CreateContainer(ctx context.Context, pod *api.PodSandbox, ctr *api.Container) (*api.ContainerAdjustment, []*api.ContainerUpdate, error) {
-	klog.V(2).Infof("CreateContainer Pod %s/%s UID %s Container %s", pod.Namespace, pod.Name, pod.Uid, ctr.Name)
+	logger := klog.LoggerWithValues(klog.FromContext(ctx), "pod", klog.KRef(pod.Namespace, pod.Name), "podUID", pod.Uid, "container", ctr.Name)
+	ctx = klog.NewContext(ctx, logger)
+	logger.V(2).Info("CreateContainer")
 	start := time.Now()
 	status := statusNoop
 	defer func() {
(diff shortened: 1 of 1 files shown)


aojea:
@ianchakeres at first sight this looks good, is it ready for review?

ianchakeres:
@aojea - yes. It is ready to review. PTAL

pkg/driver/nri_hooks.go
```diff
@@ -246,14 +254,15 @@ func attachRdmaToNS(linkDev, ns string, resourceClaimStatusDevice *resourceapply
 // attachNetdevToNS moves the host network interface into the pod network namespace,
 // applies all associated configuration (ethtool, eBPF, routes, rules, neighbors),
 // and records the resulting status conditions on resourceClaimStatusDevice.
-func attachNetdevToNS(pod *api.PodSandbox, ns, deviceName string, config DeviceConfig, resourceClaimStatusDevice *resourceapply.AllocatedDeviceStatusApplyConfiguration) error {
```
aojea, line 249: we should keep the pod reference and use it for logging the pod name and namespace as before

pkg/driver/nri_hooks.go
```diff
@@ -359,15 +370,16 @@ func (np *NetworkDriver) StopPodSandbox(ctx context.Context, pod *api.PodSandbox
...
+	logger := klog.FromContext(ctx)
 	// get the pod network namespace
 	ns := getNetworkNamespace(pod)
 	if ns == "" {
 		// some version of containerd does not send the network namespace information on this hook so
 		// we workaround it using the local copy we have in the db to associate interfaces with Pods via
 		// the network namespace id.
 		if podConfig.NetNS == "" {
-			klog.Warningf("StopPodSandbox: network namespace for DRANET pod %s/%s (UID %s) is unknown; skipping explicit device detach and relying on kernel netns teardown", pod.Namespace, pod.Name, pod.Uid)
+			logger.Info("StopPodSandbox: network namespace for DRANET pod is unknown; skipping explicit device detach and relying on kernel netns teardown")
```
aojea, line 382: I can't remember if there is any difference on `Warningf` vs Info when moving to structured logging

pkg/driver/nri_hooks.go
```diff
@@ -227,10 +234,11 @@ func (np *NetworkDriver) runPodSandbox(_ context.Context, pod *api.PodSandbox, p
...
 // attachRdmaToNS moves the RDMA link device into the pod network namespace and
 // records the RDMALinkReady status condition on resourceClaimStatusDevice.
-func attachRdmaToNS(linkDev, ns string, resourceClaimStatusDevice *resourceapply.AllocatedDeviceStatusApplyConfiguration) error {
-	klog.V(2).Infof("RunPodSandbox processing RDMA device: %s", linkDev)
+func attachRdmaToNS(ctx context.Context, linkDev, ns string, resourceClaimStatusDevice *resourceapply.AllocatedDeviceStatusApplyConfiguration) error {
+	logger := klog.LoggerWithValues(klog.FromContext(ctx), "rdmaDevice", linkDev, "netns", ns)
+	logger.V(2).Info("RunPodSandbox processing RDMA device")
 	if err := nsAttachRdmadev(linkDev, ns); err != nil {
-		klog.Infof("RunPodSandbox error getting RDMA device %s to namespace %s: %v", linkDev, ns, err)
+		logger.Info("RunPodSandbox error moving RDMA device to namespace", "err", err)
```
gauravkghildiyal: As you pointed out, this is a preexisting issue but it would be ideal to log these as Error instead of Info.

pkg/driver/nri_hooks.go
gauravkghildiyal: The change seems to be replacing `pod *api.PodSandbox` with `ctx context.Context`, so the pod name, namespace etc. are actually passed through the context itself (which is one of the essence of context logging)

pkg/driver/nri_hooks.go
gauravkghildiyal: Right, I think it was that context logging doesn't have a "Warning" level, so one has to choose between an Error or Info depending on your severity. In this case, I think it seems reasonable to treat this as an Info. Sounds fair?

gauravkghildiyal (commented):
Thanks @ianchakeres, mostly looking good.. Just one request and waiting to hear back from @aojea

pkg/driver/nri_hooks.go
ianchakeres: Done — promoted all the error-path logs in this file (the RDMA/netdev move failures, ethtool, eBPF, routing, rules, neighbor config, and the async claim-status update) from `Info` to `Error`. Since `logr`'s `Error` takes the error as its first argument, the `"err"` key is now dropped and the error is passed as `logger.Error(err, …)`. PTAL.

pkg/driver/nri_hooks.go
ianchakeres: I agree with @gauravkghildiyal's comment. Antonio let me know if you feel we should do it differently.

pkg/driver/nri_hooks.go
ianchakeres: I agree with Gaurav, and I think Info is probably the right severity.

pkg/driver/nri_hooks.go
aojea: I see,  I was ignoring the fact that data was plumbed through the context, my bad

aojea:
/lgtm

@gauravkghildiyal for approval

Thanks Ian

Issue #273, opened 2026-08-03
Flaky e2e: "no matching resources found" from kubectl wait --for=condition=ready after apply


cabrinha:
### What happened

The Bats e2e workflow is flaky. Multiple runs on the same (runtime-identical) code failed on **different** tests, each with the identical error:

```
error: no matching resources found
```

emitted by `kubectl wait --for=condition=ready pods -l <label>` immediately after `kubectl apply` of a workload (Deployment/ResourceClaimTemplate).

Observed on PR #260:
- one run failed test *"dummy interface with IP addresses ResourceClaimTemplate"* (`tests/e2e.bats:149`, `-l app=MyApp`)
- the next run (runtime-identical code) failed test *"reapply pod with dummy resource claim"* (`tests/e2e.bats:367`, `-l app=reapplyApp`)

### Root cause

`kubectl wait --for=condition=ready` does **not** wait for the object to be *created*. When the workload's controller (ReplicaSet) hasn't produced the Pod object yet, `kubectl wait -l <label>` matches zero resources and exits non-zero immediately with `no matching resources found` — as opposed to a `condition ready` timeout, which is what an actual scheduling/device failure would produce.

This `apply`-then-immediately-`wait -l` pattern is used ~18 times in `tests/e2e.bats`, so under CI load it can trip on any test, which is why the failing test differs run to run.

### Proposed fix

Wait for the Pod objects to exist before waiting for readiness, e.g. a helper that runs `kubectl wait --for=create pods -l <label>` (kubectl 1.31+) before `--for=condition=ready`. PR to follow.

gauravkghildiyal:
Fixed by https://github.com/kubernetes-sigs/dranet/pull/274.

Thanks @cabrinha

PR #274, opened 2026-08-03
tests: fix flaky e2e by waiting for pods to be created before ready

cabrinha:
## What

Adds a `wait_for_ready_pods` helper to `tests/e2e.bats` that waits for pods to be **created** before waiting for them to become ready, and routes all label-selector readiness waits through it.

## Why

`kubectl wait --for=condition=ready pods -l <label>` does not wait for the target object to exist. When a workload's ReplicaSet hasn't created the Pod objects yet, the selector matches nothing and `kubectl wait` exits immediately with:

```
error: no matching resources found
```

instead of waiting. The `apply`-then-immediately-`wait -l` pattern is used ~18 times in the suite, so under CI load it trips on a different test each run — the suite is flaky rather than deterministically broken.

## How

`wait_for_ready_pods <selector> [timeout]` runs `kubectl wait --for=create pods -l <selector>` (kubectl 1.31+) before `--for=condition=ready`. All existing label-selector readiness waits now call it; behavior is otherwise unchanged (same selectors, same timeouts). Named-pod waits (`pod/<name>`) are left as-is since those objects exist synchronously after `apply`.

### DIFF
--- a/tests/e2e.bats
+++ b/tests/e2e.bats
@@ -105,6 +105,19 @@ setup_tcx_filter() {
   docker exec "$CLUSTER_NAME"-worker2 bash -c "ip link set up dev dummy0"
 }
 
+# wait_for_ready_pods waits for the pods matching a label selector to first be
+# created and then become ready. The `--for=create` step is required because
+# `kubectl wait --for=condition=ready` does not wait for the object to exist: if
+# the workload's controller (e.g. a ReplicaSet) has not created the Pod objects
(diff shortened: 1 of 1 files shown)


gauravkghildiyal:
Thanks @cabrinha

/ok-to-test

aojea:
@cabrinha commits can not make references to issues , please fix the commit, is a policy to avoid github spam

> Fixes https://github.com/kubernetes-sigs/dranet/issues/273

PR #277, opened 2026-08-04
Add per-interface ARP configuration for Pod network devices

OguzPastirmaci:
#### What type of PR is this?

/kind feature
/kind api-change

ARP sysctls for an individual interface are reset when a network device moves into another network namespace. Linux initializes the interface from the destination namespace's `conf/default`, so `arp_ignore` and `arp_announce` values configured on the host are not preserved.

Pod `securityContext.sysctls` is not a practical replacement. The `net.ipv4.conf.*` settings are not part of Kubernetes' safe sysctl set and require cluster administrators to allow them as unsafe sysctls. The path for the assigned interface also does not exist when sandbox sysctls are applied because DRANET moves the device into the namespace afterward.

This matters on nodes with multiple NICs where several interfaces share one IP subnet, including Oracle Cloud RDMA configurations. Without `arp_ignore: 1`, an interface can answer ARP requests for an address assigned to another interface. Without `arp_announce: 2`, ARP requests can use a source address from the wrong interface, causing neighbor resolution to select the wrong link.

For workloads using VFs, an equivalent ARP policy for the entire namespace can be applied through the tuning CNI meta plugin configured by the SR-IOV Network Operator. This is how OCI RDMA VF workloads are configured today. See [this SR-IOV network configuration](https://github.com/oracle-quickstart/oci-hpc-oke/blob/main/terraform/files/nvidia-network-operator/sriov-network.yaml).

Network interfaces assigned through DRANET do not use that SR-IOV CNI path. This PR lets DRANET apply `arpIgnore` and `arpAnnounce` directly to the assigned interface inside the Pod network namespace, rather than to the namespace as a whole.

N/A

This PR adds API and runtime support for interface specific ARP configuration that is independent of the provider. The new fields are optional, so existing configurations remain unchanged when they are omitted. Users can request the settings explicitly.

A later PR will make the OKE provider request the recommended values automatically for physical RDMA interfaces, without requiring user configuration.

Supported values are:

- `arpIgnore`: `0`, `1`, `2`, `3`, or `8`
- `arpAnnounce`: `0`, `1`, or `2`

Linux uses the maximum of the namespace wide `conf/all` value and the value for an individual interface. DRANET changes only the interface value, so it cannot reduce the effective value below `conf/all`.

#### Does this PR introduce a user facing change?

```release-note
Add optional `arpIgnore` and `arpAnnounce` interface settings for configuring ARP behavior on network devices assigned to Pods.
```


pkg/driver/interface_arp.go
```diff
@@ -0,0 +1,98 @@
...
+// Pod network namespace. These live under /proc/sys, so unlike the rest of the
+// interface configuration they cannot be set through a netlink handle and
+// require joining the namespace.
+func applyInterfaceARPConfig(containerNs netns.NsHandle, ifName string, interfaceConfig apis.InterfaceConfig) error {
+	if !hasARPConfig(interfaceConfig) {
+		return nil
+	}
+
+	result := make(chan error, 1)
+	go func() {
```
aojea, line 66: why do we run inside a goroutine here?

we do this in other places like without having to use goroutines

https://github.com/kubernetes-sigs/dranet/blob/88a1dcd9050160a491be7f476d1d6fcafc9daeb3/pkg/driver/netnamespace.go#L303

and https://github.com/kubernetes-sigs/dranet/blob/88a1dcd9050160a491be7f476d1d6fcafc9daeb3/pkg/driver/netnamespace.go#L192-L219

aojea:
overall looks good, one comment about the use of goroutines for entering on namespaces

pkg/driver/interface_arp.go
drkennetz: In a follow-up, should we wrap these `lock / defer` in a function? As noted, this pattern has shown up 3 times now, but this PR isn't the right place because it touches unnecessary parts of the code.

aojea: I may not understand the problem

drkennetz: In your linked inline examples:
```go
defer runtime.UnlockOSThread()          // (A) runs last
if err := netns.Set(containerNs); ...
defer netns.Set(origns) // nolint:errcheck   // (B) runs first, error IGNORED
```
This pattern would (B) try to restore the namespace, then (A) unconditionally unlock. If (B) fails, the error is swallowed and the thread is unlocked while still in the container's netns. Any goroutine later scheduled onto that thread would run in that poisoned netns, no?

With the throwaway go func() you can sacrifice that go routine and its thread and let the caller continue normally.

Unless I'm understanding incorrectly, I actually think there's a subtle bug in those other two linked examples where you could end up with a poisoned thread.

aojea: oh, subtle, goroutines and namespaces are complicated indeed https://news.ycombinator.com/item?id=14470231

We may need to follow up and maybe consolidate this into a library like the cni folks https://github.com/containernetworking/plugins/blob/main/pkg/ns/ns_linux.go or just reuse theirs

drkennetz: This is my thought too, let's just try to capture it correctly (whichever way that is) and re-use that implementation.

PR #279, opened 2026-08-05
Update DeviceIdentifiers.Name to use interfaceName

npinaeva:
#### What type of PR is this?
/kind bug

device.Name is normalized and coming from PCIAddress or netlink interface name. PCI address is already a part of DeviceIdentifiers, so there is no use in adding its normalized version.
For interfaceName, if the original string is already a valid DNS-1123 label, it's returned as is. Otherwise, it's encoded using Base32, prefixed with NormalizedPrefix, and returned. The means, e.g. if an interfaceName has a capital letter, DeviceIdentifiers.Name will be something like net-mr2w23lzkjqws3bq instead of an actual interface name.

Since device.Name in general is not a hardware identifier, replacing it with the actual ifaceName to avoid getting no useful IDs for non-PCI devices.

I guess since we haven't had a release with providers yet, it is not a braking change?

```release-note
NONE
```

### DIFF
--- a/pkg/inventory/db.go
+++ b/pkg/inventory/db.go
@@ -532,14 +532,19 @@ func (db *DB) getProviderAttributes(device *resourceapi.Device, instance cloudpr
 		return nil
 	}
 
-	id := cloudprovider.DeviceIdentifiers{
-		Name: device.Name,
+	id := cloudprovider.DeviceIdentifiers{}
+	if ifaceNameAttr, ok := device.Attributes[apis.AttrInterfaceName]; ok && ifaceNameAttr.StringValue != nil {
+		id.Name = *ifaceNameAttr.StringValue
 	}
 	if macAttr, ok := device.Attributes[apis.AttrMac]; ok && macAttr.StringValue != nil {
 		id.MAC = *macAttr.StringValue
 	}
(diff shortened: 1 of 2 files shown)


gauravkghildiyal:
Thanks @npinaeva 

> I guess since we haven't had a release with providers yet, it is not a braking change?

This looks like a bug fix to me. I see that only Azure is using the Name field to get attributes but the logic expects the Name to be the interface name which in more cases would have been wrong:

https://github.com/kubernetes-sigs/dranet/blob/ee43ec03d59a354b7b94afce726f83a9d5aecde9/pkg/cloudprovider/azure/azure.go#L202

@tamilmani1989 can you sanity check this please and LGTM if it looks good?

/ok-to-test
/approve

npinaeva:
> I guess since we haven't had a release with providers yet, it is not a braking change?

I was trying to see if I need release notes, considering it is a user-facing change, but the feature itself hasn't been "released" yet?

gauravkghildiyal:
> > I guess since we haven't had a release with providers yet, it is not a braking change?
> 
> I was trying to see if I need release notes, considering it is a user-facing change, but the feature itself hasn't been "released" yet?

Ah I see what you mean, yeah we can skip this one.

gauravkghildiyal:
I consider this is low risk and clearly seems like a fix, proceeding

/lgtm

PR #280, opened 2026-08-06
Update Helm chart

OguzPastirmaci:
#### What type of PR is this?

/kind bug

`install.yaml` mounts `/var/run/dranet` from the host. The Helm chart did not include this mount.

The Dranet binary stores its bbolt database at `/var/run/dranet/dranet.db` by default. Without the host mount, the database stays in the container file system. Kubernetes removes this database when it replaces the Dranet pod.

Dranet stores prepared pod device configuration in this database. The NRI hooks need this configuration after a Dranet restart.

This PR mounts the host `/var/run/dranet` directory in the Dranet container. The host path uses `DirectoryOrCreate`. The database now survives Dranet pod replacement.

This PR also:

- Adds Helm values for `dbPath`, `profileProvider`, `webhookURL`, and `featureGates`.
- Adds `webhook` as a cloud provider hint.
- Lets users add service account annotations.
- Validates webhook settings and database paths.

N/A

All new argument values are optional. If users omit a value, the Dranet binary uses its built in default.

Set `args.dbPath` to an empty string to disable database persistence. If users set a custom database path, they must place it under `/var/run/dranet`.

An older chart stores the database in the container file system. Dranet cannot move this database to the new host mount during an upgrade.

Before the first upgrade, stop pods that use Dranet managed devices. Wait for the pods to terminate. Then upgrade the Helm release.

```release-note
Updated the Helm chart to preserve pod device configuration across pod replacements. Action required: before you upgrade from a chart that does not mount `/var/run/dranet`, stop pods that use devices managed by Dranet.
```

### DIFF
--- a/deployments/helm/dranet/README.md
+++ b/deployments/helm/dranet/README.md
@@ -25,20 +25,37 @@ The following table lists the configurable parameters and their default values:
 | `podLabels` | Labels to add to pods | `{}` |
 | `logVerbosity` | Log verbosity level | `4` |
 | `metricsPort` | Port for the metrics/healthz server and readiness probe | binary default: `9177` |
-| `metricsPath` | HTTP path for the readiness probe | `/healthz` |
+| `metricsPath` | HTTP path for the startup and readiness probes | `/healthz` |
 | `tolerations` | Pod tolerations | `[{operator: Exists, effect: NoSchedule}]` |
 | `resources.requests.cpu` | CPU resource request | `100m` |
 | `resources.requests.memory` | Memory resource request | `50Mi` |
 | `resources.limits.cpu` | CPU resource limit | `""` (not set) |
 | `resources.limits.memory` | Memory resource limit | `""` (not set) |
+| `serviceAccount.annotations` | Annotations to add to the service account | `{}` |
 | `args.filter` | CEL expression to filter network interface attributes | see binary default |
+| `args.dbPath` | Persistent bbolt database path; set to `""` to use in-memory state | binary default: `/var/run/dranet/dranet.db` |
 | `args.inventoryMinPollInterval` | Minimum interval between two consecutive inventory polls | binary default: `2s` |
 | `args.inventoryMaxPollInterval` | Maximum interval between two consecutive inventory polls | binary default: `1m` |
 | `args.inventoryPollBurst` | Number of inventory polls that can be run in a burst | binary default: `5` |
 | `args.moveIBInterfaces` | If true, InfiniBand (IPoIB) interfaces are moved into the pod network namespace | binary default: `true` |
-| `args.cloudProviderHint` | Hint for the cloud provider plugin (`GCE`, `AZURE`, `OKE`, `NONE`); auto-detected if unset | binary default: `""` |
+| `args.cloudProviderHint` | Hint for the cloud provider plugin (`GCE`, `AZURE`, `OKE`, `AWS`, `ALIBABA`, `webhook`, `NONE`); auto-detected if unset | binary default: `""` |
+| `args.profileProvider` | Provider for user profile configuration (`cloud`, `webhook`, `none`) | binary default: `cloud` |
+| `args.webhookURL` | HTTP, HTTPS, or Unix socket URL; required when either provider uses `webhook` | binary default: `""` |
+| `args.featureGates` | Comma-separated feature gate settings in `key=value` format | binary default: `""` |
 
 > **Note:** All `args.*` fields are optional. When omitted, the flag is not passed to the binary and the binary's built-in default applies.
+
+The chart mounts `/var/run/dranet` from the host so the default database survives
+DRANET pod replacement. Custom database paths must be placed under that directory
+to remain persistent. Set `args.dbPath` to an empty string to use in-memory state.
+
+When upgrading from a chart that does not mount `/var/run/dranet`, the existing
+container-local database cannot be copied into the new host path. Before the first
+upgrade that enables this mount, stop workloads that use DRANET-managed devices.
+After the upgrade, restart those workloads.
+
+When `args.profileProvider` or `args.cloudProviderHint` is `webhook`, set
+`args.webhookURL` to the webhook endpoint.
 
 Parameters can be set at install time using `--set` or a custom values file:
 
--- a/deployments/helm/dranet/values.yaml
+++ b/deployments/helm/dranet/values.yaml
@@ -19,14 +19,18 @@ logVerbosity: 4
 metricsPort: ~
 metricsPath: /healthz
 
-# dranet daemon arguments — omit any field to use the binary's built-in default
+# DRANET daemon arguments. Omit a field to use the binary's built-in default.
 args: {}
 #  filter: '!("dra.net/type" in attributes) || attributes["dra.net/type"].StringValue  != "veth"'
+#  dbPath: "/var/run/dranet/dranet.db"
 #  inventoryMinPollInterval: "2s"
 #  inventoryMaxPollInterval: "1m"
 #  inventoryPollBurst: 5
 #  moveIBInterfaces: true
 #  cloudProviderHint: ""
+#  profileProvider: "cloud"
+#  webhookURL: ""
+#  featureGates: ""
 
 # kubeletRootDir is the kubelet data directory (its --root-dir). The driver's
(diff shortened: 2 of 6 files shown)


.github/workflows/helm-lint.yml
```diff
@@ -4,6 +4,10 @@ on:
     branches: [ main ]
     paths:
       - 'deployments/helm/**'
+      - 'cmd/dranet/app.go'
+      - 'install.yaml'
```
fmuyassarov: I guess there is nothing that ties the helm charts to `install.yaml` right? Let's drop this if so.

fmuyassarov: same here

deployments/helm/dranet/README.md
```diff
@@ -25,21 +25,38 @@ The following table lists the configurable parameters and their default values:
...
-| `metricsPath` | HTTP path for the readiness probe | `/healthz` |
+| `metricsPath` | HTTP path for the startup and readiness probes | `/healthz` |
 | `tolerations` | Pod tolerations | `[{operator: Exists, effect: NoSchedule}]` |
 | `resources.requests.cpu` | CPU resource request | `100m` |
 | `resources.requests.memory` | Memory resource request | `50Mi` |
 | `resources.limits.cpu` | CPU resource limit | `""` (not set) |
 | `resources.limits.memory` | Memory resource limit | `""` (not set) |
+| `serviceAccount.annotations` | Annotations to add to the service account | `{}` |
 | `args.filter` | CEL expression to filter network interface attributes | see binary default |
+| `args.dbPath` | Persistent database path; set to `""` to use in-memory state | binary default: `/var/run/dranet/dranet.db` |
```
fmuyassarov:
```suggestion
| `args.dbPath` | Persistent bblot database path; set to `""` to use in-memory state | binary default: `/var/run/dranet/dranet.db` |
```

deployments/helm/dranet/README.md
```diff
@@ -25,21 +25,38 @@ The following table lists the configurable parameters and their default values:
...
 | `args.inventoryMinPollInterval` | Minimum interval between two consecutive inventory polls | binary default: `2s` |
 | `args.inventoryMaxPollInterval` | Maximum interval between two consecutive inventory polls | binary default: `1m` |
 | `args.inventoryPollBurst` | Number of inventory polls that can be run in a burst | binary default: `5` |
 | `args.moveIBInterfaces` | If true, InfiniBand (IPoIB) interfaces are moved into the pod network namespace | binary default: `true` |
-| `args.cloudProviderHint` | Hint for the cloud provider plugin (`GCE`, `AZURE`, `OKE`, `NONE`); auto-detected if unset | binary default: `""` |
+| `args.cloudProviderHint` | Hint for the cloud provider plugin (`GCE`, `AZURE`, `OKE`, `AWS`, `ALIBABA`, `webhook`, `NONE`); auto-detected if unset | binary default: `""` |
+| `args.profileProvider` | Provider for user profile configuration (`cloud`, `webhook`, `none`) | binary default: `cloud` |
+| `args.webhookURL` | HTTP, HTTPS, or Unix socket URL; required when either provider uses `webhook` | binary default: `""` |
+| `args.featureGates` | Comma-separated feature gate settings | binary default: `""` |
```
fmuyassarov: nit, can we add `key=value` somewhere in the description ?

deployments/helm/dranet/values.yaml
```diff
@@ -19,14 +19,18 @@ logVerbosity: 4
 metricsPort: ~
 metricsPath: /healthz
 
-# dranet daemon arguments — omit any field to use the binary's built-in default
+# Dranet daemon arguments. Omit a field to use the binary's built-in default.
```
fmuyassarov: nit please, since we are touching it here anyways
```suggestion
# DRANET daemon arguments. Omit a field to use the binary's built-in default.
```

hack/verify-helm.sh
```diff
@@ -0,0 +1,119 @@
...
+  grep -Fq -- "${argument}" <<<"${configured_manifest}" || fail "${argument} is missing"
+done
+
+grep -Fq -- 'example.com/test: value' <<<"${configured_manifest}" || fail "service account annotation is missing"
+
+expect_rejected "webhook cloud provider without a URL" --set-string args.cloudProviderHint=webhook
+expect_rejected "webhook profile provider without a URL" --set-string args.profileProvider=webhook
+expect_rejected "non-string service account annotation" --set serviceAccount.annotations.test=1
+expect_rejected "database path outside the persistent mount" --set-string args.dbPath=/var/lib/dranet/dranet.db
+expect_rejected "database path that leaves the persistent mount" --set-string args.dbPath=/var/run/dranet/../dranet.db
```
fmuyassarov: can we add a visible confirmation with echo at the end that we have succeeded?

fmuyassarov (commented):
Thanks for the fixes @OguzPastirmaci . It is true that directory mounting was missed as part of #115 implementation and I don't see any reason to not replicate it to Helm charts.  +1 from me.

However regarding the template validation, in my experience bash script can become difficult to maintain over time and it also doesn't provide much in terms of structured assertions. I do think validation is something we should have but maybe it would be worth checking if something like https://github.com/helm-unittest/helm-unittest would be a better fit? It doesn't have to be done in this patch. I’m completely fine to see it as a follow up. And if you would rather not spend time on it that's completely fine, I can take a look into it.

hack/verify-helm.sh
fmuyassarov: this can be disregarded if we drop the bash script of course

OguzPastirmaci:
> Thanks for the fixes @OguzPastirmaci . It is true that directory mounting was missed as part of #115 implementation and I don't see any reason to not replicate it to Helm charts. +1 from me.
> 
> However regarding the template validation, in my experience bash script can become difficult to maintain over time and it also doesn't provide much in terms of structured assertions. I do think validation is something we should have but maybe it would be worth checking if something like https://github.com/helm-unittest/helm-unittest would be a better fit? It doesn't have to be done in this patch. I’m completely fine to see it as a follow up. And if you would rather not spend time on it that's completely fine, I can take a look into it.

Thanks for the review, @fmuyassarov. I addressed the comments and removed the shell verifier. I will start a separate discussion about Helm chart validation.

fmuyassarov (approved):
Thank you!

fmuyassarov:
@aojea @gauravkghildiyal could you folks help with approval here if LGTY too?

deployments/helm/dranet/README.md
```diff
@@ -25,21 +25,38 @@ The following table lists the configurable parameters and their default values:
...
 > **Note:** All `args.*` fields are optional. When omitted, the flag is not passed to the binary and the binary's built-in default applies.
 
+The chart mounts `/var/run/dranet` from the host so the default database survives
+DRANET pod replacement. Custom database paths must be placed under that directory
+to remain persistent. Set `args.dbPath` to an empty string to use in-memory state.
+
+When upgrading from a chart that does not mount `/var/run/dranet`, the existing
+container-local database cannot be copied into the new host path. Stop workloads
+that use DRANET-managed devices and wait for their pods to terminate before the
+first upgrade that enables this mount.
```
gauravkghildiyal: Good idea with thinking of a recommendation here.

Is it simpler to recommend a rolling restart of their workload pods when this new DRANET with DB in host path starts getting used?

gauravkghildiyal (commented):
Thanks @OguzPastirmaci 

One small question and this should be good to go.

fmuyassarov:
Forgot to comment, if you are going to make another push, I would kindly ask to perhaps squash the last commit (https://github.com/kubernetes-sigs/dranet/pull/280/changes/9b906a14d47f7b6bd1f4da9097a452cfb8137bb4) into the first one

deployments/helm/dranet/README.md
OguzPastirmaci: Thanks. I wanted to keep the recommendation generic. Distributed AI workloads commonly use gang scheduling, so they need a coordinated restart.

I can simplify the recommendation to:
“Before the first upgrade that enables this mount, stop workloads that use DRANET-managed devices. After the upgrade, restart those workloads.”

Let me know how this sounds.

gauravkghildiyal: Thanks for the clarification. Your new proposal sounds fair to me.

PR #282, opened 2026-08-10
Fix webhook API

### LINKED ISSUE #234: BYODP webhook providers cannot garbage-collect orphaned IP allocations when using whereabouts
### Problem

The BYODP documentation (`site/content/docs/contributing/webhook-providers.md`, introduced in PR #223) clearly states that GC (garbage collection) of orphaned allocations is the webhook provider's responsibility:

> "Best-effort teardown: A failed ReleaseProfileConfig is logged but not retried by DRANET. The provider therefore owns leak reclamation and must be able to garbage-collect orphaned allocations on its own, otherwise resources leak permanently."

However, the current design does not give webhook providers enough information to implement GC when using whereabouts as the IPAM backend.

### How IP leak happens

When `/ReleaseProfileConfig` fails (e.g., webhook crash, node restart), the IP reservation remains in the whereabouts `IPPool` even after the Pod is deleted. Since dranet does not retry failed releases, these orphaned reservations leak permanently unless someone garbage-collects them.

### Why the provider cannot GC

whereabouts provides `ip-reconciler` as its GC mechanism. ip-reconciler checks each `IPPool` allocation and deletes orphans by examining the `PodRef` field:

```go
// whereabouts IPAllocation (pkg/api/.../ippool_types.go)
type IPAllocation struct {
    ContainerID string `json:"id"`
    PodRef      string `json:"podref"`   // "<namespace>/<name>"
    IfName      string `json:"ifname,omitempty"`
}
```

The `PodRef` value comes from `CNI_ARGS` when the provider executes the whereabouts binary:
```
CNI_ARGS=...;K8S_POD_NAMESPACE=<ns>;K8S_POD_NAME=<name>;...
```

**The problem is that `ProfileRequest` does not contain Pod namespace or name:**

```go
// dranet ProfileRequest (pkg/cloudprovider/webhook/webhook.go)
type ProfileRequest struct {
    Device   cloudprovider.DeviceIdentifiers `json:"device"`
    ClaimUID types.UID                       `json:"claim_uid"`
    Config   *apis.NetworkConfig             `json:"config,omitempty"`
}
```

The provider receives only NIC identifiers (`Device`), the ResourceClaim UID (`ClaimUID`), and user intent (`Config.Profile`). **There is no Pod information in the wire type.** The provider cannot fill `K8S_POD_NAMESPACE` and `K8S_POD_NAME` with real values.

The reference implementation `cmd/webhook-whereabouts` uses hardcoded dummy values for this reason:
```go
"CNI_ARGS=IgnoreUnknown=1;K8S_POD_NAMESPACE=default;K8S_POD_NAME=pod-whereabouts;K8S_POD_INFRA_CONTAINER_ID=" + string(req.ClaimUID),
```

This causes all IP allocations to have the same PodRef `"default/pod-whereabouts"`, which makes ip-reconciler unable to identify orphans.

### Additional concern: network-status annotation

Even if the provider could somehow set the correct PodRef, ip-reconciler has a second check. It verifies that the Pod's `k8s.v1.cni.cncf.io/network-status` annotation contains the allocated IP ([whereabouts issue #638](https://github.com/k8snetworkplumbingwg/whereabouts/issues/638) reports a similar problem). This annotation is typically written by Multus after CNI execution. In dranet-based setups that do not use Multus, this annotation may not exist, which would cause ip-reconciler to incorrectly delete valid allocations from running Pods.

### Summary

| What | Status |
|---|---|
| GC responsibility assigned to provider | Documented in PR #223 |
| Provider can implement GC with whereabouts ip-reconciler | **No** — wire type lacks Pod info for PodRef |
| Reference implementation handles GC | **No** — uses dummy PodRef |
| ip-reconciler works without network-status annotation | **No** — may incorrectly delete valid allocations |

### Possible solutions

I am not proposing a specific solution here, but listing some directions:

1. **Add Pod namespace/name to `ProfileRequest`** — dranet already knows which Pod triggered `NodePrepareResources`, so it could pass this information to the webhook
2. **dranet writes `network-status` annotation on the Pod** — this would make ip-reconciler work correctly

### Environment

- dranet: `main` branch (as of 2026-06-19)
- whereabouts: latest
- Kubernetes: v1.33+ (DRA v1)

kanlkan:
#### What type of PR is this?

/kind bug

This PR fix webhook API.
`GetProfileConfig` now receives the complete `ResourceClaim` instead of only its UID. The whereabouts webhook uses:
* the ResourceClaim UID as `CNI_CONTAINERID`
* the actual Pod namespace and name from `status.reservedFor`
* the same normalized `CNI_IFNAME` for both CNI ADD and DEL

#234 

None

```release-note
The Profile Provider webhook API has changed.
POST /GetProfileConfig now contains a complete ResourceClaim in the claim field instead of only claim_uid. Custom Profile Provider implementations must be updated before or together with DRANET. During a rolling upgrade, providers may need to temporarily accept both request formats.
POST /ReleaseProfileConfig remains UID-based because NodeUnprepareResources does not provide the complete ResourceClaim.
```


aojea:
@kanlkan Thanks for putting this together!

I think we should not add the garbage collector here. Since the webhook runs as a DaemonSet, even with leader election, it is not the right architectural place for a cluster-wide GC controller. For garbage collection, we should either rely on the existing whereabouts IPAM reconciler.

Regarding my original comment in [issue #234](https://github.com/kubernetes-sigs/dranet/issues/234#issuecomment-4810008301), let's scope this PR to just keep the Profile/API changes:

1. The DRANET changes to pass *resourceapi.ResourceClaim to GetProfileConfig.
2. The webhook changes in cmd/webhook-whereabouts/main.go that extract the real Pod identity from the claim and pass it via CNI_ARGS.

This cleanly and elegantly fixes the original bug (where the hardcoded pod name was breaking the existing CNI garbage collectors) without adding massive complexity to the webhook. 

Could you update the PR to strip out the new GC additions (gc.go) and keep just the API and documentation changes?

kanlkan:
@aojea 
Thank you for your suggestion. I'll update the PR to focus on the Profile/API changes.

However, I don't think the existing Whereabouts IPAM reconciler fully covers the garbage collection needs when Whereabouts is used through the webhook. So, after this PR is merged, I'll take a closer look at the behavior of the existing IPAM reconciler.
If I find any gaps, I'll follow up separately.

aojea:
> o, after this PR is merged, I'll take a closer look at the behavior of the existing IPAM reconciler.

I don't mind providing a separate deployment / component for that if there is no other piece in the ecosystem, but coupling it to the existing webhook does not look right for several reasons: scalability, security, separation of concerns , ...

kanlkan:
@aojea 
I’ve updated the PR to focus on the API changes and addressed the review comments. I’ve also updated the title and description. Could you please take another look?

tests/manifests/whereabouts_webhook_daemonset_localtest.yaml
```diff
@@ -0,0 +1,111 @@
+apiVersion: v1
```
aojea, line 1: this file seems a leftover of previous change

tests/manifests/whereabouts_webhook_daemonset_localtest.yaml
```diff
@@ -0,0 +1,111 @@
...
+    spec:
+      hostNetwork: true
+      serviceAccountName: whereabouts-webhook
+      tolerations:
+      - operator: Exists
+        effect: NoSchedule
+      containers:
+      - name: webhook
+        #image: dranet/webhook-whereabouts:test
+        image: 192.168.21.17:30003/dci-community-general/webhook-whereabouts:fix-webhook-api
```
aojea, line 78: yeah, this seems it leaked from some internal testing :)

pkg/cloudprovider/webhook/webhook.go
```diff
@@ -51,6 +51,14 @@ type Capabilities struct {
...
 // API Contracts (JSON payloads)
 type ProfileRequest struct {
+	Device cloudprovider.DeviceIdentifiers `json:"device"`
+	Claim  *resourceapi.ResourceClaim      `json:"claim"`
+	Config *apis.NetworkConfig             `json:"config,omitempty"`
+}
+
+// ProfileReleaseRequest contains the stable identity needed to release a
+// profile. NodeUnprepareResources does not provide the full ResourceClaim.
+type ProfileReleaseRequest struct {
```
aojea, line 61: I think here is ok to break the symmetry ,

aojea:
This LGTM, there is a file leak.
On another note, despite is a breaking change I think that the providers are so new that we may live with it?

/assign @gauravkghildiyal for more eyes

kanlkan:
@aojea 
Thank you for reviewing this PR. I am currently away on summer vacation until August 25th and will address your suggestions upon my return. Apologies for the delay.

aojea:
enjoy vacation, no rush

tests/manifests/whereabouts_webhook_daemonset_localtest.yaml
kanlkan: Thanks, I've removed this file.

tests/manifests/whereabouts_webhook_daemonset_localtest.yaml
kanlkan: Thanks, I've removed this file.

kanlkan:
@aojea 

>On another note, despite is a breaking change I think that the providers are so new that we may live with it?

That makes sense to me. The Profile Provider API is still new, so keeping the implementation simple seems reasonable.
I've added comments for release note in "Does this PR introduce a user-facing change?" section in the first comment of this PR.

aojea:
/ok-to-test
/lgtm

/assign @gauravkghildiyal 

can you PTAL? the only doubt I have is about the need to pass the claim object on release, right now it passes the uuid only ...

aojea:
needs rebase

gauravkghildiyal:
> can you PTAL? the only doubt I have is about the need to pass the claim object on release, right now it passes the uuid only ...

Sure, please allow me a day to get to this.

gauravkghildiyal:
/lgtm
/approve

> can you PTAL? the only doubt I have is about the need to pass the claim object on release, right now it passes the uuid only ...

Based on https://github.com/kubernetes/dynamic-resource-allocation/blob/e7d9a3a759f351bfd6406197da239463860fdb33/kubeletplugin/draplugin.go#L156-L164:

```golang
	// UnprepareResourceClaims must undo whatever work PrepareResourceClaims did.
	//
	// At the time when this gets called, the original ResourceClaims may have
	// been deleted already. They also don't get cached by the kubelet. Therefore
	// parameters for each ResourcClaim are only the UID, namespace and name.
	// It is the responsibility of the DRA driver to cache whatever additional
	// information it might need about prepared resources.
```

So essentially, it appears we can't do much better unless DRANET locally stores the ResourceClaim to pass it to ReleaseProfileConfig. Similar to how DRA expects the DRA Driver to cache the claim, DRANET could expect the downstream webhook server to cache it too if it needs.

That said, I do expect at some point this will be needed for the in-driver stuff. But we should wait for that requirement to show up.

PR #283, opened 2026-08-10
Add DRA NUMA list attribute

killianmuldoon:
#### What type of PR is this?
/kind feature

This PR uses the upstream DRA NUMA helpers to enable reporting of the upstream standard NUMA List attribute in DRANet where available. The list attribute is [described in a KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/6072-dra-standard-numanode/README.md). It enables matching by closest NUMA on more complex hardware where the network device NUMA node may not directly match the node of other hardware.

A bump to Kubernetes 1.37-rc.0 libraries is included as a dependency. This feature is enabled with a feature gate on Kubernetes 1.37, and the upstream helper libraries are only available in 1.37 tags.

N/A

```release-note
Added - standardized NUMA list attribute behind feature gate.
```


cmd/webhook-whereabouts/go.sum
```diff
@@ -47,40 +40,37 @@ github.com/stretchr/testify v1.11.1 h1:7s2iGBzp5EwR7/aIZr8ao5+dra3wiQyKjjFuvgVKu
...
+golang.org/x/tools v0.47.0/go.mod h1:dFHnyTvFWY212G+h7ZY4Vsp/K3U4/7W9TyVaAul8uCA=
 gopkg.in/inf.v0 v0.9.1 h1:73M5CoZyi3ZLMOyDlQh031Cx6N9NDJ2Vvfl76EDAgDc=
 gopkg.in/inf.v0 v0.9.1/go.mod h1:cWUDdTG/fYaXco+Dcufb5Vnc6Gp2YChqWtbxRZE0mXw=
 gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
 gopkg.in/yaml.v3 v3.0.1/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
-k8s.io/api v0.36.1 h1:XbL/EMj8K2aJpJtePmqUyQMsM0D4QI2pvl7YKJ20FTY=
-k8s.io/api v0.36.1/go.mod h1:KOWo4ey3TINlXjeHVuwB3i+tXXnu+UcwFBHlI/9dvEo=
-k8s.io/apimachinery v0.36.1 h1:G63Gjx2W+q0YD+72Vo8oY0nDnePVwnuzTmmy5ENrVSA=
-k8s.io/apimachinery v0.36.1/go.mod h1:ibYOR00vW/I1kzvi5SF0dRuJ52BvKtfvRdOn35GPQ+8=
+k8s.io/api v0.37.0-rc.0 h1:CgvGMEmo+Y37oJ7KfUr+ExMDU1isvQwmdgtz8q3ZxTM=
```
killianmuldoon, line 59: Not sure if the repo has a pattern for doing Kubernetes bumps - they're prerequisite here. I can wait until the repo does it's official bump and then rebase.

pkg/driver/dra_hooks.go
```diff
@@ -542,6 +542,12 @@ func (np *NetworkDriver) HandleError(ctx context.Context, err error, msg string)
 	runtime.HandleErrorWithContext(ctx, err, msg)
 }
 
+// WatchHealthStatus implements the Kubernetes 1.37 DRAPlugin contract. DRANET
+// disables the optional health service.
+func (np *NetworkDriver) WatchHealthStatus(context.Context, chan<- kubeletplugin.DeviceHealthReport) error {
+	return kubeletplugin.ErrHealthNotSupported
```
killianmuldoon, line 551: This is required in 1.37 - but not implemented as part of this change. I imagine it needs a good amount more conversation.

gauravkghildiyal: Is have a stub implementation required or optional?

pkg/features/features.go
```diff
@@ -22,6 +25,10 @@ var DefaultFeatureGate featuregate.FeatureGate = DefaultMutableFeatureGate
 
 func init() {
 	err := DefaultMutableFeatureGate.Add(map[featuregate.Feature]featuregate.FeatureSpec{
+		DRAListTypeAttributes: {
+			Default:    false,
+			PreRelease: featuregate.Alpha,
```
gauravkghildiyal, line 32: It seems like upstream suggests that the operator should match this to the feature gate value of the DRAListTypeAttributes in the control plane. Can we please add a note somewhere that to that effect and that enabling this in a cluster which doesn't have list type attributes enabled will result in failures.

cmd/webhook-whereabouts/go.sum
gauravkghildiyal: Thanks, I think it's okay to bump alongside this PR. Before DRANET's next release though, we would want to change this to the final upstream 1.37 release versions though (and not remain at RC)

gauravkghildiyal (commented):
Thanks @killianmuldoon

pkg/driver/dra_hooks.go
killianmuldoon: It's a required part of the interface in 1.37  - it was [added to the DRAPlugin interface as part of this change](https://github.com/kubernetes/dynamic-resource-allocation/commit/ccc762df3e62f0544d74a65fe5de1922cae0f5dd).

cmd/webhook-whereabouts/go.sum
killianmuldoon: I can track bumping to the full 1.37 tag and open a PR once it's published.

pkg/features/features.go
killianmuldoon: Added a short note in the quick start around enabling the list mode - let me know if that's what you meant.

cmd/webhook-whereabouts/go.sum
gauravkghildiyal: Thanks!

pkg/driver/dra_hooks.go
gauravkghildiyal: Gotcha, thanks!

pkg/features/features.go
gauravkghildiyal: Yep looks good. Would it also make sense to add it as a comment alongside the feature description above?

pkg/inventory/db.go
```diff
@@ -812,6 +829,19 @@ func (db *DB) addPCIAttributes(devices []resourceapi.Device, pciInfo *ghw.PCIInf
 		}
 		normalizedAddr := names.NormalizePCIAddress(*pciAddrAttr.StringValue)
 
+		if _, hasAttr := devices[i].Attributes[deviceattribute.StandardDeviceAttributeNUMANode]; !hasAttr {
```
gauravkghildiyal, line 837: pre-existing: Can I request you to also move the `GetPCIeRootAttributeByPCIBusID` up before the `if !inGhw` check since it follows the same standard device attribute rationale?

pkg/inventory/db.go
```diff
@@ -812,6 +829,19 @@ func (db *DB) addPCIAttributes(devices []resourceapi.Device, pciInfo *ghw.PCIInf
...
+			numaAttr, err := deviceattribute.GetNUMANodeAttributeByPCIBusID(
+				*pciAddrAttr.StringValue,
+				db.numaAttributeForm,
+				db.deviceAttributeMachineModifiers...,
+			)
+			if err != nil {
+				klog.V(4).Infof("Not publishing standardized NUMA attribute for PCI device %s: %v", normalizedAddr, err)
```
gauravkghildiyal, line 844: Should we log this as Error instead?

(Matches the other standard attribute `GetPCIeRootAttributeByPCIBusID` handling)

pkg/inventory/db_test.go
```diff
@@ -653,6 +656,93 @@ func TestAddPCIAttributes(t *testing.T) {
 		db := &DB{}
 		_ = db.addPCIAttributes(devices, nil)
 	})
+
+	t.Run("publishes standardized scalar NUMA attribute by default", func(t *testing.T) {
```
gauravkghildiyal, line 691: The tests are helpful. Thanks for adding them!

gauravkghildiyal (commented):
Thanks @killianmuldoon. Took a closer look, mostly looking good.

One request I had was if we could be a slightly more detailed with the release notes. For example, let's include the full attribute name of the attribute being added. Also, there is this bit of nuance with the standard NUMA attribute that I remember discussing which we should be linking to https://www.kubernetes.dev/resources/keps/6072/#consumer-expectations.

pkg/inventory/db.go
killianmuldoon: I'm not so sure here - the error could be very noisy if there's devices without NUMA information. That could potentially happen on some VM setups. That said I'm happy either way here.

pkg/inventory/db.go
killianmuldoon: Moved it now

pkg/driver/dra_hooks.go
aojea: @killianmuldoon please add a TODO, we already report health so implementing this sounds reasonable if helps reliability ... although I want to understand better this feature , how it works and what problems solve

aojea:
> This feature is enabled with a feature gate on Kubernetes 1.37, and the upstream helper libraries are only available in 1.37 tags.

@gauravkghildiyal any special reason to feature gate this option? is a read only attribute from a kubernetes well defined value so it feels we should expose it once we move to 1.37

pkg/driver/dra_hooks.go
fmuyassarov: WatchHealthStatus is about device health that can be propagated all the way to the workload pod's status `containerStatus.AllocatedResourcesStatus` if the device claimed by the the pod becomes unhealthy. It is sort of alternative (if we can say so) to the events which is general but this is specifically for the DRA drivers. 

xref: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/4680-add-resource-health-to-pod-status/README.md

aojea: we definitively should have this , but also put some user stories on how we want to implement it , opened https://github.com/kubernetes-sigs/dranet/issues/293 for tracking, just add TODO and reference that issue please

killianmuldoon: I added a TODO with a link to the issue - let me know if you'd prefer a different format or wording.

pkg/inventory/db.go
gauravkghildiyal: You're right. Thinking about this more, GetPCIeRootAttributeByPCIBusID has a less chance of failing compared to NUMA (just because a PCI Device is more than likely to have a PCIe root, but not necessary have NUMA information populated). Sorry for asking you to change it back to the original.

gauravkghildiyal:
> > This feature is enabled with a feature gate on Kubernetes 1.37, and the upstream helper libraries are only available in 1.37 tags.
> 
> @gauravkghildiyal any special reason to feature gate this option? is a read only attribute from a kubernetes well defined value so it feels we should expose it once we move to 1.37

This is slightly confusing and non-ideal from upstream itself, but the idea is that the attribute `resource.kubernetes.io/numaNode` can both be a scalar value (e.g. `0`) or a list value (e.g. `[0, 1, 2, 3]`). 

Defining a "list" type attribute is still alpha in DRA. So what this change is feature gating is NOT whether to publish the attribute, but whether to publish it as a scalar or a list. If you think when we change this from scalar to list then it would be a breaking change, you are right, but that's the recommendation https://www.kubernetes.dev/resources/keps/6072/#consumer-expectations without an alternative option.

gauravkghildiyal:
@killianmuldoon One last thing and we should be good: https://github.com/kubernetes-sigs/dranet/pull/283#discussion_r3788251099

pkg/inventory/db.go
killianmuldoon: No issue changing it back - set it to Info now!

PR #287, opened 2026-08-12
Download bpftool to a file before extracting in e2e tests

OguzPastirmaci:
#### What type of PR is this?
/kind failing-test

The e2e tests downloaded bpftool anonymously inside the cluster in two places: on the kind node and in the test pod. These downloads fail when the GitHub release CDN returns errors to shared runner addresses, which failed the bpf tests (for example [this run](https://github.com/kubernetes-sigs/dranet/actions/runs/31631649160/job/94231624860)). An earlier version of this PR kept the in-cluster downloads and added retries, and CI [failed the same way](https://github.com/kubernetes-sigs/dranet/actions/runs/31634819312/job/94242344393).

This version removes the in-cluster downloads. The bats workflow downloads bpftool once on the runner with an authenticated `gh release download`, with retries and `--clobber`. The tests copy the binary into the kind node with `docker cp` and into the test pod with `kubectl cp`. Local runs without a token keep a curl fallback. Verified green on a fork run of the bats workflow.

None

```release-note
NONE
```

### DIFF
--- a/.github/workflows/bats.yml
+++ b/.github/workflows/bats.yml
@@ -57,6 +57,26 @@ jobs:
           # Clean
           sudo rm -rf "${TMP_DIR}"
 
+      - name: Fetch bpftool
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          # Authenticated download: anonymous release fetches from shared
+          # runner addresses are unreliable and have failed the bpf tests.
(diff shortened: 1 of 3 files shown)


gauravkghildiyal:
Thanks @OguzPastirmaci 

/lgtm
/approve

PR #288, opened 2026-08-14
Add PCI mocking capability to exercise PCI in E2E

### LINKED ISSUE #68: Add test support for PCI-E devices
Currently all the tests we have run with dummy linux devices. This is good but does not test the PCI-E logic, such as reading hardware address and other such logic. Some ideas are to emulate a PCI device and make it so we can configure it with a dummy driver. However we do it, we should make it so the device can configure its name, hardware address and other parts that DRANet directly accesses.

michaelasp:
Adds PCI support for testing for network devices in E2E tests.

Fixes #68

```release-note
Add test support for PCI devices
```

### DIFF
--- a/cmd/dranet/app.go
+++ b/cmd/dranet/app.go
@@ -68,6 +68,7 @@ var (
 	profileProvider   string
 	webhookURL        string
 	featureGates      string
+	sysfsRoot         string
 
 	kubeletRootDir string
 
@@ -89,6 +90,7 @@ func init() {
 	flag.StringVar(&webhookURL, "webhook-url", "", "URL for the webhook provider (required if using webhook for either provider)")
 	flag.StringVar(&kubeletRootDir, "kubelet-root-dir", "/var/lib/kubelet", "The kubelet data directory (its --root-dir). The driver's registration socket lives under <dir>/plugins_registry and its dra.sock under <dir>/plugins/<driver-name>. Set this to match the kubelet --root-dir on clusters that relocate it.")
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
+	flag.StringVar(&sysfsRoot, "dranet-experimental-sysfs-root", "/sys", "The path to the sysfs root directory. Set to a custom path during testing or device simulation.")
 
 	flag.Usage = func() {
 		fmt.Fprint(os.Stderr, "Usage: dranet [options]\n\n")
@@ -197,7 +199,13 @@ func main() {
 		klog.Fatalf("failed to setup providers: %v", err)
 	}
 
+	// DRANET_EXPERIMENTAL_SYSFS_ROOT The path to the sysfs root directory. Set to a custom path during testing or device simulation.
+	if envRoot := os.Getenv("DRANET_EXPERIMENTAL_SYSFS_ROOT"); envRoot != "" && sysfsRoot == "/sys" {
+		sysfsRoot = envRoot
+	}
+
 	optsDb := []inventory.Option{
+		inventory.WithSysfsRoot(sysfsRoot),
 		inventory.WithRateLimiter(rate.NewLimiter(rate.Every(minPollInterval), pollBurst)),
 		inventory.WithMaxPollInterval(maxPollInterval),
 		inventory.WithMoveIBInterfaces(moveIBInterfaces),
--- a/Makefile
+++ b/Makefile
@@ -27,13 +27,18 @@ build-dranet:
 build-dranetctl:
 	go build -v -o "$(OUT_DIR)/dranetctl" ./cmd/dranetctl
 
+build-test-tools: build-mockpci
+
+build-mockpci:
+	go build -v -o "$(OUT_DIR)/mockpci" ./cmd/mockpci
+
 clean:
 	rm -rf "$(OUT_DIR)/"
 
 test:
 	CGO_ENABLED=1 go test -v -race -count 1 ./...
 
-e2e-test:
+e2e-test: build-test-tools
 	bats --verbose-run tests/
 
 # code linters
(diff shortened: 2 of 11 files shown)


aojea:
This is great and very needed, however, I'm worried about mounting a tmpfs directly over /sys/class/net and /sys/bus/pci/devices on the host.

In a Kind cluster we dynamically creates veth and dummy interfaces for pods. If we freeze /sys with a tmpfs overlay, any interfaces created after the mount will be completely hidden from the rest of the node, which may break something.

I can see that most projects add a configurable SYSFS_ROOT env var to the plugin instead of mutating the host. However, this brings up another problem: if we override sysfs to a fake directory, how does DRANET discover both the fake PCI devices and the real virtual/hardware interfaces on the node?

cc @byako — we'd love to get your thoughts! I see the [Intel device-faker](https://github.com/intel/intel-resource-drivers-for-kubernetes/tree/main/cmd/device-faker) generates an isolated fake sysfs. How do you handle cases where the plugin needs to see both real devices and fake devices at the same time during E2E tests?

byako:
@aojea I'm flattered, thank you!

Short answer is - we don't, but at least from the top of my head I think it's possible to combine fake and real devices, as long as you don't expect hot-pluggable _devfs_ devices to show up inside the container without restarting it. For sysfs that is, it should be possible, and by having both available instead of mounting tmpfs over real sysfs subdir.

I would do it by expanding the needed bits and pieces in the fake sysfs (e.g. /sysfs) root with symlinks to the real sysfs (which is anyway available at /sys). This is far from simple, and will require watching real sysfs (/sys) with fsnotify, or smth like it, for changes and reflecting them with 3-5 symlinks (or removing them) for every new interface added (or deleted).

We already do similar setup to tes binding management in device-faker [for functional testing](https://github.com/intel/intel-resource-drivers-for-kubernetes/blob/1da939faa33a90dd641d0327cdcdcbbfaa1acdb4/pkg/fakesysfs/gpu-driver-binder.go#L35), and making this possible in a side-car device-faker container for e2e shouldn't be complicated.

For GPU, for example, I would need 
```shell
/sysfs/devices/<PCIeROOT>/<host-pci-device> -> /sys/devices/<PCIeROOT>/devices/<host-pci-device>
/sysfs/bus/pci/devices/<host-pci-device> -> /sysfs/devices/<PCIeROOT>/<host-pci-device>
/sysfs/class/drm/cardX -> /sysfs/bus/pci/devices/<host-pci-device>/drm/cardX
/sysfs/buc/pci/drivers/<GPU-driver>/<host-pci-device> -> /sysfs/devices/<PCIeROOT>/<host-pci-device>
```
that makes it 4 symlinks, not too many to be honest, and only one actually points to outside of fake sysfs.

This approach, of course, would require some pre-validation of what needs to be faked to not overlap with what's physically / virtually available in real sysfs.

install.yaml
```diff
@@ -153,6 +153,9 @@ spec:
         - name: bpf-programs
           mountPath: /sys/fs/bpf
           mountPropagation: HostToContainer
+        - name: sys
```
aojea: is this safe or are we going to get hit by security folks?

michaelasp: ~~Yeah I don't think we need this at all, it's a remnant of a different approach I was trying. Let me remove and see if tests pass.~~

Nvm, we need this for testing due to the tmpfs. Moved to only e2e.

michaelasp:
I updated this to pass the sysfs root point and mount both fake devices and real ones simultaneously. This should allow for both to exist simultaneously without directly mounting over the old /sys of the host. Hopefully that is a better approach, though does require some actual code change which was what I was mostly trying to avoid with the other approach. I think this makes sense though in order to let us add fake devices alongside real ones.

install.yaml
michaelasp: Removed entirely with the new approach of an overlay sysfs.

cmd/dranet/app.go
```diff
@@ -89,6 +90,7 @@ func init() {
 	flag.StringVar(&webhookURL, "webhook-url", "", "URL for the webhook provider (required if using webhook for either provider)")
 	flag.StringVar(&kubeletRootDir, "kubelet-root-dir", "/var/lib/kubelet", "The kubelet data directory (its --root-dir). The driver's registration socket lives under <dir>/plugins_registry and its dra.sock under <dir>/plugins/<driver-name>. Set this to match the kubelet --root-dir on clusters that relocate it.")
 	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
+	flag.StringVar(&sysfsRoot, "sysfs-root", "/sys", "The path to the sysfs root directory. Set to a custom path during testing or device simulation.")
```
aojea: this is my struggle, making this a flag becomes an user faced api, making this a magic env var "DRANET_EXPERIMENTAL_SYSFS_ROOT" removes any guarantee to users (learn this from @BenTheElder in kind https://github.com/kubernetes-sigs/cloud-provider-kind/issues/404

cmd/dranet/app.go
```diff
@@ -197,7 +199,12 @@ func main() {
 		klog.Fatalf("failed to setup providers: %v", err)
 	}
 
+	if envRoot := os.Getenv("SYSFS_ROOT"); envRoot != "" && sysfsRoot == "/sys" {
```
aojea:
```suggestion
    // DRANET_EXPERIMENTAL_SYSFS_ROOT The path to the sysfs root directory. Set to a custom path during testing or device simulation.
	if envRoot := os.Getenv("DRANET_EXPERIMENTAL_SYSFS_ROOT"); envRoot != "" {
```

cmd/dranet/app.go
michaelasp: Yeah while this is mostly just for us to add the overlay it still adds a magic var for testing. I was discussing this with Ben as well, curious what he thinks here :)

Anything that we add will eventually become part of the api contract unfortunately. Good to point this out.

cmd/dranet/app.go
michaelasp: Done, also update the flag to the same name.

PR #290, opened 2026-08-16
driver: sort a copy of the routes instead of the shared slice

thc1006:
#### What type of PR is this?

/kind bug

While reading the route setup I noticed that `applyRoutingConfig` sorts its `routeConfig` argument in place. That slice isn't a private copy: it comes from the pod config store, and `GetPodConfig` returns the `PodConfig` by value, so the `Routes` slice header is copied while its backing array still points at the stored config. Sorting it in place reorders the routes the store keeps for that pod, and because the sort runs after the store's read lock has been released, it can also race with another goroutine reading the same pod's config.

The order the routes are applied in still matters (link-local routes have to go in before the universe routes that depend on them), so I kept the sort but now sort a clone of the slice instead of the one shared with the store, leaving the caller's slice untouched. The routes are only reordered and never modified, so a shallow `slices.Clone` is enough here.

None filed; this is a small follow-up to the route work in #254 and #256.

`Test_applyRoutingConfig` exercises the full path in a temporary network namespace: the interface uses a /32 address so the gateway is not on-link, and the universe route only installs once the link-local route to the gateway has been added, so the test fails if the scope ordering is dropped. It also checks that the input slice keeps its original order, which catches an in-place sort of the shared config.

```release-note
NONE
```

### DIFF
--- a/pkg/driver/netnamespace.go
+++ b/pkg/driver/netnamespace.go
@@ -63,13 +63,16 @@ func applyRoutingConfig(containerNsPAth string, ifName string, routeConfig []api
 	// # ip route show dev eth0
 	//   10.0.5.0/24 via 10.0.5.1 proto dhcp src 10.0.5.8
 	//   10.0.5.1 proto dhcp scope link src 10.0.5.8
-	slices.SortFunc(routeConfig, func(a, b apis.RouteConfig) int {
-		// Routes with scope RT_SCOPE_LINK (253) should come before RT_SCOPE_UNIVERSE (0)
+	// Sort a clone: routeConfig aliases the pod's stored slice, so sorting it in
+	// place would reorder the stored routes and race with their readers.
+	sortedRoutes := slices.Clone(routeConfig)
+	slices.SortFunc(sortedRoutes, func(a, b apis.RouteConfig) int {
+		// Routes with scope RT_SCOPE_LINK (253) should come before RT_SCOPE_UNIVERSE (0).
 		// A higher scope value means it's processed earlier.
 		return int(b.Scope) - int(a.Scope)
 	})
 
-	for _, route := range routeConfig {
+	for _, route := range sortedRoutes {
 		table := route.Table
 		// If VRF is enabled (vrfTable > 0), all routes for this interface
 		// must go into the VRF table to be reachable via the VRF device.
--- a/pkg/driver/netnamespace_test.go
+++ b/pkg/driver/netnamespace_test.go
@@ -17,9 +17,126 @@ limitations under the License.
 package driver
 
 import (
+	"crypto/rand"
+	"fmt"
+	"os"
+	"path"
+	"runtime"
 	"testing"
+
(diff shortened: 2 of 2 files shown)


pkg/driver/netnamespace.go
```diff
@@ -55,21 +55,9 @@ func applyRoutingConfig(containerNsPAth string, ifName string, routeConfig []api
...
-	//   inet 10.0.5.8/32 scope global dynamic eth0
-	// # ip route show dev eth0
-	//   10.0.5.0/24 via 10.0.5.1 proto dhcp src 10.0.5.8
-	//   10.0.5.1 proto dhcp scope link src 10.0.5.8
-	slices.SortFunc(routeConfig, func(a, b apis.RouteConfig) int {
-		// Routes with scope RT_SCOPE_LINK (253) should come before RT_SCOPE_UNIVERSE (0)
-		// A higher scope value means it's processed earlier.
-		return int(b.Scope) - int(a.Scope)
-	})
+	sortedRoutes := sortRoutesByScope(routeConfig)
```
aojea: keep it simpler, just add the clone and sort and do not create the additional helper,

aojea:
just add the test and clone and sort without adding the additional helper

thc1006:
> just add the test and clone and sort without adding the additional helper

Roger!!!

pkg/driver/netnamespace.go
thc1006: Thank you, I dropped the helper and inlined the clone and sort as you suggested.

aojea: I think you didn't push 😅

thc1006: Sorry!!! that push didn't include the change; it's in now (f6d5d6a) with the helper removed and the clone and sort inlined.

now OK!!!

pkg/driver/netnamespace.go
```diff
@@ -55,21 +55,15 @@ func applyRoutingConfig(containerNsPAth string, ifName string, routeConfig []api
...
-	// This is important because universe routes might depend on link-local ones.
-	// For example, in GCE VMs:
-	// # ip addr show eth0
-	//   inet 10.0.5.8/32 scope global dynamic eth0
-	// # ip route show dev eth0
-	//   10.0.5.0/24 via 10.0.5.1 proto dhcp src 10.0.5.8
-	//   10.0.5.1 proto dhcp scope link src 10.0.5.8
-	slices.SortFunc(routeConfig, func(a, b apis.RouteConfig) int {
-		// Routes with scope RT_SCOPE_LINK (253) should come before RT_SCOPE_UNIVERSE (0)
-		// A higher scope value means it's processed earlier.
```
aojea, line 68: why don't keep the comment, it shows the reason why, can you add it again?

thc1006: Thank you so much, you're right that the GCE example explains the reason far better than my shorter version, I've added it back in a792273.

PR #300, opened 2026-08-25
inventory: cap pci device attribute length

aggarwal0009:
#### What type of PR is this?

/kind bug

DRANET publishes the PCI product name as the dra.net/pciDevice attribute. Some PCI device names obtained from pci.ids can exceed Kubernetes' device attribute string length limit (resourceapi.DeviceAttributeMaxValueLength). For example, Mellanox ConnectX-3 Pro VFs (15b3:1004) expose a 67-byte product name, causing the API server to reject the entire ResourceSlice with a validation error and preventing resources from being advertised on the affected node. This PR truncates PCI device names to the Kubernetes-defined maximum length before publishing them as the dra.net/pciDevice attribute, preventing ResourceSlice validation failures while preserving the existing attribute semantics.

N/A

This issue was observed on a ConnectX-3 Pro VF (`15b3:1004`) where the PCI device description is 67 bytes long. Kubernetes rejects the entire ResourceSlice when the attribute exceeds the device attribute length limit. This change mirrors existing handling in DRANET for other attributes that may exceed `resourceapi.DeviceAttributeMaxValueLength`.

```release-note
Prevent ResourceSlice creation failures caused by PCI device names that exceed Kubernetes device attribute length limits by truncating the published `dra.net/pciDevice` attribute value.
```

### DIFF
--- a/pkg/inventory/db.go
+++ b/pkg/inventory/db.go
@@ -868,7 +868,22 @@ func (db *DB) addPCIAttributes(devices []resourceapi.Device, pciInfo *ghw.PCIInf
 			devices[i].Attributes[apis.AttrPCIVendor] = resourceapi.DeviceAttribute{StringValue: &pciDev.Vendor.Name}
 		}
 		if _, hasAttr := devices[i].Attributes[apis.AttrPCIDevice]; !hasAttr && pciDev.Product != nil {
-			devices[i].Attributes[apis.AttrPCIDevice] = resourceapi.DeviceAttribute{StringValue: &pciDev.Product.Name}
+			productName := pciDev.Product.Name
+
+			if len(productName) > resourceapi.DeviceAttributeMaxValueLength {
+				klog.V(4).Infof(
+					"Truncating PCI device name %q from %d to %d bytes",
+					productName,
+					len(productName),
+					resourceapi.DeviceAttributeMaxValueLength,
+				)
+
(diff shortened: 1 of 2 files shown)


Bowser1704:
Thanks @aggarwal0009!

Besides `pciDevice`, I think`pciVendor` also needs to be considered here.

tamilmani1989 (approved):
lgtm

aojea:
> Besides `pciDevice`, I think`pciVendor` also needs to be considered here.

Yeah, that is true.

Thinking holistically about this, we already do this in other places and my understanding is that we should follow three different strategies depending on the attribute:

- Truncate: Descriptive text (pciDevice, pciVendor names, alias), a prefix is still human-useful and startsWith/contains CEL still works. Exact match CEL breaks though
- Drop: (coreweave fabric IDs already do this) — a truncated ID can falsely prefix-match a different device, which is worse than not having anything
- Normalize / Encode / Truncate + hash suffix: we do not do this and I fail to see a good example for this

Then is the question, do we normalize at build time or at publish time, right now we are doing at build time that helps us to understand the reason of the normalization, despite more verbose I think that I prefer to keep doing it this way vs building something at publish time , since that will not have the context of the attribute and we can always revisit the decision with the time.

/lgtm
/approve

@aggarwal0009 do you mind following up and fix `pciVendor` too?

Issue #301, opened 2026-08-25
Add support for CoreWeave Kubernetes Service (CKS)


anson627:
## Summary

Add and validate support for running DRANET on CoreWeave Kubernetes Service (CKS).

## Motivation

CKS is used for GPU-intensive and distributed AI/ML workloads that can benefit from DRANET high-performance networking and topology-aware allocation of GPUs and NICs. DRANET currently documents KIND and GKE setup, and its cloud-provider hints do not include CoreWeave/CKS.

## Proposed scope

- Determine the CKS prerequisites for Kubernetes Dynamic Resource Allocation (DRA), NRI, and the container runtime.
- Identify whether CKS needs a native cloud-provider implementation, can use the webhook provider, or works with provider discovery disabled.
- Validate network-device discovery, resource publication, scheduling, and attachment on CKS nodes.
- Add installation and configuration documentation for CKS, including any required feature gates and Helm values.
- Add a reproducible smoke test or end-to-end validation where practical.

## Acceptance criteria

- DRANET can be installed successfully on a supported CKS cluster.
- A workload can request a network device through DRA and receive the expected interface configuration.
- Provider-specific limitations and configuration are documented.
- The supported CKS and Kubernetes versions are documented.

PR #241, opened 2026-06-26
feat: implement IPVLAN Subinterface configuration and provision

ngcxy:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
This PR introduces support for subinterface provisioning in the DRANET driver, currently implementing IPVLAN as the subinterface type. The feature addresses the networking requirement for #239 by creating a new subinterface inside the pod network namespace without moving the host’s parent interface. It also provides the groundwork for #63 to later enable multiple-subinterface creation on a single NIC.

The key changes include:

*API*
- Add optional `interface.type` to `NetworkConfig` to select how an allocated device is presented to the Pod.
   - `passthrough` (default) moves the device into the Pod's netns; `ipvlan` keeps the device on the host and creates an IPVLAN subinterface inside the Pod.
- Add `IPVlanConfig` to configure the IPVLAN `mode`/`flag` (currently `l2`/`bridge`).
- Add `interface.addressing` (`static`/`dhcp`/`unnumbered`) to select the IP configuration strategy; `unnumbered` covers deliberately addressless subinterfaces. Deprecate the boolean `interface.dhcp` in favor of `addressing: dhcp` (backward compatible — `dhcp: true` is normalized to `addressing: dhcp` when `addressing` is unset).

*IPAM*
- Add a dedicated `pkg/ipam` package with a concurrency-safe node-local `LocalIPAM` that owns the set of in-use addresses and hands out one address per IP family; allocated IPs are released on claim teardown.
- Resolve subinterface addresses from the cloud provider's profile via `GetProfileConfig`, which calls `LocalIPAM`.
- Move pod config store construction into `app.go` (injected as a required `Start()` parameter) and seed its in-use subinterface IPs into the cloud provider via `cloudprovider.InstanceOptions.ReservedAddresses`, so IPAM state is recovered on restart and in-use addresses are not handed out twice.

*Driver*
- During ResourceClaim preparation, assign the subinterface name, resolve its address(es) from `interface.addresses` or a cloud profile, and configure source-based routing.
   - User-configured routes/rules are preserved — automatic source-based routing is skipped when the user already specified them.
   - Fail the claim when a subinterface resolves with no address (and `addressing: unnumbered` is not set) instead of silently bringing up a dead link that still reports Ready.
- Reject subinterfaces on devices in exclusive RDMA mode (shared RDMA mode is required).
- Integrate subinterface create/delete (IPVLAN) inside the Pod network namespace via the NRI run/stop sandbox hooks.

*GCE*
[...]

PR #302, opened 2026-08-25
Add CoreWeave CKS cloud provider support

anson627:
Part of #301.

Adds native CKS discovery through Kubernetes Node labels and publishes InfiniBand fabric, superpod, leafgroup, rack, speed, instance type, and per-device leaf-switch attributes. Also adds the `CKS` provider hint to the binary and Helm chart.

Verification:
- `go test ./...`
- `hack/verify-gofmt.sh`
- `helm lint --strict deployments/helm/dranet`
- `helm template dranet deployments/helm/dranet --set args.cloudProviderHint=CKS`
- Read-only validation against CKS `use15` confirmed the implemented fabric and `neighbors.expected.ibpN.device` label contract.

```release-note
Add CoreWeave Kubernetes Service (CKS) cloud-provider discovery and InfiniBand topology attributes.
```

### DIFF
--- a/cmd/dranet/app.go
+++ b/cmd/dranet/app.go
@@ -192,11 +192,16 @@ func main() {
 		}
 		opts = append(opts, driver.WithFilter(prg))
 	}
-	providerDependencies := discovery.Dependencies{
-		Nodes:    clientset.CoreV1().Nodes(),
-		NodeName: nodeName,
+	providerOpts := providerOptions{
+		cloudProviderHint: cloudProviderHint,
+		profileProvider:   profileProvider,
+		webhookURL:        webhookURL,
+		dependencies: discovery.Dependencies{
+			NodeClient: clientset.CoreV1().Nodes(),
+			NodeName:   nodeName,
+		},
 	}
-	cloudInst, profProv, err := setupProviders(ctx, cloudProviderHint, profileProvider, webhookURL, providerDependencies)
+	cloudInst, profProv, err := setupProviders(ctx, providerOpts)
 	if err != nil {
 		klog.Fatalf("failed to setup providers: %v", err)
 	}
@@ -253,28 +258,35 @@ func printVersion() {
 	klog.Infof("dranet go %s build: %s time: %s", info.GoVersion, vcsRevision, vcsTime)
 }
 
-func setupProviders(ctx context.Context, cloudProviderHint string, profileProvider string, webhookURL string, dependencies discovery.Dependencies) (cloudprovider.CloudInstance, cloudprovider.ProfileProvider, error) {
+type providerOptions struct {
+	cloudProviderHint string
+	profileProvider   string
+	webhookURL        string
+	dependencies      discovery.Dependencies
+}
+
+func setupProviders(ctx context.Context, opts providerOptions) (cloudprovider.CloudInstance, cloudprovider.ProfileProvider, error) {
 	var cloudInst cloudprovider.CloudInstance
 	var profProv cloudprovider.ProfileProvider
 	var err error
 
 	var hint discovery.CloudProviderHint
 	// Auto-discover cloud provider if not explicitly set
-	if cloudProviderHint == "" {
-		hint = discovery.DiscoverCloudProviderWithDependencies(ctx, webhookURL, dependencies)
+	if opts.cloudProviderHint == "" {
+		hint = discovery.DiscoverCloudProviderWithDependencies(ctx, opts.webhookURL, opts.dependencies)
 	} else {
-		hint = discovery.CloudProviderHint(cloudProviderHint)
+		hint = discovery.CloudProviderHint(opts.cloudProviderHint)
 	}
 
 	// Setup the Underlay (Hardware Discovery / Cloud Instance Info)
-	cloudInst, err = discovery.GetInstancePropertiesWithDependencies(ctx, hint, webhookURL, dependencies)
+	cloudInst, err = discovery.GetInstancePropertiesWithDependencies(ctx, hint, opts.webhookURL, opts.dependencies)
 	if err != nil {
 		klog.Infof("failed to initialize cloud provider %q: %v", hint, err)
 		cloudInst = nil
 	}
 
 	// Setup the Overlay (Profile Provider / User Intent)
-	switch profileProvider {
+	switch opts.profileProvider {
 	case "cloud":
 		if p, ok := cloudInst.(cloudprovider.ProfileProvider); ok {
 			profProv = p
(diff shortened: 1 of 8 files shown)


cmd/dranet/app.go
```diff
@@ -192,7 +192,11 @@ func main() {
 		}
 		opts = append(opts, driver.WithFilter(prg))
 	}
-	cloudInst, profProv, err := setupProviders(ctx, cloudProviderHint, profileProvider, webhookURL)
+	providerDependencies := discovery.Dependencies{
+		Nodes:    clientset.CoreV1().Nodes(),
```
gauravkghildiyal: nit: Consider naming this `NodeClient`

pkg/cloudprovider/discovery/discovery.go
```diff
@@ -38,12 +40,31 @@ const (
...
+
 // DiscoverCloudProvider probes the environment to detect which cloud provider DRANET is running on.
 func DiscoverCloudProvider(ctx context.Context, webhookURL string) CloudProviderHint {
+	return DiscoverCloudProviderWithDependencies(ctx, webhookURL, Dependencies{})
+}
+
+// DiscoverCloudProviderWithDependencies probes the environment using additional
+// Kubernetes-local provider inputs when available.
+func DiscoverCloudProviderWithDependencies(ctx context.Context, webhookURL string, dependencies Dependencies) CloudProviderHint {
+	// CKS is checked first because its authoritative metadata is already in the
```
gauravkghildiyal: In the sense that fetching the Kubernetes Node object also requires making an API call, I don't think this case is any different than the rest of the cases so we can append this later.

That said, this auto detection system is clearly suboptimal for whoever is last in the list. I'd highly recommend explicitly using the `--cloud-provider-hint` flag whenever we are documenting this.

 (Separate from this and needn't be changed in this PR, but I think it's fair to sort these alphabetically)

cmd/dranet/app.go
```diff
@@ -249,21 +253,21 @@ func printVersion() {
 	klog.Infof("dranet go %s build: %s time: %s", info.GoVersion, vcsRevision, vcsTime)
 }
 
-func setupProviders(ctx context.Context, cloudProviderHint string, profileProvider string, webhookURL string) (cloudprovider.CloudInstance, cloudprovider.ProfileProvider, error) {
+func setupProviders(ctx context.Context, cloudProviderHint string, profileProvider string, webhookURL string, dependencies discovery.Dependencies) (cloudprovider.CloudInstance, cloudprovider.ProfileProvider, error) {
```
gauravkghildiyal: You'll have some merge conflicts with https://github.com/kubernetes-sigs/dranet/pull/241 which is also adding things similar to this additional dependency.

Please be on the lookout for the final state after it merges (if it merges before)

gauravkghildiyal (commented):
Excited to see a Coreweave provider. Thanks @anson627

cmd/dranet/app.go
aojea: @gauravkghildiyal 
My latest comment  on #241 added each new parameters as a new paramater of this function that does not scale too ... let's get an agreement on this extension mechanism to unblock both

cmd/dranet/app.go
anson627: Good suggestion. I’ll rename Dependencies.Nodes to NodeClient and update the call sites and tests

pkg/cloudprovider/discovery/discovery.go
anson627: Makes sense. I’ll move the CKS probe after the existing native provider probes instead of special-casing it first. The CKS example in #303 explicitly uses `--cloud-provider-hint=CKS`, and I’ll keep that as the documented path. I’ll leave the broader alphabetical reordering for a separate cleanup.

cmd/dranet/app.go
anson627: Agreed. My preference is a single dependencies/options struct passed to setupProviders, rather than adding one positional parameter per provider. That keeps the signature stable as integrations add optional `dependencies.discovery.Dependencies` was intended as that extension point, CKS currently needs only NodeClient and NodeName. I’m happy to align its name and location with the approach agreed in #241 and rebase once that shape settles

aojea:
failure looks legit 

> 2026-09-01T16:53:55.9357743Z === RUN   TestDiscoverCloudProviderDetectsCKS
2026-09-01T16:53:55.9358912Z I0901 16:53:53.955822   16638 aws.go:168] could not reach IMDS for EC2 detection: operation error ec2imds: GetInstanceIdentityDocument, failed to get API token, operation error ec2imds: getToken, http response error StatusCode: 400, request to EC2 IMDS failed
2026-09-01T16:53:55.9360232Z     discovery_test.go:42: DiscoverCloudProviderWithDependencies() = "AZURE", want "CKS"
2026-09-01T16:53:55.9360666Z --- FAIL: TestDiscoverCloudProviderDetectsCKS (0.01s)

anson627:
CI is fixed now

pkg/cloudprovider/discovery/discovery.go
```diff
@@ -38,35 +40,67 @@ const (
...
 }
 
 // GetInstanceProperties initializes and returns the specified cloud provider instance.
 func GetInstanceProperties(ctx context.Context, hint CloudProviderHint, webhookURL string) (cloudprovider.CloudInstance, error) {
+	return GetInstancePropertiesWithDependencies(ctx, hint, webhookURL, Dependencies{})
+}
+
+// GetInstancePropertiesWithDependencies initializes the specified cloud provider
+// using additional Kubernetes-local provider inputs when available.
+func GetInstancePropertiesWithDependencies(ctx context.Context, hint CloudProviderHint, webhookURL string, dependencies Dependencies) (cloudprovider.CloudInstance, error) {
```
gauravkghildiyal: I'm thinking that we don't need two versions of functions like `GetInstanceProperties/GetInstancePropertiesWithDependencies` and `DiscoverCloudProvider/DiscoverCloudProviderWithDependencies`. Let's update the existing function signatures to use the new Dependency function params?

gauravkghildiyal (commented):
Looking good, just one last cleanup request unless I misunderstood the need.

cmd/dranet/app.go
gauravkghildiyal: This sounds reasonable and should work for the other PR too. Let's merge this one first so the other one can rebase.

pkg/cloudprovider/discovery/discovery.go
anson627: Done, thanks!

gauravkghildiyal:
Awesome, thanks for the prompt resolution.

/lgtm
/approve

gauravkghildiyal:
@anson627 Should we add a release note for this? (If yes, can you update the PR description)

anson627:
> @anson627 Should we add a release note for this? (If yes, can you update the PR description)

Done, thanks!

PR #305, opened 2026-08-27
build(deps): bump Kubernetes to 1.37.0

killianmuldoon:
#### What type of PR is this?

/kind cleanup

Bump Kubernetes libraries to 1.37.0. This is a follow up to #283 where the libraries were bumped to rc.0.

```release-note
Bump Kubernetes libraries to 1.37.0
```


fmuyassarov:
LGTM
I wonder if we should bump the k8s to 1.37 in other places too, example
https://github.com/kubernetes-sigs/dranet/blob/6a1f6506560da0ba365974bfcc4bd7a76d94c3d0/.github/workflows/bats.yml#L24
or
https://github.com/kubernetes-sigs/dranet/blob/6a1f6506560da0ba365974bfcc4bd7a76d94c3d0/.github/workflows/e2e.yml#L28

PR #313, opened 2026-09-03
inventory: cap pci vendor attribute length

aggarwal0009:
#### What type of PR is this?
/kind bug

Some PCI vendor names reported by ghw/pcidb can exceed the DRA API's per-attribute value length limit, causing the ResourceSlice to be rejected with a validation error like Too long: may not be more than 64 bytes. This previously only surfaced for the PCI device (product) name; the vendor name has the same class of bug.

N/A

Refer this PR --> https://github.com/kubernetes-sigs/dranet/pull/300
This change was requested by [aojea](https://github.com/aojea)

```release-note
Prevent ResourceSlice creation failures caused by PCI vendor names that exceed Kubernetes device attribute length limits by truncating the published `dra.net/pciVendor` attribute value.
```


aojea:
/lgtm
/approve

Thanks for following up

PR #314, opened 2026-09-03
driver: version the checkpoint database schema

aojea:
Add a schema version to the bbolt checkpoint database so future format changes can be migrated safely.

The database is stamped with `meta/schemaVersion`; existing databases without it are treated as version 1. Migrations run one version at a time in the same transaction as the version stamp, so an interrupted migration rolls back cleanly. A database newer than the running daemon is refused before any mutation, since checkpointed state (e.g. DHCP leases) cannot be rebuilt from the system.

Tests pin the on-disk wire format and cover upgrade, rollback and recovery scenarios, and DEVELOPER.md documents the behavior and when to bump the version.

```release-note
Implement database versioning for local storage
```

### DIFF
--- a/pkg/driver/pod_device_config_bolt.go
+++ b/pkg/driver/pod_device_config_bolt.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	bolt "go.etcd.io/bbolt"
@@ -30,6 +31,8 @@ import (
 
 // Bucket layout:
 //
+//	meta (root bucket)
+//	  └── schemaVersion = decimal string, see checkpointSchemaVersion
 //	pod_configs (root bucket)
 //	  └── <POD_UID> (nested bucket per pod)
 //	        └── device_configs (nested bucket for device configs)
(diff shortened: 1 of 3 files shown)


aojea:
Follow up on https://github.com/kubernetes-sigs/dranet/pull/304#discussion_r3901499686

pkg/driver/pod_device_config_bolt.go
```diff
@@ -58,19 +75,54 @@ func newBoltCheckpointer(path string) (*boltCheckpointer, error) {
...
 	}
 
 	return &boltCheckpointer{db: db}, nil
 }
 
+// migrateSchema upgrades the checkpoint layout to target one version at a
+// time and stamps the version, all inside the caller's transaction so a
+// failure rolls back to the previous layout untouched. A database newer than
+// target is refused: checkpointed state such as DHCP leases cannot be rebuilt
+// from the system, so failing loudly beats silently dropping it.
```
rbtr: I find these new comments hard to parse; because this behavior is operationally important it is worth making it very clear that a schema migration from eg v1->v4 goes 1->2->3->4, but if 3->4 fails, the db is rolled all the way back to 1 by the transaction. And so we won't leave the db on v3 for a previously deployed version of DRANET which only speaks v1.

aojea: fair, let me change it

aojea:
folks don't be shy to review if you feel you have experience on the area

rbtr (approved):
lgtm fwiw

rbtr:
> /assign [@rbtr](https://github.com/rbtr)

@aojea would you sponsor me for kube-sigs?

aojea:
> > /assign [@rbtr](https://github.com/rbtr)
> 
> @aojea would you sponsor me for kube-sigs?

absolutely

gauravkghildiyal:
The tests and documentation are helpful. Thank you!

/lgtm

PR #316, opened 2026-09-03
Fix network configuration merging

### LINKED ISSUE #294: MergeNetworkConfig changes provider data and ignores explicit zero values
Came across this one when testing different ARP settings. `MergeNetworkConfig` has two issues.

First, the result can share pointers and maps with the input configurations. A merge can change cached provider data. Concurrent merges can also cause a data race.

Second, explicit zero values do not always override provider values.

For example:

```text
user arpIgnore: 0
provider arpIgnore: 1
result arpIgnore: 1
```

This also affects values such as:

- `dhcp: false`
- `forwarding: false`
- `arpAnnounce: 0`

## Expected behavior

- The merge must not change either input.
- The result must not share mutable data with either input.
- A non-nil user pointer must override the provider value.
- A nil user pointer must preserve the provider value.
- Existing slice merge behavior must remain unchanged.

## Suggested fix

Deep-copy both inputs before the merge. Then apply each non-nil user pointer explicitly.

The main downside is that each new mutable field must be added manually going forward.

## Note
[#241](https://github.com/kubernetes-sigs/dranet/pull/241) adds new interface fields that also need this behavior.

This change can wait until #241 merges. The implementation can then cover all fields in one change.

OguzPastirmaci:
#### What type of PR is this?

/kind bug

`MergeNetworkConfig` reused pointers, slices, and maps from its inputs. The device config store keeps the provider configuration for later requests. A merge could change this cached configuration through shared pointers. A later merge could then use the changed values. Concurrent merge calls that used the same cached configuration could also cause a data race.

The merge also treated pointer values such as `false` and `0` as empty. Explicit user values such as `dhcp: false` or `arpIgnore: 0` could not override provider values. mergo does this by design (darccio/mergo#200), so this change replaces mergo.

This change:

- Merges every field by name. Strings take the first non-empty value. Pointers take a copy of the user value when set, including `false` and `0`. Slices are appended and deduplicated with the user entry last. Maps are combined with user entries winning.
- Returns a result that shares no memory with either input.
- Applies the same merge rules when either input is nil.
- Removes mergo from go.mod. It has no other user in the repository.
- Adds coverage tests that walk the `NetworkConfig` type with reflection, so a new field is covered without a field list. They check that a value set on one side survives the merge, one field at a time, that a user pointer wins even when it points at a zero value, and that the result shares no pointer, slice, or map with the inputs, including when one input is nil.

Fixes #294

The first commit is the original fix: a deep copy and pointer override lists on top of mergo. The second commit replaces it with the explicit merge, following the review discussion. The tests from the first commit are kept unchanged and pass against the explicit merge. The new coverage tests can replace the list-based ones in a follow-up once reviewers agree they cover the same ground.

Tested on an OKE cluster with a provider profile that sets the type, address, ARP pointers, routes, and rules of an IPvlan child. A claim with `name`, `mtu`, and an explicit `arpIgnore: 0` produced a child with the user name and MTU, `arp_ignore=0` over the profile value, and the `arp_announce`, routes, and rules from the profile.

This change also applies to profile merges. An explicit `false` or `0` device value now overrides the profile value. The precedence order is user configuration, device configuration, and then profile configuration.

```release-note
Fixed network configuration merging. Explicit `false` and `0` user values now override provider values. Merging no longer changes cached provider configuration.
```

### DIFF
--- a/pkg/apis/config_merge.go
+++ b/pkg/apis/config_merge.go
@@ -17,33 +17,33 @@ limitations under the License.
 package apis
 
 import (
+	"maps"
+	"slices"
+
 	"dario.cat/mergo"
+	"k8s.io/utils/ptr"
 )
 
-// MergeNetworkConfig merges a cloud provider configuration into a user configuration.
-// It returns a new *NetworkConfig with the merged result.
-// It follows a strict "User wins" strategy: any scalar setting defined in the user config
-// overrides the cloud provider config. For slices, the two configurations are combined,
-// but duplicates are resolved in favor of the user config.
-// The user parameter is assumed to be non-nil. The cloud parameter can be nil, resulting in
-// a copy of the user parameter.
+// MergeNetworkConfig combines a provider configuration with a user configuration.
+// A nil input is treated as an empty configuration.
+// User values, including explicit pointer zero values, override provider values.
+// Slices are combined, duplicates keep user values, and neither input is changed.
 func MergeNetworkConfig(user, cloud *NetworkConfig) *NetworkConfig {
+	if user == nil {
+		user = &NetworkConfig{}
+	}
 	if cloud == nil {
-		copy := *user
-		return &copy
+		cloud = &NetworkConfig{}
 	}
 
-	merged := &NetworkConfig{}
-
-	// Start with the cloud configuration as the base
-	if err := mergo.Merge(merged, cloud); err != nil {
-		return &NetworkConfig{} // or log error? We'll just return nil or early? Let's just return what we have.
-	}
+	merged := deepCopyNetworkConfig(cloud)
+	userCopy := deepCopyNetworkConfig(user)
 
 	// Merge the user configuration on top, overriding cloud settings and appending slices.
-	if err := mergo.Merge(merged, user, mergo.WithOverride, mergo.WithAppendSlice); err != nil {
+	if err := mergo.Merge(merged, userCopy, mergo.WithOverride, mergo.WithAppendSlice); err != nil {
 		return &NetworkConfig{}
 	}
+	applyUserPointerOverrides(&merged.Interface, &userCopy.Interface)
 
 	// Deduplicate slices where order or uniqueness matters.
 	// For addresses, we just unique them.
@@ -62,6 +62,78 @@ func MergeNetworkConfig(user, cloud *NetworkConfig) *NetworkConfig {
 	}
 
 	return merged
+}
+
+func deepCopyNetworkConfig(config *NetworkConfig) *NetworkConfig {
+	if config == nil {
+		return nil
+	}
+
+	// Keep this copy in sync with mutable fields in types.go.
+	copy := *config
+	copy.Interface.Addresses = slices.Clone(config.Interface.Addresses)
+	copy.Interface.DHCP = clonePointer(config.Interface.DHCP)
+	copy.Interface.MTU = clonePointer(config.Interface.MTU)
+	copy.Interface.HardwareAddr = clonePointer(config.Interface.HardwareAddr)
+	copy.Interface.GSOMaxSize = clonePointer(config.Interface.GSOMaxSize)
+	copy.Interface.GROMaxSize = clonePointer(config.Interface.GROMaxSize)
+	copy.Interface.GSOIPv4MaxSize = clonePointer(config.Interface.GSOIPv4MaxSize)
+	copy.Interface.GROIPv4MaxSize = clonePointer(config.Interface.GROIPv4MaxSize)
+	copy.Interface.DisableEBPFPrograms = clonePointer(config.Interface.DisableEBPFPrograms)
+	copy.Interface.Forwarding = clonePointer(config.Interface.Forwarding)
+	copy.Interface.ARPIgnore = clonePointer(config.Interface.ARPIgnore)
+	copy.Interface.ARPAnnounce = clonePointer(config.Interface.ARPAnnounce)
+	if config.Interface.VRF != nil {
+		vrf := *config.Interface.VRF
+		vrf.Table = clonePointer(config.Interface.VRF.Table)
+		copy.Interface.VRF = &vrf
+	}
+	if config.Interface.IPVlan != nil {
+		ipvlan := *config.Interface.IPVlan
+		copy.Interface.IPVlan = &ipvlan
+	}
+
+	copy.Routes = slices.Clone(config.Routes)
+	copy.Rules = slices.Clone(config.Rules)
+	copy.Neighbors = slices.Clone(config.Neighbors)
+	if config.Ethtool != nil {
+		ethtool := *config.Ethtool
+		ethtool.Features = maps.Clone(config.Ethtool.Features)
+		ethtool.PrivateFlags = maps.Clone(config.Ethtool.PrivateFlags)
+		copy.Ethtool = &ethtool
+	}
+
+	return &copy
+}
+
+func applyUserPointerOverrides(merged, user *InterfaceConfig) {
+	overridePointer(&merged.DHCP, user.DHCP)
+	overridePointer(&merged.MTU, user.MTU)
+	overridePointer(&merged.HardwareAddr, user.HardwareAddr)
+	overridePointer(&merged.GSOMaxSize, user.GSOMaxSize)
+	overridePointer(&merged.GROMaxSize, user.GROMaxSize)
+	overridePointer(&merged.GSOIPv4MaxSize, user.GSOIPv4MaxSize)
+	overridePointer(&merged.GROIPv4MaxSize, user.GROIPv4MaxSize)
+	overridePointer(&merged.DisableEBPFPrograms, user.DisableEBPFPrograms)
+	overridePointer(&merged.Forwarding, user.Forwarding)
+	overridePointer(&merged.ARPIgnore, user.ARPIgnore)
+	overridePointer(&merged.ARPAnnounce, user.ARPAnnounce)
+	if user.VRF != nil && user.VRF.Table != nil {
+		overridePointer(&merged.VRF.Table, user.VRF.Table)
+	}
+}
+
+func clonePointer[T any](value *T) *T {
+	if value == nil {
+		return nil
+	}
+	return ptr.To(*value)
+}
+
+func overridePointer[T any](destination **T, source *T) {
+	if source != nil {
+		*destination = clonePointer(source)
+	}
 }
 
 // deduplicateStrings compacts a slice of strings keeping the last occurrence
--- a/pkg/apis/config_merge_test.go
+++ b/pkg/apis/config_merge_test.go
@@ -17,6 +17,10 @@ limitations under the License.
 package apis
 
 import (
+	"encoding/json"
+	"reflect"
+	"strings"
+	"sync"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -62,6 +66,18 @@ func TestMergeNetworkConfig(t *testing.T) {
 					DHCP: ptr.To(true),
 				},
 			},
+		},
+		{
+			name:  "user profile overrides provider profile",
+			user:  &NetworkConfig{Profile: "user-profile"},
+			cloud: &NetworkConfig{Profile: "provider-profile"},
+			want:  &NetworkConfig{Profile: "user-profile"},
+		},
+		{
+			name:  "empty user profile keeps provider profile",
+			user:  &NetworkConfig{},
+			cloud: &NetworkConfig{Profile: "provider-profile"},
+			want:  &NetworkConfig{Profile: "provider-profile"},
 		},
 		{
 			name: "merge slices without duplicates",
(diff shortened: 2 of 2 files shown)


dkennetzoracle (approved):
Reviewed this offline with @OguzPastirmaci - since this is not solely an OKE specific change, would be good to get another set of eyes on it but the change is fairly contained.

aojea:
@OguzPastirmaci if you are going to keep contributing please think in joining the org ... @dkennetzoracle can sponsor now ... and also add the /ok-to-test tag ;) ... I can be the second sponsor

/ok-to-test

pkg/apis/config_merge.go
```diff
@@ -64,6 +63,80 @@ func MergeNetworkConfig(user, cloud *NetworkConfig) *NetworkConfig {
 	return merged
 }
 
+func deepCopyNetworkConfig(config *NetworkConfig) *NetworkConfig {
```
aojea: I wonder if we can have a way to use reflection to add a test that ensure `deepCopyNetworkConfig` is updated once we add a new field

pkg/apis/config_merge.go
```diff
@@ -64,6 +63,80 @@ func MergeNetworkConfig(user, cloud *NetworkConfig) *NetworkConfig {
...
+	overridePointer(&merged.DisableEBPFPrograms, user.DisableEBPFPrograms)
+	overridePointer(&merged.Forwarding, user.Forwarding)
+	overridePointer(&merged.ARPIgnore, user.ARPIgnore)
+	overridePointer(&merged.ARPAnnounce, user.ARPAnnounce)
+	if user.VRF != nil && user.VRF.Table != nil {
+		overridePointer(&merged.VRF.Table, user.VRF.Table)
+	}
+}
+
+func clonePointer[T any](value *T) *T {
```
aojea: kubernetes has https://pkg.go.dev/k8s.io/utils/ptr just FYI

pkg/apis/config_merge.go
aojea: ok, didn't get to the test when I added the comment

pkg/apis/config_merge.go
dkennetzoracle: nice!

OguzPastirmaci: TIL and nice indeed!

OguzPastirmaci:
> @OguzPastirmaci if you are going to keep contributing please think in joining the org ... @dkennetzoracle can sponsor now ... and also add the /ok-to-test tag ;) ... I can be the second sponsor
> 
> /ok-to-test

@aojea I'll definitely take you and @dkennetzoracle up on that :)

pkg/apis/config_merge.go
```diff
@@ -64,6 +64,78 @@ func MergeNetworkConfig(user, cloud *NetworkConfig) *NetworkConfig {
...
+	overridePointer(&merged.GSOMaxSize, user.GSOMaxSize)
+	overridePointer(&merged.GROMaxSize, user.GROMaxSize)
+	overridePointer(&merged.GSOIPv4MaxSize, user.GSOIPv4MaxSize)
+	overridePointer(&merged.GROIPv4MaxSize, user.GROIPv4MaxSize)
+	overridePointer(&merged.DisableEBPFPrograms, user.DisableEBPFPrograms)
+	overridePointer(&merged.Forwarding, user.Forwarding)
+	overridePointer(&merged.ARPIgnore, user.ARPIgnore)
+	overridePointer(&merged.ARPAnnounce, user.ARPAnnounce)
+	if user.VRF != nil && user.VRF.Table != nil {
+		overridePointer(&merged.VRF.Table, user.VRF.Table)
```
aojea: if `merged.VRF` is nil this panics

aojea:
Reviewing this it seems that we have to take a decision or to drop `mergo` directly and do our own thing or to understand better mergo and handle correctly the merging. @liggitt you always have more experience on this problems between serialization and type merges, is this something you can advice ? I personally don't like the  idea of generating the deepcopy manually once we add a new field, we do that in client go Config though ... but if is the only way

aojea:
Ok, it seems this is by design in mergo https://github.com/darccio/mergo/issues/200 so we need to find an alternative

aojea:
Let's ditch `mergo`, I just asked an AI tool to give me a snippet for reference but this seems the best alternative (code needs to be evaluated)

```go
func MergeNetworkConfig(user, cloud *NetworkConfig) *NetworkConfig {
	if user == nil {
		user = &NetworkConfig{}
	}
	if cloud == nil {
		cloud = &NetworkConfig{}
	}
	merged := &NetworkConfig{
		Profile:   firstNonEmpty(user.Profile, cloud.Profile),
		Interface: mergeInterfaceConfig(&user.Interface, &cloud.Interface),
		Routes:    deduplicateRoutes(append(slices.Clone(cloud.Routes), user.Routes...)),
		Rules:     deduplicateRules(append(slices.Clone(cloud.Rules), user.Rules...)),
		Neighbors: deduplicateNeighbors(append(slices.Clone(cloud.Neighbors), user.Neighbors...)),
		Ethtool:   mergeEthtool(user.Ethtool, cloud.Ethtool),
	}
	if merged.Interface.Type != InterfaceTypeIPVLAN {
		merged.Interface.IPVlan = nil
	}
	return merged
}
func mergeInterfaceConfig(user, cloud *InterfaceConfig) InterfaceConfig {
	return InterfaceConfig{
		Name:                firstNonEmpty(user.Name, cloud.Name),
		Type:                InterfaceType(firstNonEmpty(string(user.Type), string(cloud.Type))),
		Addressing:          AddressingMode(firstNonEmpty(string(user.Addressing), string(cloud.Addressing))),
		Addresses:           deduplicateStrings(append(slices.Clone(cloud.Addresses), user.Addresses...)),
		DHCP:                overridePtr(user.DHCP, cloud.DHCP),
		MTU:                 overridePtr(user.MTU, cloud.MTU),
		HardwareAddr:        overridePtr(user.HardwareAddr, cloud.HardwareAddr),
		GSOMaxSize:          overridePtr(user.GSOMaxSize, cloud.GSOMaxSize),
		GROMaxSize:          overridePtr(user.GROMaxSize, cloud.GROMaxSize),
		GSOIPv4MaxSize:      overridePtr(user.GSOIPv4MaxSize, cloud.GSOIPv4MaxSize),
		GROIPv4MaxSize:      overridePtr(user.GROIPv4MaxSize, cloud.GROIPv4MaxSize),
		DisableEBPFPrograms: overridePtr(user.DisableEBPFPrograms, cloud.DisableEBPFPrograms),
		Forwarding:          overridePtr(user.Forwarding, cloud.Forwarding),
		ARPIgnore:           overridePtr(user.ARPIgnore, cloud.ARPIgnore),
		ARPAnnounce:         overridePtr(user.ARPAnnounce, cloud.ARPAnnounce),
		VRF:                 mergeVRF(user.VRF, cloud.VRF),
		IPVlan:              mergeIPVlan(user.IPVlan, cloud.IPVlan),
	}
}
func overridePtr[T any](<user, cloud *T>) *T {
	if user != nil {
		return ptr.To(*user)
	}
	if cloud != nil {
		return ptr.To(*cloud)
	}
	return nil
}
func firstNonEmpty[T ~string](<user, cloud T>) T {
	if user != "" {
		return user
	}
	return cloud
}
```
 something like this will also fail if we add a new field and we forget to convert it

```go
func TestMergeCoversAllInterfaceFields(t *testing.T) {
	// 1. Populate user and cloud with non-zero values for every field
	user := &NetworkConfig{
		Interface: InterfaceConfig{
			Name: "eth0", Type: InterfaceTypeIPVLAN, Addressing: AddressingModeDHCP,
			Addresses: []string{"10.0.0.1/32"}, DHCP: ptr.To(false), MTU: ptr.To[int32](1400),
			HardwareAddr: ptr.To("00:11:22:33:44:55"), GSOMaxSize: ptr.To[int32](65536),
			GROMaxSize: ptr.To[int32](65536), GSOIPv4MaxSize: ptr.To[int32](65536),
			GROIPv4MaxSize: ptr.To[int32](65536), DisableEBPFPrograms: ptr.To(true),
			Forwarding: ptr.To(true), ARPIgnore: ptr.To[int32](0), ARPAnnounce: ptr.To[int32](0),
			VRF: &VRFConfig{Name: "vrf0", Table: ptr.To(100)},
			IPVlan: &IPVlanConfig{Mode: IPVlanModeL2, Flag: IPVlanFlagBridge},
		},
	}
	cloud := &NetworkConfig{
		Interface: InterfaceConfig{
			Name: "eth1", Type: InterfaceTypePassthrough, Addressing: AddressingModeStatic,
			Addresses: []string{"192.168.1.1/32"}, DHCP: ptr.To(true), MTU: ptr.To[int32](1500),
			HardwareAddr: ptr.To("aa:bb:cc:dd:ee:ff"), GSOMaxSize: ptr.To[int32](32768),
			GROMaxSize: ptr.To[int32](32768), GSOIPv4MaxSize: ptr.To[int32](32768),
			GROIPv4MaxSize: ptr.To[int32](32768), DisableEBPFPrograms: ptr.To(false),
			Forwarding: ptr.To(false), ARPIgnore: ptr.To[int32](2), ARPAnnounce: ptr.To[int32](2),
			VRF: &VRFConfig{Name: "vrf1", Table: ptr.To(200)},
			IPVlan: &IPVlanConfig{Mode: IPVlanModeL2, Flag: IPVlanFlagBridge},
		},
	}

	merged := MergeNetworkConfig(user, cloud)

	// 2. Walk every field of InterfaceConfig: verify it's mapped and not aliased
	valMerged := reflect.ValueOf(merged.Interface)
	valUser := reflect.ValueOf(user.Interface)
	valCloud := reflect.ValueOf(cloud.Interface)
	typ := valMerged.Type()

	for i := 0; i < typ.NumField(); i++ {
		f := typ.Field(i)
		mField := valMerged.Field(i)

		switch f.Type.Kind() {
		case reflect.Pointer:
			if mField.IsNil() {
				t.Errorf("field %s is nil in merged config (forgotten in merge?)", f.Name)
				continue
			}
			if mField.Pointer() == valUser.Field(i).Pointer() || mField.Pointer() == valCloud.Field(i).Pointer() {
				t.Errorf("field %s aliases input pointer", f.Name)
			}
		case reflect.Slice:
			if mField.Len() == 0 {
				t.Errorf("slice field %s was not merged", f.Name)
			}
			if mField.Pointer() == valUser.Field(i).Pointer() || mField.Pointer() == valCloud.Field(i).Pointer() {
				t.Errorf("slice field %s aliases input backing array", f.Name)
			}
		}
	}
}

```

OguzPastirmaci:
> Let's ditch `mergo`, I just asked an AI tool to give me a snippet for reference but this seems the best alternative (code needs to be evaluated)

Sounds good. Pushed some changes based on the direction and snippet you shared @aojea. Please note this was mostly AI generated code that we can use as a starting point. Tested in a live cluster.

pkg/apis/config_merge.go
```diff
@@ -17,53 +17,148 @@ limitations under the License.
...
+	}
+	if user == nil {
+		user = &VRFConfig{}
+	}
+	if cloud == nil {
+		cloud = &VRFConfig{}
+	}
+	return &VRFConfig{
+		Name:  firstNonEmpty(user.Name, cloud.Name),
+		Table: overridePtr(user.Table, cloud.Table),
```
dkennetzoracle, line 94: To note, this behavior existed before this PR, but I want to call it out just so we're aligned.

user: `vrf: {name: "b"}`
provider: `vrf: {name: "a", table: 100}`
yields: :`vrf: {name: "b", table: 100}`

I don't think this is the behavior we want. This kind of defeats the purpose of deriving table IDs from the name and leaks routes across VRF domains.

Again, mergo also behaved this way, but I think we want to fix this.

The minimal fix would be, when both sides set a name and they differ, take the user's VRF block wholesale (leaving `Table` nil so `Default()` re-derives it).

pkg/apis/config_merge_coverage_test.go
```diff
@@ -0,0 +1,313 @@
...
+			return
+		}
+		if a.Kind() == reflect.Pointer {
+			assertNoSharedMemory(t, path, a.Elem(), b.Elem())
+		}
+	case reflect.Slice:
+		if a.Len() == 0 || b.Len() == 0 {
+			return
+		}
+		if a.Pointer() == b.Pointer() {
```
dkennetzoracle, line 294: If any field in the array ever gained a pointer, they could share memory (IE if we ever added a pointer or map to RouteConfig) `deduplicateRoutes` would have the same pointer.. Example:

```go
deduplicateRoutes(append(slices.Clone(cloud.Routes), user.Routes...))
...
merged.Routes[0].Metric == cloud.Routes[0].Metric // same pointer

*merged.Routes[0].Metric = 9 
```
The backing arrays differ so they'd pass this test, but the pointers in the field of `RouteConfig` would end up being the same. Should we use `reflect.DeepEqual(a, b)` instead?

pkg/apis/config_merge_coverage_test.go
```diff
@@ -0,0 +1,313 @@
...
+		if !zero {
+			v.SetString(marker + ":" + path)
+		}
+	case reflect.Bool:
+		v.SetBool(!zero)
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		v.SetInt(n)
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		v.SetUint(uint64(n))
+	case reflect.Slice:
```
dkennetzoracle, line 98: Similar as the problem below in assertNoSharedMem - since this doesn't traverse types it treats every slice as a leaf without descending into element types.

pkg/apis/config_merge_coverage_test.go
dkennetzoracle: In fact, we may just want to assert that fields cannot be pointers @OguzPastirmaci @aojea - would be an easier fix.

pkg/apis/config_merge_coverage_test.go
```diff
@@ -0,0 +1,313 @@
...
+func TestMergeKeepsEveryFieldFromEitherSide(t *testing.T) {
+	full := fullConfig("x", 0)
+	for _, tc := range []struct {
+		name        string
+		user, cloud *NetworkConfig
+	}{
+		{"user only", full, &NetworkConfig{}},
+		{"cloud only", &NetworkConfig{}, full},
+		{"user only, nil cloud", full, nil},
+		{"cloud only, nil user", nil, full},
```
dkennetzoracle: I think this just compares the input against itself. I think adding:
`want := fullConfig("x", 0)`

and comparing against that would be better.

Today, if merge mutated its input this test would compare mutated input against mutated output and pass.

pkg/apis/config_merge_coverage_test.go
```diff
@@ -0,0 +1,313 @@
...
+		m := reflect.MakeMap(v.Type())
+		m.SetMapIndex(key, val)
+		v.Set(m)
+	default:
+		panic("setValue: unsupported kind " + v.Kind().String() + " at " + path)
+	}
+}
+
+// fullConfig sets every leaf. Numbers are offset + leaf ordinal, so two configs
+// with different offsets share no number. Keep offset + leaf count under 256
```
dkennetzoracle: We should make this an asserting (uint8) not a comment.

pkg/apis/config_merge_test.go
```diff
@@ -240,3 +286,494 @@ func TestMergeNetworkConfig(t *testing.T) {
...
+	}
+	for _, name := range deepCopyInterfacePointerFields {
+		expectedDeepCopies = append(expectedDeepCopies, "Interface."+name)
+	}
+
+	wantDeepCopy := listAsSet(expectedDeepCopies)
+	gotDeepCopy := listAsSet(mutableFieldPaths(reflect.TypeOf(NetworkConfig{}), ""))
+	for path := range gotDeepCopy {
+		if !wantDeepCopy[path] {
+			t.Errorf("mutable field %s is not covered by deepCopyNetworkConfig", path)
```
dkennetzoracle: nit: I think this name is old

pkg/apis/config_merge_test.go
```diff
@@ -240,3 +286,494 @@ func TestMergeNetworkConfig(t *testing.T) {
...
+		}
+	}
+	for path := range wantDeepCopy {
+		if !gotDeepCopy[path] {
+			t.Errorf("deepCopyNetworkConfig lists unknown mutable field %s", path)
```
dkennetzoracle: nit: name

pkg/apis/config_merge_test.go
```diff
@@ -240,3 +286,494 @@ func TestMergeNetworkConfig(t *testing.T) {
...
+		switch field.Type.Kind() {
+		case reflect.Pointer:
+			paths = append(paths, path)
+			if field.Type.Elem().Kind() == reflect.Struct {
+				paths = append(paths, mutableFieldPaths(field.Type.Elem(), path)...)
+			}
+		case reflect.Slice:
+			paths = append(paths, path)
+			if field.Type.Elem().Kind() == reflect.Struct {
+				paths = append(paths, mutableFieldPaths(field.Type.Elem(), path+"[]")...)
```
dkennetzoracle, line 700: This is what I was referring to above about traversing the slice rather than comparing the pointers.

pkg/apis/config_merge_test.go
```diff
@@ -240,3 +286,494 @@ func TestMergeNetworkConfig(t *testing.T) {
...
+	}
+	copy := &NetworkConfig{}
+	if err := json.Unmarshal(data, copy); err != nil {
+		t.Fatalf("unmarshal NetworkConfig: %v", err)
+	}
+	return copy
+}
+
+// These lists keep the merge helpers in sync with the configuration types.
+var deepCopyInterfacePointerFields = []string{
```
dkennetzoracle, line 519: Are we reconciling tests this PR or next PR?

dkennetzoracle (changes requested):
A few changes from my end and some clarifications on the PR.  Overall, solid.

pkg/apis/config_merge.go
aojea: Interesting, this brings the topic on when data should be defaulted ... so basically , we treat this type like an unit ... I think we need to write some API convention doc in this folder first and agree on that

pkg/apis/config_merge_coverage_test.go
aojea: the more opinionated we are the more work we need to do in apimachinery, and I do not think we want to invest time on that ... `reflect.DeepEqual(a, b)` sounds like a simple solution to test but I'm unclear now what are the consequences of imposing that constrain

dkennetzoracle: I believe this test may already exist in the prior test which did list-wise comparison (see comments in config_merge_test.go:697):

> This is what I was referring to above about traversing the slice rather than comparing the pointers.

```go
		case reflect.Slice:
			paths = append(paths, path)
			if field.Type.Elem().Kind() == reflect.Struct {
				paths = append(paths, mutableFieldPaths(field.Type.Elem(), path+"[]")...)
```

pkg/apis/config_merge_coverage_test.go
OguzPastirmaci: Fixed in f1070a9.

pkg/apis/config_merge_coverage_test.go
OguzPastirmaci: Fixed in f1070a9.

pkg/apis/config_merge.go
drkennetz: @aojea I think "when" and "which" data should be defaulted, but both of those require more information around API convention. If we could say... "the following data are always defaulted" and "the following data is never defaulted" that would help although I'm not sure how feasible it is.

PR #317, opened 2026-09-03
apis: default and validate the merged NetworkConfig

### LINKED ISSUE #315: The merged NetworkConfig is never defaulted or validated before it is applied
# Summary

`MergeNetworkConfig` combines the user's `ResourceClaim` config with the cloud provider's device config and profile config. The result is what actually gets applied to netlink — but it never passes through `Default()` or the validation functions in `pkg/apis/validation.go`.

Validation only ever runs on the user's raw claim parameters: `dra_hooks.go:217` (`apis.ValidateConfig`) and `dra_hooks.go:278` (`apis.ValidateRDMAOnlyConfig`). `NetworkConfig.Default()` has exactly one caller, `validation.go:64`, inside `ValidateConfig` — so it too only ever sees the user's config. `getDeviceNetworkConfig` (`dra_hooks.go:709-724`) returns mergedConf with no further checks.

## Why this matters

Provider-supplied configs are unvalidated external input — WebhookProvider.GetDeviceConfig/GetProfileConfig (webhook.go:230-254) just unmarshal whatever the webhook returns. So the invariants in validation.go hold for user configs and silently do not hold for anything a provider contributes, or for any cross-field combination that only appears after merging.

## Proposed fix

Extract the post-unmarshal half of ValidateConfig into func (c *NetworkConfig) Validate() []error (moving c.Default() and the four validate blocks verbatim), have ValidateConfig call it, then add one guard in getDeviceNetworkConfig after the profile merge returning fmt.Errorf("merged network config for device %s is invalid: %v", device, errs). No validation rules change; behavior-preserving for the existing caller, so validation_test.go should pass unchanged.

Notes: related to #294 but distinct: #294 is about MergeNetworkConfig's internal correctness, this is about the merged result never being checked regardless. Also flags an open question about validating provider configs at ingest (db.go:570-604) for a nicer error surface.

dkennetzoracle:
#### What type of PR is this?

/kind cleanup
/kind api-change

MergeNetworkConfig combines the user's ResourceClaim config with the cloud provider's device config and the resolved profile config, and the result is what gets applied to netlink. Validation only ever ran on the user's raw claim parameters, and NetworkConfig.Default() only ever ran inside ValidateConfig, so nothing defaulted or checked the merged config.

Provider and profile configs are unvalidated external input, and some invariants only become checkable once everything is merged:

  - A provider supplying "type: IPVLAN" with an empty "ipvlan" block was never defaulted, so addIPVlan rejected the empty mode with "unsupported ipvlan mode". The identical config from a user works, because ValidateConfig defaults it first.
  - A user supplying rules plus a provider supplying a VRF produced a merged config that validation declares unsupported, with nothing left to reject it.

Extract the post-unmarshal half of ValidateConfig into NetworkConfig.Validate() and call it on the merged config in getDeviceNetworkConfig. No validation rules change; ValidateConfig keeps its existing behavior by delegating to Validate.

Fixes #315 
Found in #316

There is a behavior change for GCE + addressing: unnumbered.

Validating the merged config newly rejects one combination that is currently tolerated, reachable on GCE without any user errors. 

GCEInstance.GetProfileConfig (gce.go:197) gates only on IsSubinterface(). An Unnumbered IPVLan interface is a sub interface, so GCE allocates addresses from IPAM even though the user explicitly waved addressing. The merged config then carries both:

```
addressing=Unnumbered addrs=[10.128.0.5/32]
errs=[interface.addressing: 'Unnumbered' cannot be set together with addresses]
```

Today, that reaches Dra_hooks.go:454, which logs `skipping address and route configuration` and proceeds. With this PR it's a failure.

I think failing is the correct case here: `validateInterfaceConfig` explicitly forbids the combination, and allocating an IPAM lease for an interface the user asked to be unnumbered is a provider bug that should surface rather than be silently absorbed.

The GCE-side fix would be to skip allocation and PBR synthesis when `Addressing == Unnumbered`. I could take a shot at it, but I can't test it myself.

Alternatively I could change the PR to let this slide, but I don't think we should.

NONE

```release-note
Cleaned up config validation, where merged configs were only getting validated and defaulted before the merge. 
```

### DIFF
--- a/pkg/apis/validation.go
+++ b/pkg/apis/validation.go
@@ -37,7 +37,7 @@ const (
 )
 
 // ValidateConfig unmarshals and validates the NetworkConfig from a runtime.RawExtension.
-// It performs strict unmarshalling and then calls specific validation functions for each part of the config.
+// It performs strict unmarshalling and then defaults and validates the result via Validate.
 // Returns the parsed NetworkConfig and a slice of errors if any validation fails.
 func ValidateConfig(raw *runtime.RawExtension) (*NetworkConfig, []error) {
 	if raw == nil || raw.Raw == nil || len(raw.Raw) == 0 {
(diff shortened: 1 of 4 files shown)


aojea:
good 

/lgtm
/approve

PR #318, opened 2026-09-04
fix: clear copied host route source not assigned in the Pod

yeonghoo2:
/kind bug

`getRouteInfo()` copies the host routes of an interface into the Pod, including
their preferred source (`src` in `ip route`). When the Pod gets different
addresses than the host had, that source does not exist in the Pod namespace
and the kernel rejects the route with EINVAL (`fib_create_info`: "Invalid
prefsrc address"). `applyRoutes` only tolerates EEXIST, so prepare fails and
the Pod is stuck in `ContainerCreating` while the host has already lost the
interface.

This happens whenever a copied route carries a source and the Pod addresses
differ from the host ones: `addressing: dhcp`, or `addresses` set in the claim.
The kernel adds a `src` to the connected route of any interface that has an
address on the host, so it shows up on interfaces the host had configured, for
example a PF:

```
172.17.14.0/24 dev ens4013np0 proto kernel scope link src 172.17.14.28
```

The fix clears the source of a copied route when it is not one of the Pod's
addresses. A source that matches a Pod address is kept, so a Pod that keeps the
host addresses is unchanged. Without a source the kernel picks one itself. This follows the existing filters in `getRouteInfo` that drop
copied routes which cannot be valid in the Pod (local table, IPv6 link-local,
IPv6 `proto kernel`).

N/A

`TestClearStaleRouteSources` covers the source being cleared, kept when it
matches a Pod address, and routes without a source left untouched.

`go build ./...`, `go vet` and `gofmt` are clean. `go test ./pkg/driver/` has
the same six failures as an unmodified `main` in my environment (tests that
need a real netns or ethtool); no new failures.

We have been running this fix in production with DHCP mode for a few weeks.

```release-note
Fix Pod creation failing with EINVAL when a host route copied into the Pod has a preferred source that is not one of the Pod's addresses (e.g. with DHCP or addresses set in the claim).
```

### DIFF
--- a/pkg/driver/dra_hooks.go
+++ b/pkg/driver/dra_hooks.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"slices"
 	"sort"
 	"strings"
@@ -404,6 +405,7 @@ func (np *NetworkDriver) prepareResourceClaim(ctx context.Context, claim *resour
 				errorList = append(errorList, err)
 				continue
 			}
+			clearStaleRouteSources(routes, deviceCfg.NetworkInterfaceConfigInPod.Interface.Addresses)
 			deviceCfg.NetworkInterfaceConfigInPod.Routes = append(deviceCfg.NetworkInterfaceConfigInPod.Routes, routes...)
 
 			// If VRF is enabled, we do not need to copy the rules from the host
@@ -702,6 +704,26 @@ func getRouteInfo(nlHandle nlwrap.Handle, ifName string, link netlink.Link) ([]a
 		}
 	}
 	return routes, tables, nil
+}
+
+// clearStaleRouteSources drops the preferred source of copied host routes when
+// that address is not among the addresses the interface gets in the Pod. The
+// kernel rejects a route whose prefsrc is not a local address with EINVAL, and
+// the Pod may get different addresses than the host had (DHCP, or addresses
+// set in the claim). Without a source the kernel picks one itself.
+func clearStaleRouteSources(routes []apis.RouteConfig, addresses []string) {
+	podIPs := sets.New[string]()
+	for _, addr := range addresses {
+		if ip, _, err := net.ParseCIDR(addr); err == nil {
+			podIPs.Insert(ip.String())
+		}
+	}
+	for i := range routes {
+		if routes[i].Source != "" && !podIPs.Has(routes[i].Source) {
+			klog.V(4).Infof("Clearing source %s of route %s: address not assigned in the Pod", routes[i].Source, routes[i].Destination)
+			routes[i].Source = ""
+		}
+	}
 }
 
 // getDeviceNetworkConfig merges the user configuration with the cloud provider configuration and resolves the dynamic profile.
--- a/pkg/driver/dra_hooks_test.go
+++ b/pkg/driver/dra_hooks_test.go
@@ -1416,3 +1416,59 @@ func testPrepareResourceClaim_Namespaced(t *testing.T) {
 		})
 	}
 }
+
+func TestClearStaleRouteSources(t *testing.T) {
+	testCases := []struct {
+		name        string
+		routes      []apis.RouteConfig
+		addresses   []string
+		wantSources []string
+	}{
+		{
+			name:        "source not assigned in the pod is cleared",
+			routes:      []apis.RouteConfig{{Destination: "172.17.14.0/24", Source: "172.17.14.28"}},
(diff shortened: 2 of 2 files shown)


pkg/driver/dra_hooks_test.go
```diff
@@ -1416,3 +1416,44 @@ func testPrepareResourceClaim_Namespaced(t *testing.T) {
...
+		},
+		{
+			name: "routes without source are untouched",
+			routes: []apis.RouteConfig{
+				{Destination: "0.0.0.0/0", Gateway: "172.17.14.1"},
+				{Destination: "10.0.0.0/8", Gateway: "172.17.14.1", Source: "172.17.14.28"},
+			},
+			addresses:   nil,
+			wantSources: []string{"", ""},
+		},
```
anson627, line 1462: nit: add an IPv6 case to confirm canonicalized CIDRs match RouteConfig.Source as expected

pkg/driver/dra_hooks_test.go
```diff
@@ -1416,3 +1416,44 @@ func testPrepareResourceClaim_Namespaced(t *testing.T) {
...
+			wantSources: []string{""},
+		},
+		{
+			name:        "source assigned in the pod is kept",
+			routes:      []apis.RouteConfig{{Destination: "172.17.14.0/24", Source: "172.17.14.28"}},
+			addresses:   []string{"172.17.14.28/24"},
+			wantSources: []string{"172.17.14.28"},
+		},
+		{
+			name: "routes without source are untouched",
```
dkennetzoracle: Nit: this has a source so it is clearing right? Consider rename or split

dkennetzoracle (commented):
/lgtm

pkg/driver/dra_hooks_test.go
yeonghoo2: Thanks for the review. Added two IPv6 cases, one with a non-canonical pod address that has to match the canonical source.

pkg/driver/dra_hooks_test.go
yeonghoo2: Thanks for the review, good catch. Split it into two cases: a route without a source, and a source cleared when the pod has no addresses.

pkg/driver/dra_hooks.go
```diff
@@ -704,6 +706,26 @@ func getRouteInfo(nlHandle nlwrap.Handle, ifName string, link netlink.Link) ([]a
 	return routes, tables, nil
 }
 
+// clearStaleRouteSources drops the preferred source of copied host routes when
+// that address is not among the addresses the interface gets in the Pod. The
+// kernel rejects a route whose prefsrc is not a local address with EINVAL, and
+// the Pod may get different addresses than the host had (DHCP, or addresses
+// set in the claim). Without a source the kernel picks one itself.
+func clearStaleRouteSources(routes []apis.RouteConfig, addresses []string) {
```
aojea, line 714: Can you explain more the use case behind this behavior?
there seems to be a. lot of assumptions here ,so a partial reconfiguration of a route looks confusing to me

Where are those routes coming from? prepopulated by the system or used configured?
What happens if the gateway is not in the same subnet?

yeonghoo2: The routes are system-populated, not user-configured: when netplan gives the interface an address, the kernel adds the connected route with that address as `src` (`fib_magic`).

The case we hit: a bare-metal PF `ens4013np0` had `172.17.14.28/24` from netplan and the Pod got a different address on it, so the copied `172.17.14.0/24 dev ens4013np0 proto kernel scope link src 172.17.14.28` failed with EINVAL and the sandbox never came up. It happens whenever the Pod address differs from the host one, whether from a profile, `addressing: dhcp`, or `addresses` in the claim.

Why only clear `src`: dropping the route would take the Pod's connected/default route with it, and skipping the copy when addresses are overridden would change behaviour for claims that set `addresses` and rely on the host routes. Clearing the hint keeps the route and lets the kernel pick the source as for any route without `src`, in line with the other filters in `getRouteInfo`.

Gateway in another subnet: `src` does not affect gateway reachability. The connected route is copied too, so `default via 172.17.14.254` still installs and traffic is sourced from the Pod address. Whether the upstream router accepts that source is unchanged by this PR; without it the Pod does not start at all.

anson627:
e2e failure says "ERROR: unknown cluster "kind"", unrelated to this change

aojea:
/lgtm
/approve

Ok, https://github.com/kubernetes-sigs/dranet/pull/318#discussion_r3945862878

PR #319, opened 2026-09-04
docs: update Kubernetes version and mark DRA as stable

rikatz:
#### What type of PR is this?

/kind documentation

DRA is stable/GA since Kubernetes 1.35. This PR:
* Update the docs statement about DRA being Beta, and instead mark it as stable
* Fixes the kind configuration file that was pointing to a wrong/unexisting tag

```release-note
NONE
```


aojea:
/lgtm
/approve

Thanks

PR #320, opened 2026-09-07
Azure: publish per-device IPv4 subnet attribute

nitin-nizhawan:
## What type of PR is this?
/kind feature

## What this PR does
Publishes each Azure network interface's IPv4 subnet CIDR as the device-specific `azure.dra.net/ipv4Subnet` ResourceSlice attribute.

The Azure IMDS interface is matched to the DraNet device by normalized MAC address. The attribute is omitted when the device MAC, matching interface, or complete IPv4 subnet metadata is unavailable. Existing node-level Azure attributes are unchanged.

An IPv4-specific attribute name is used because Azure dual-stack NICs can also have an IPv6 subnet, while Azure IMDS currently exposes explicit subnet address and prefix metadata under the IPv4 interface data.

## Testing
- `go test -race -count 1 ./pkg/cloudprovider/azure`
- `go test ./pkg/inventory ./pkg/driver`

```release-note
Expose the Azure IPv4 subnet CIDR for each network device as the `azure.dra.net/ipv4Subnet` ResourceSlice attribute.
```


nitin-nizhawan:
Closing this upstream PR. The change will remain as a commit in the contributor fork for now.

PR #325, opened 2026-09-08
cloudprovider: Extend the OKE provider with fabric discovery and refresh

OguzPastirmaci:
#### What type of PR is this?

/kind feature

The OKE cloud provider publishes three new node-level attributes on every device. The five existing topology attributes do not change.

- `oke.dra.net/shape`, from `/opc/v2/instance/`.
- `oke.dra.net/rdmaFabricIpv6` and `oke.dra.net/rdmaFabricPlanes`, from `rdmaFabricData`.

The provider refreshes the snapshot every 5 minutes with jitter. Before this PR, `GetInstance` returned an error when IMDS did not answer within 15 seconds. Dranet then ran without the OKE provider. Now `GetInstance` returns the provider with the metadata it has. It keeps refreshing in the background. Automatic detection still needs the unchanged 5 second probe of `/opc/v2/instance/` to succeed. The `--cloud-provider-hint=OKE` flag is the alternative.

N/A

```release-note
The OKE cloud provider publishes `oke.dra.net/shape`, `oke.dra.net/rdmaFabricIpv6`, and `oke.dra.net/rdmaFabricPlanes`, refreshes its metadata every 5 minutes, and stays active when IMDS is not ready at startup.
```

### DIFF
--- a/pkg/cloudprovider/oke/oke.go
+++ b/pkg/cloudprovider/oke/oke.go
@@ -17,13 +17,14 @@ limitations under the License.
 package oke
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"net/http"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -44,13 +45,26 @@ const (
 	AttrOKERackId          = OKEAttrPrefix + "/" + "rackId"
 	AttrOKEGpuMemoryFabric = OKEAttrPrefix + "/" + "gpuMemoryFabricId"
 
+	// Instance shape (from /opc/v2/instance/).
+	AttrOKEShape = OKEAttrPrefix + "/" + "shape"
+
+	// RDMA fabric attributes (from the rdmaFabricData object in /opc/v2/host/).
+	AttrOKERDMAFabricIPv6   = OKEAttrPrefix + "/" + "rdmaFabricIpv6"
+	AttrOKERDMAFabricPlanes = OKEAttrPrefix + "/" + "rdmaFabricPlanes"
+
 	// imdsEndpoint is the Oracle Cloud Instance Metadata Service endpoint.
 	imdsEndpoint = "http://169.254.169.254/opc/v2"
+
+	imdsInitialRetryInterval = 1 * time.Second
+	imdsInitialWait          = 15 * time.Second
+	imdsRefreshInterval      = 5 * time.Minute
+	// imdsRequestTimeout limits each request, at startup and in the background.
+	imdsRequestTimeout = 5 * time.Second
 )
 
-// imdsHostRDMATopologyData contains the RDMA topology fields from the OCI
-// IMDS host metadata response. This is only populated when RDMA topology
-// data is enabled for the tenancy.
+// imdsHostRDMATopologyData contains the RDMA topology fields from the
+// /opc/v2/host/ response. IMDS populates them only for instances in a
+// dedicated pool, with RDMA topology data enabled for the tenancy.
 type imdsHostRDMATopologyData struct {
 	CustomerGpuMemoryFabric string `json:"customerGpuMemoryFabric"`
 	CustomerHPCIslandId     string `json:"customerHPCIslandId"`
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/oke/oke.go
```diff
@@ -59,54 +87,122 @@ type imdsHostRDMATopologyData struct {
...
-		attributes[AttrOKEGpuMemoryFabric] = resourceapi.DeviceAttribute{StringValue: &o.GpuMemoryFabric}
+	if metadata.GpuMemoryFabric != "" {
+		attributes[AttrOKEGpuMemoryFabric] = resourceapi.DeviceAttribute{StringValue: &metadata.GpuMemoryFabric}
+	}
+	if metadata.Shape != "" {
+		attributes[AttrOKEShape] = resourceapi.DeviceAttribute{StringValue: &metadata.Shape}
+	}
+	if metadata.RDMAFabric != nil {
+		attributes[AttrOKERDMAFabricIPv6] = resourceapi.DeviceAttribute{BoolValue: &metadata.RDMAFabric.IPv6}
+		attributes[AttrOKERDMAFabricPlanes] = resourceapi.DeviceAttribute{IntValue: &metadata.RDMAFabric.Planes}
```
dkennetzoracle, line 171: Could either of these be nil even though RDMAFabric is not nil? Seems like this could be a null ptr dereference if so. In all other cases we are checking the values directly, but here you are just ensuring that the structure is not nil, so the fields could still be nil. I wouldn't hate something like:
```go
if metadata.RDMAFabric != nil {
    if metadata.RDMAFabric.IPV6 { // this is also a bool, should we convert to string?
     ...
    }
    if metadata.RDMAFabric.Planes != 0 { // check this conversion since it is int64
    ....
    }
```

dkennetzoracle: Oh I see below on 283 that a partial response is treated like an absent one.

I would still prefer this check because if anyone changes the behavior below (a partial response becomes allowed), they will have to do this in the future. Your check is correct for today's behavior (and a bit less verbose), but this way would be correct for today and if that behavior changed.

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +255,340 @@ func OnOKE(ctx context.Context) bool {
...
+	fabric := getRDMAFabricData(ctx, client, endpoint, &host, requiresFabricData)
+
+	instance := &imdsInstanceMetadata{}
+	if err := queryIMDS(ctx, client, endpoint+"/instance/", instance); err != nil {
+		klog.Warningf("Could not query OCI IMDS instance metadata: %v", err)
+		instance = nil
+	}
+
+	metadata, err := metadataFromIMDS(&host, instance, fabric)
+	if err == nil && fabric != nil {
```
dkennetzoracle: Does this + the case below permanently pin the fallback value? IE on startup, you fail transiently from /host/rdmaFabricData which falls back. Then you stamp with requiresFabricData which satisfies metadataComplete which causes the case to hit current.FabricConfirmed which rejects it...


Should we set FabricConfirrmed only when the direct query actually succeeded?

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +255,340 @@ func OnOKE(ctx context.Context) bool {
...
+	}
+	return metadata, nil
+}
+
+// getRDMAFabricData prefers the direct endpoint when fabric data is required
+// and falls back to the object embedded in the host response.
+func getRDMAFabricData(ctx context.Context, client *http.Client, endpoint string, host *imdsHostMetadata, required bool) *rdmaFabric {
+	if required {
+		var direct imdsRDMAFabricData
+		if err := queryIMDS(ctx, client, endpoint+"/host/rdmaFabricData/", &direct); err != nil {
```
dkennetzoracle: Is there a good reason to prefer this instead of just using the embedded object?

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +255,340 @@ func OnOKE(ctx context.Context) bool {
...
+}
+
+// fetchOKEMetadata reads the host metadata first, the fabric data second when
+// it is required, and the optional instance metadata last.
+func fetchOKEMetadata(ctx context.Context, client *http.Client, endpoint string, requiresFabricData bool) (*okeMetadata, error) {
+	var host imdsHostMetadata
+	if err := queryIMDS(ctx, client, endpoint+"/host/", &host); err != nil {
+		return nil, err
+	}
+	fabric := getRDMAFabricData(ctx, client, endpoint, &host, requiresFabricData)
```
dkennetzoracle: Do we need to do 2 queries here, rather than just pulling from the first?

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +255,340 @@ func OnOKE(ctx context.Context) bool {
...
+
+	instance := &imdsInstanceMetadata{}
+	if err := queryIMDS(ctx, client, endpoint+"/instance/", instance); err != nil {
```
dkennetzoracle, line 311: It is possible for this to become unreachable now because the function returns early on the query failure above, which I think defeats the purpose. This is the metadata that's always available (IE instance metadata). I think this should be independent of the host query result, right?

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +255,340 @@ func OnOKE(ctx context.Context) bool {
...
+		return 0, fmt.Errorf("could not read %s: %w", name, err)
+	}
+	value, err := strconv.Atoi(strings.TrimSpace(string(data)))
+	if err != nil {
+		return 0, fmt.Errorf("could not parse %s: %w", name, err)
+	}
+	return value, nil
+}
+
+func interfaceNameForPCIAddress(pciAddress string) (string, error) {
```
dkennetzoracle: Comment here: `pkg/inventory/db.go:750-757` let's review this

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +255,340 @@ func OnOKE(ctx context.Context) bool {
...
+func isEthernetFabricDevice(id cloudprovider.DeviceIdentifiers) bool {
+	ifName, err := interfaceNameForPCIAddress(id.PCIAddress)
+	if err != nil || !isFabricInterface(ifName) {
+		return false
+	}
+	hardwareType, err := interfaceHardwareType(ifName)
+	return err == nil && hardwareType == unix.ARPHRD_ETHER
+}
+
+func hasEthernetFabricInterface() bool {
```
dkennetzoracle: Comment here: `pkg/inventory/db.go:750-757` let's review this.

pkg/cloudprovider/oke/oke_test.go
```diff
@@ -170,13 +222,631 @@ func TestOCIDSuffix(t *testing.T) {
...
+	got, err := instance.start(ctx, server.Client(), server.URL)
+	if !errors.Is(err, context.Canceled) {
+		t.Fatalf("start() returned error %v, want context.Canceled", err)
+	}
+	if got != nil {
+		t.Fatalf("start() returned %v with an error, want nil", got)
+	}
+}
+
+func TestStartRecoversFromHostOutage(t *testing.T) {
```
dkennetzoracle: Could we also exercise /host/ down and /instance/ up?

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +229,201 @@ func OnOKE(ctx context.Context) bool {
...
+		cancel()
+		if lastErr != nil {
+			klog.Infof("could not get OCI IMDS metadata ... retrying: %v", lastErr)
+			return false, nil
 		}
 		return true, nil
 	})
 	if err != nil {
-		if errors.Is(err, context.DeadlineExceeded) {
-			return nil, fmt.Errorf("please enable TopologyData for your tenancy: %w", err)
```
dkennetzoracle, line 235: I see this was deleted which probably makes sense for customer tenancies. However, if it does fail, now the query to get "shape" is unreachable with the other changes in the PR.

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +229,201 @@ func OnOKE(ctx context.Context) bool {
...
+	o.fetchMetadata = func(ctx context.Context) (*okeMetadata, error) {
+		return fetchOKEMetadata(ctx, client, endpoint)
+	}
+
+	var lastErr error
+	err := wait.PollUntilContextTimeout(ctx, o.initialRetryInterval, o.initialWait, true, func(ctx context.Context) (bool, error) {
+		requestCtx, cancel := context.WithTimeout(ctx, o.initialRequestTimeout)
+		lastErr = o.refreshMetadata(requestCtx)
+		cancel()
+		if lastErr != nil {
```
dkennetzoracle, line 402: I think this makes non-retry able errors retryable forever. IE, if for some reason we parsed an OCID incorrectly because OCIDs change in some meaningful way, we will return early on a parse error above which will just continually be retried.

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +229,201 @@ func OnOKE(ctx context.Context) bool {
...
+		return next, ""
+	}
+
+	merged := *next
+	merged.HPCIslandId = cmp.Or(merged.HPCIslandId, current.HPCIslandId)
+	merged.NetworkBlockId = cmp.Or(merged.NetworkBlockId, current.NetworkBlockId)
+	merged.LocalBlockId = cmp.Or(merged.LocalBlockId, current.LocalBlockId)
+	merged.RackId = cmp.Or(merged.RackId, current.RackId)
+	merged.GpuMemoryFabric = cmp.Or(merged.GpuMemoryFabric, current.GpuMemoryFabric)
+	merged.Shape = cmp.Or(merged.Shape, current.Shape)
```
dkennetzoracle, line 334: Why do we change everything except rdmaFabricData? It seems like this is present to handle the case of missing data, but we have opposite behavior here depending on which field it is (IE these vs rdmaFabric), and only the fabric case is treated as an error. It seems like these fields are at least "as immutable" as the rdmaFabricData?

pkg/cloudprovider/oke/oke.go
```diff
@@ -155,86 +229,201 @@ func OnOKE(ctx context.Context) bool {
...
+}
 
-		body, err := io.ReadAll(resp.Body)
-		if err != nil {
-			klog.Infof("could not read OCI IMDS host response ... retrying: %v", err)
-			return false, nil
-		}
+// fabric returns the value when both fields are present, otherwise nil.
+// A partial object is treated like an absent one.
+func (d *imdsRDMAFabricData) fabric() *rdmaFabric {
```
dkennetzoracle, line 251: I think I get the intent here, but should we split these for diagnostic purposes? IE, could there ever be a case with incomplete metadata from the response which is categorically different than how this is defined?

pkg/cloudprovider/oke/oke_test.go
```diff
@@ -170,13 +222,631 @@ func TestOCIDSuffix(t *testing.T) {
...
+			t.Errorf("reader %d observed no snapshot", i)
+		}
+	}
+}
+
+func TestStartReturnsMetadata(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
+		case "/host/":
+			_, _ = w.Write([]byte(`{"networkBlockId":"network-1","rackId":"rack-1","rdmaFabricData":{"ipv6":true,"planes":0}}`))
```
dkennetzoracle: Is `0` a real value for Planes? I see above that many test cases default `ipv6: false, planes: 0`, but this one has true for ipv6 and planes still 0.

dkennetzoracle (commented):
Will sync on a call to better understand the changes. For now, there is a lot of encoded logic that I don't fully understand.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: The direct endpoint and the confirmation state were removed in the latest push. The provider reads the fabric object from the /host/ response only.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: > Is there a good reason to prefer this instead of just using the embedded object?

Nope. The latest push uses the embedded object onnly now

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: The direct query is removed in the latest push. Only /host and /instance are read now.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: Your network namespace concern is valid, removed both helpers.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: > I think this makes non-retry able errors retryable forever. IE, if for some reason we parsed an OCID incorrectly because OCIDs change in some meaningful way, we will return early on a parse error above which will just continually be retried.

Good catch. I will change a malformed OCID to drop only that field and log it, so a bad value does not block the snapshot. Transport and JSON failures will still be retried on the refresh cadence, since a read that failed once can succeed later.

pkg/cloudprovider/oke/oke_test.go
OguzPastirmaci: Yes, here's an example from A100 nodes:

`{"ipv6": false, "planes": 0}`

OguzPastirmaci:
> Will sync on a call to better understand the changes. For now, there is a lot of encoded logic that I don't fully understand.

Sync'ed offline with @dkennetzoracle.

The provider now reads the fabric object from the /host/ response only, with no direct endpoint, no gate, no confirmed state, no wake, and no sysfs. I replied on each thread; the remaining ones will land in the next push.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: No nil dereference. The pointers exist only on the IMDS wire type. `fabric()` returns nil unless both fields are present, so the snapshot holds a complete value or nil. I keep them as boolean and integer attributes so `false` and `0` stay real values.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: Done in 496c18a. Each request has its own 5 second limit, and the snapshot is built from whichever endpoint answered. During startup the shape is stored while the poll keeps trying the host for the rest of the window. Publication starts after GetInstance returns, when the inventory begins scanning.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: Fair point. I will remove the special fabric pinning so every field merges the same way: a present value replaces the previous one, and an omitted value keeps the last known one.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: Agreed. The next push reads host and instance independently, so the shape is published even when the host read fails. The startup poll still waits for the host data.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: The provider requires both fields. A partial object is treated as absent and logged at verbosity 2 with the missing field names, so it is visible when debugging without a new code path. Both test environments I checked return the full object.

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: Done in 496c18a

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: Done in 496c18a

pkg/cloudprovider/oke/oke.go
OguzPastirmaci: Done in 496c18a

dkennetzoracle (approved):
Comments addressed and spoke offline for changes. Approved on my end!

Issue #296, opened 2026-08-25
prepareResourceClaim idempotence: reuse already-stored device config for a claim


ngcxy:
### Context

Currently, `prepareResourceClaim` regenerates configuration for the claim on every call - it invokes `getDeviceNetworkConfig` (and therefore `GetProfileConfig`) for the claim unconditionally, with no check for an existing entry in `podConfigStore` for the same claim. 
```go
func (np *NetworkDriver) prepareResourceClaim(ctx context.Context, claim *resourceapi.ResourceClaim) kubeletplugin.PrepareResult {
    ...
    mergedConf, err := np.getDeviceNetworkConfig(result.Device, claim.UID, userConf)
    ...

func (np *NetworkDriver) getDeviceNetworkConfig(device string, claimUID types.UID, userConf *apis.NetworkConfig) (*apis.NetworkConfig, error) {
    cloudConf, ok := np.netdb.GetDeviceConfig(device)
    ...
    if mergedConf.Profile != "" {
        profileConf, err := np.netdb.GetProfileConfig(device, claimUID, mergedConf)
        ...
```

This behavior will introduce resource leakage after #241, which implements a `LocalIPAM` to dynamically allocate addresses with a stateful `allocatedIPs` set:
- Re-invoking `getDeviceNetworkConfig` allocates a new address and overwrites the checkpoint entry, orphaning the previously allocated one.
- `LocalIPAM` reseeds its in-use set from the now-stale checkpoint, so the orphaned address can later be handed to a different pod.

Thanks to @Bowser1704 for pointing this out with the details in https://github.com/kubernetes-sigs/dranet/pull/241#issuecomment-5236639985 

### Implementation concern

As @gauravkghildiyal noted in https://github.com/kubernetes-sigs/dranet/pull/241#discussion_r3835653897, adding a short-circuit check at the top of `prepareResourceClaim` is unsafe:

`podConfigStore` is already written **early** (when `netconf.Profile != ""`) as an in-flight rollback marker, *before* setup completes.
- early store: writes an incomplete `deviceCfg`
```go
    // Store early to guarantee profile cleanup on subsequent failures within this loop.
    // If the preparation fails later, Kubelet will call UnprepareResourceClaims,
    // which will find this early config and release the allocated profile.
    if netconf.Profile != "" {
        if err := np.podConfigStore.SetDeviceConfig(podUID, result.Device, deviceCfg); err != nil {
        ...
```

[...]

PR #326, opened 2026-09-09
feat: Apply interface settings to IPVLAN subinterfaces

### LINKED ISSUE #297: Evaluate and enable (or reject) advanced interface fields for subinterfaces
#241 introduces subinterface (IPVLAN) support. In that design, [`addIPVlan`](https://github.com/kubernetes-sigs/dranet/pull/241/changes#diff-793bda334ae28da43dcfc9a20b07429d7483a698bbdcff477ed30305413b09ccR43) only applies the IPVLAN `mode`/`flag` and `addresses` from the interface configuration. To avoid silently dropping other config fields, a validation guard is added to reject the fields that are meaningless and not handled for subinterfaces. :

```go
// validateSubinterfaceOnlyConfig checks that a NetworkConfig does not contain
// network-specific fields that are meaningless for a subinterface (e.g. IPVLAN).
func validateSubinterfaceOnlyConfig(cfg *InterfaceConfig, fieldPath string) (allErrors []error) {
    if cfg.MTU != nil || cfg.HardwareAddr != nil ||
        cfg.GSOMaxSize != nil || cfg.GROMaxSize != nil ||
        cfg.GSOIPv4MaxSize != nil || cfg.GROIPv4MaxSize != nil ||
        cfg.ARPIgnore != nil || cfg.ARPAnnounce != nil ||
        cfg.Addressing == AddressingModeDHCP || (cfg.DHCP != nil && *cfg.DHCP) {
        allErrors = append(allErrors, fmt.Errorf("%s: mtu, hardwareAddr, gso/gro sizes, arpIgnore, arpAnnounce and dhcp are not yet supported for subinterface types (type: %s)", fieldPath, cfg.Type))
    }
    return allErrors
}
```

In the long term, we need to evaluate how each of these fields should be treated for the subinterface scenario.

**Reference**

The issue was raised by @OguzPastirmaci, full context in https://github.com/kubernetes-sigs/dranet/pull/241#discussion_r3836404021

OguzPastirmaci:
#### What type of PR is this?

/kind bug

IPVLAN subinterfaces currently ignore several fields from the interface configuration.

This PR:

- Sets the requested MTU when the IPVLAN child is created.
- Applies `gsoMaxSize`, `groMaxSize`, `gsoIPv4MaxSize`, and `groIPv4MaxSize` to the child.
- Applies `arpIgnore` and `arpAnnounce` after the child is renamed and before addresses are added.
- Rejects `hardwareAddr` because the child uses the parent MAC address.
- Rejects DHCP because the current subinterface path is unsupported and untested.
- Validates the final configuration after provider and profile data are merged.
- Deletes the child when creation or later interface configuration fails.
- Releases an allocated profile when final configuration validation fails.

Fixes #297

If profile release fails after validation, Prepare returns both errors. Retry-safe Prepare behavior remains tracked by #296.

User zero-value overrides for these settings depend on #316. This PR does not change configuration merge precedence.

```release-note
Apply configured MTU, GSO and GRO sizes, and ARP settings to IPVLAN subinterfaces. Reject hardware address and DHCP configuration for IPVLAN subinterfaces.
```


aojea:
looks like a bug to me, right :)

/assign @ngcxy 

Sunyee good PR to exercise now the reviewers and maintainer skills since you added the feature ;)

aojea:
> @aojea: GitHub didn't allow me to assign the following users: ngcxy.
> 
> Note that only [kubernetes-sigs members](https://github.com/orgs/kubernetes-sigs/people) with read permissions, repo collaborators and people who have commented on this issue/PR can be assigned. Additionally, issues/PRs can only have 10 assignees at the same time. For more information please see [the contributor guide](https://git.k8s.io/community/contributors/guide/first-contribution.md#issue-assignment-in-github)

@ngcxy please look for someone to sponsor you and see if you can become a member of the org so you can get assigned to reviews

OguzPastirmaci:
> looks like a bug to me, right :)
> 
> /assign @ngcxy
> 
> Sunyee good PR to exercise now the reviewers and maintainer skills since you added the feature ;)

Yeah, closer to a bug than a feature :)

/kind bug

ngcxy:
> @ngcxy please look for someone to sponsor you and see if you can become a member of the org so you can get assigned to reviews

Just sent the request, thanks for reminding!

ngcxy:
Thanks for taking this on, @OguzPastirmaci! This change lets IPVLAN subinterfaces accept the link settings for MTU, GSO/GRO sizes, and ARP ignore/announce. Reading through it, it looks great to me — everything lines up with how the feature is meant to work.

I especially like the defensive cleanup around the subinterface lifecycle — in `subinterfaces.go`, the deferred rollback in `addIPVlan` tears down the child interface if a later setup step fails; and in `nri_hooks.go`, the half-created subinterface is deleted when `configureNetdevInNS` fails. Together they make sure a partial failure doesn't leave orphaned links behind.

Just got my org membership — let me try adding myself as a reviewer:

/assign @ngcxy
/lgtm

drkennetz:
@ngcxy open a PR to add yourself to the OWNERS file

OguzPastirmaci:
> /lgtm

@aojea @gauravkghildiyal when you have a moment, could you take a look at this one for approval?

We have a couple of follow-ups (the acceptRA setting and the OKE IPvlan ipv4 profile) that build on this, so I would love to get them lined up once this lands. Happy to address any further comments. Thanks!

gauravkghildiyal:
Sorry haven't had a chance. But will do at the earliest. Expect no longer than today/tomorrow.

pkg/driver/nri_hooks.go
gauravkghildiyal, line 316: Pre-existing issue but just noticed this as we are now deleting the subinterface if configureNetdevInNS fails. We should only set the NetworkDeviceReady=True condition and related info only when configureNetdevInNS also succeeds right?

pkg/driver/subinterfaces_test.go
```diff
@@ -17,59 +17,81 @@ limitations under the License.
...
+	origns netns.NsHandle
+	testNS netns.NsHandle
+	nsPath string
+	parent string
+}
+
+func newIPVlanTestEnv(t *testing.T, parentMTU int) *ipvlanTestEnv {
+	t.Helper()
 	if os.Getuid() != 0 {
 		t.Skip("Test requires root privileges.")
```
gauravkghildiyal, line 59: Non blocking: This is a pre-existing gap and my bad for failing to recommend this before. We don't need to fix it in this PR but maybe we can avoid needing root if we use: https://github.com/kubernetes-sigs/dranet/blob/8ba85701e5098a168f09472cabbcb7e18e139177/internal/testutils/userns.go#L61-L67

@ngcxy FYI maybe something that needs to be improved separately.

gauravkghildiyal (commented):
Just one request but looking good.

(One non blocking comment that we can pick separately)

pkg/driver/subinterfaces_test.go
OguzPastirmaci: Agreed, and thanks for the pointer. This was an issue in my Linux instance I use for testing as well with Ubuntu 24 because of AppArmor restrictions. I was running tests as root to bypass it.

I can send a follow-up PR with the conversion and a stronger support check. @ngcxy  happy to hand it over if you would rather take it.

pkg/driver/nri_hooks.go
OguzPastirmaci: Good catch, thanks. Fixed in 3d87a9b: the Ready condition and the network data are now set only after configureNetdevInNS succeeds, so a failed configuration leaves the status untouched.

For context, runPodSandbox returns on this error before it applies the status, so nothing false was published before. This change also keeps the status object consistent when configuration fails.

Btw, attachNetdevToNS has the same ordering for passthrough. Happy to change it here as well or in a follow-up, whichever you prefer.

pkg/driver/subinterfaces_test.go
ngcxy: Thanks for the catch! I'd be happy to take this on. I'll open a follow-up PR for the conversion :)

pkg/driver/nri_hooks.go
gauravkghildiyal: That's okay we can do it in a followup. Thanks!

pkg/driver/subinterfaces_test.go
gauravkghildiyal: Thanks both!

Issue #322, opened 2026-09-07
[flake] bats e2e test


aojea:
See in https://github.com/kubernetes-sigs/dranet/actions/runs/34124942870/job/101751319092?pr=321

```
not ok 17 unprepare claims on driver restart after pod deletion during driver downtime
# (in test file tests/e2e.bats, line 592)
#   `kubectl wait --namespace=kube-system --for=condition=Ready pod -l app=dranet --field-selector spec.nodeName="$NODE_NAME" --timeout=60s' failed
# deviceclass.resource.k8s.io/dra.net created
# resourceclaim.resource.k8s.io/dummy-interface-static-ip created
# pod/pod1 created
# pod/pod1 condition met
# pod/pod1 condition met
# node/dranet-test-cluster-worker not labeled
# daemonset.apps/dranet patched
# Waiting for daemon set "dranet" rollout to finish: 0 out of 2 new pods have been updated...
# Waiting for daemon set "dranet" rollout to finish: 0 out of 2 new pods have been updated...
# Waiting for daemon set "dranet" rollout to finish: 1 out of 2 new pods have been updated...
# Waiting for daemon set "dranet" rollout to finish: 1 out of 2 new pods have been updated...
# Waiting for daemon set "dranet" rollout to finish: 1 of 2 updated pods are available...
# daemon set "dranet" successfully rolled out
# Warning: Immediate deletion does not wait for confirmation that the running resource has been terminated. The resource may continue to run on the cluster indefinitely.
# pod "pod1" force deleted from default namespace
# daemonset.apps/dranet patched (no change)
# node/dranet-test-cluster-worker unlabeled
# error: no matching resources found
# --- Test failed. Dumping debug information ---
# --- DeviceClasses ---
# --- deviceclass.resource.k8s.io/dra.net ---
# apiVersion: resource.k8s.io/v1
# kind: DeviceClass
# metadata:
#   annotations:
#     kubectl.kubernetes.io/last-applied-configuration: |
#       {"apiVersion":"resource.k8s.io/v1","kind":"DeviceClass","metadata":{"annotations":{},"name":"dra.net"},"spec":{"selectors":[{"cel":{"expression":"device.driver == \"dra.net\""}}]}}
#   creationTimestamp: "2026-09-07T13:11:50Z"
#   generation: 1
#   name: dra.net
#   resourceVersion: "2012"
#   uid: f083283b-e339-4b63-b7dd-20287cc5dd13
# spec:
#   selectors:
[...]

PR #328, opened 2026-09-10
tests: fix flaky e2e tests caused by racy 'kubectl wait'

ngcxy:
#### What type of PR is this?

/kind flake

The bats-based e2e and webhook tests intermittently fail with "error: no matching resources found". The root cause is that `kubectl wait --for=condition=ready` errors out immediately when no resource currently matches its selector; it does not wait for pods that do not exist yet.

This PR:

- Add `--for=create` to the racy waits, so kubectl polls until the pod exists and then waits for readiness.
- In e2e.bats, restore the DaemonSet with `affinity:` (YAML null) so the merge patch actually removes the `nodeAffinity`.

Related to #322

```release-note
NONE
```

### DIFF
--- a/tests/e2e.bats
+++ b/tests/e2e.bats
@@ -585,11 +585,11 @@ EOF
 spec:
   template:
     spec:
-      affinity: {}
+      affinity:
 EOF
 )
   kubectl label node "$NODE_NAME" e2e-test-do-not-schedule-
-  kubectl wait --namespace=kube-system --for=condition=Ready pod -l app=dranet --field-selector spec.nodeName="$NODE_NAME" --timeout=60s
+  kubectl wait --namespace=kube-system --for=create --for=condition=Ready pod -l app=dranet --field-selector spec.nodeName="$NODE_NAME" --timeout=60s
 
   # The driver should asynchronously process the cleanup after the restart.
   # Inspect the bbolt database again to verify the pod configs have been deleted.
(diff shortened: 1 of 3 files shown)


OguzPastirmaci (commented):
Thanks, the fix looks good to me. Two things:

**1. Suite setup has the same race.** [`tests/setup_suite.bash:46`](https://github.com/kubernetes-sigs/dranet/blob/ce29f8eefd32f2fd6bc894e69970984f2f20ebc5/tests/setup_suite.bash#L46) waits on the Dranet pods right after applying the DaemonSet, under `set -eu`, so a hit there aborts the whole suite. Since the DaemonSet is freshly created there, `rollout status` could be a better fit than a pod wait:

```bash
kubectl rollout status daemonset/dranet -n kube-system --timeout=120s
```

It waits for the controller to observe the new generation and for all desired pods to become available, so it cannot return early on the initial zero counts. `--timeout` is needed because `rollout status` defaults to never timing out. Could you add that to this PR? [`hack/ci/e2e.sh:213`](https://github.com/kubernetes-sigs/dranet/blob/ce29f8eefd32f2fd6bc894e69970984f2f20ebc5/hack/ci/e2e.sh#L213) has the same pattern, optional.

**2. Optional:** the restore patch now really changes the pod template, which is the same situation as the previous test, and that one waits with `rollout status` after its patch. Doing the same here, in place of the `kubectl wait` on the worker pod, waits for all desired Pods and does not need the double `--for`:

```bash
kubectl rollout status ds/dranet --namespace=kube-system --timeout=120s
```

The patch increments the DaemonSet generation, so the wait cannot succeed using status from before the restore.

anson627:
The creation wait addresses the empty-selector race, and setting affinity to null correctly removes it during restoration

/lgtm

anson627 (approved):

aojea:
/ok-to-test
/lgtm
/approve

TIL

PR #329, opened 2026-09-11
driver: Handle missing namespace paths during subinterface cleanup

OguzPastirmaci:
#### What type of PR is this?

/kind bug

cri-o removes the pod network namespace before the NRI `StopPodSandbox` hook runs, so `nsDeleteSubinterface` fails to open the path and the hook logs `Failed to delete subinterface ... no such file or directory`. The kernel deletes any IPvlan child still in the namespace when it destroys the namespace, so there is nothing left to do.

Treat a missing namespace path like a link that is already gone: return nil and log at verbosity 2. Cleanup is unchanged when the path still exists, including the normal containerd sequence. The pod shutdown result does not change.

#137 

```release-note
Dranet no longer logs "Failed to delete subinterface" when the pod network namespace path is already missing.
```

### DIFF
--- a/pkg/driver/subinterfaces.go
+++ b/pkg/driver/subinterfaces.go
@@ -19,8 +19,10 @@ package driver
 import (
 	"crypto/rand"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"net"
+	"os"
 
 	"github.com/vishvananda/netlink"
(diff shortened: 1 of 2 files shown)


gauravkghildiyal:
Thanks @OguzPastirmaci 

/lgtm
/approve

PR #330, opened 2026-09-12
fix: open only a NETLINK_ROUTE socket for default nlwrap handles

yeonghoo2:
/kind bug

`nlwrap.NewHandle()` and `NewHandleAt()` are called without a family everywhere
in the driver. With no family, `netlink.NewHandle` opens ROUTE, XFRM and
NETFILTER sockets up front and fails if any of them cannot be opened. On a node
where `xfrm_user` or `nfnetlink` cannot be loaded (kernel built without them,
module loading disabled, `/lib/modules` not matching the running kernel) every
handle creation fails and `NodePrepareResources` fails for the whole node:

```
error creating netlink handle protocol not supported
```

DRANET only uses NETLINK_ROUTE through these handles; the RDMA path already
passes `NETLINK_RDMA` explicitly. This defaults nlwrap handles to
NETLINK_ROUTE so the two unused families are not opened. Callers that pass a
family are unchanged.

`nlwrap` also wraps `ConntrackDeleteFilters`, which would need NETFILTER, but
nothing in the driver calls it.

N/A

We hit this on VM guests whose image carried modules from a different kernel
build, so `xfrm_user` could not be loaded and no Pod could be prepared on those
nodes.

```release-note
Fix `NodePrepareResources` failing with "protocol not supported" on nodes where the xfrm_user or nfnetlink kernel modules cannot be loaded.
```


internal/nlwrap/netlink.go
```diff
@@ -35,16 +36,26 @@ type Handle struct {
...
+func defaultFamilies(nlFamilies []int) []int {
+	if len(nlFamilies) == 0 {
+		return []int{unix.NETLINK_ROUTE}
+	}
+	return nlFamilies
+}
+
 func NewHandle(nlFamilies ...int) (Handle, error) {
-	nlh, err := netlink.NewHandle(nlFamilies...)
+	nlh, err := netlink.NewHandle(defaultFamilies(nlFamilies)...)
```
aojea, line 52: so upstream magically defaults to the 3 supported families, we magically default to the one family we use more. 
I do not like magical defaults, but we are building on the same behavior, just update the function with a comment explaining that `NewHandle()` defaults to `NETLINK_ROOTE` if no family is provided

internal/nlwrap/netlink.go
```diff
@@ -35,16 +36,26 @@ type Handle struct {
 	*netlink.Handle
 }
 
+// defaultFamilies narrows the netlink default (ROUTE, XFRM, NETFILTER) to
+// NETLINK_ROUTE, the only family used through these handles. Opening the
```
aojea: we also use `unix.NETLINK_RDMA` ... the comment is not accurate

aojea:
@yeonghoo2 rebase and address the comments,  the `func NewHandle` is the one that needs to carry the comment explaining the new behavior, also add something to ensure that if you want to use a different family ensure to pass it as a parameter ,

internal/nlwrap/netlink.go
yeonghoo2: Done, moved it to the `NewHandle` doc comment.

internal/nlwrap/netlink.go
yeonghoo2: Thanks for catching that, you're right. Fixed: the comment now says other families such as `NETLINK_RDMA` must be passed explicitly.

yeonghoo2:
Rebased on main and moved the comment to `NewHandle`, including the note that other families must be passed as a parameter.

aojea:
/lgtm
/approve

seems reasonable , we just use our own default

PR #331, opened 2026-09-12
test: populate NetworkInterfaceStateInPod in the checkpoint fixture

yeonghoo2:
/kind failing-test

`main` fails `TestDeviceConfigFixturePopulated` since #304 and #314 merged in the same tide batch: #304 added `DeviceConfig.NetworkInterfaceStateInPod`, #314 added the test that requires every `DeviceConfig` field to be set in `fullDeviceConfig()`, and neither PR's CI saw the other.

This sets the field in the fixture, adds it to the golden wire format, and adds `NetworkInterfaceState` and `DHCPLeaseRecord` to the populated-field check. The field is additive (`omitempty`), so no `checkpointSchemaVersion` bump per DEVELOPER.md.

Follow-up to #304 and #314.

```release-note
NONE
```

### DIFF
--- a/pkg/driver/pod_device_config_bolt_version_test.go
+++ b/pkg/driver/pod_device_config_bolt_version_test.go
@@ -95,6 +95,13 @@ func fullDeviceConfig() DeviceConfig {
 				PrivateFlags: map[string]bool{"my-flag": false},
 			},
 		},
+		NetworkInterfaceStateInPod: &NetworkInterfaceState{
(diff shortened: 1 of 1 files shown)


yeonghoo2:
@aojea #304 and #314 landed in the same tide batch and the new fixture check did not see the field from #304; this fills it in. Also unblocks #330.

aojea:
/ok-to-test
/lgtm
/approve

Thanks for the quick analysis, I just found it a few minutes ago

PR #321, opened 2026-09-07
Azure: publish per-device IPv4 subnet attribute

nitin-nizhawan:
## What type of PR is this?
/kind feature

## What this PR does
Publishes each Azure network interface's IPv4 subnet CIDR as the device-specific `azure.dra.net/ipv4Subnet` ResourceSlice attribute.

The Azure IMDS interface is matched to the DraNet device by normalized MAC address. The attribute is omitted when the device MAC, matching interface, or complete IPv4 subnet metadata is unavailable. Existing node-level Azure attributes are unchanged.

An IPv4-specific attribute name is used because Azure dual-stack NICs can also have an IPv6 subnet, while Azure IMDS currently exposes explicit subnet address and prefix metadata under the IPv4 interface data.

## Testing
- `go test -race -count 1 ./pkg/cloudprovider/azure`
- `go test ./pkg/inventory ./pkg/driver`

```release-note
Expose the Azure IPv4 subnet CIDR for each network device as the `azure.dra.net/ipv4Subnet` ResourceSlice attribute.
```

PR #335, opened 2026-09-16
Azure: publish per-device IPv4 subnet attribute

anson627:
#### What type of PR is this?

/kind feature

Publish each Azure device's IPv4 subnet CIDR as the `azure.dra.net/ipv4Subnet` ResourceSlice attribute. Match Azure IMDS interfaces by normalized MAC address and omit the attribute when the device MAC or complete matching subnet metadata is unavailable.

Related to #321.

This takes over the patch originally authored by Nitin Nizhawan in #321 (commit eb999222ab47fb8d2e1822fb7569c92383c41d7b). The patch is unchanged and applied on current upstream main as a single commit authored by Anson Qian.

Validation: Linux/arm64 compilation of the Azure package tests passed (`GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go test -c ./pkg/cloudprovider/azure`). Running `go test -race -count=1 ./pkg/cloudprovider/azure` on macOS was blocked by existing Linux-only dependencies. `git diff --check` passed. Tests cover MAC normalization, matching-interface selection, missing or unmatched MAC addresses, and incomplete subnet metadata.

```release-note
Azure devices now expose their IPv4 subnet CIDR through the azure.dra.net/ipv4Subnet ResourceSlice attribute when matching IMDS subnet metadata is available.
```


aojea:
I'm unclear on the rules here, but the path author has now a CLA so I think is legit

tamilmani1989 (approved):

gauravkghildiyal:
(This has LGTM from @tamilmani1989 hence approving)

/lgtm
/approve

PR #336, opened 2026-09-17
inventory: use interface name for provider identifiers

Bowser1704:
#### What type of PR is this?

/kind bug
/kind api-change

`GetDeviceAttributes` already passes the interface name in `DeviceIdentifiers.Name`, following #279. Configuration and profile calls still pass the DRA device name, which may be PCI-derived or normalized and cannot reliably identify a local network interface.

Use a shared helper to pass the actual interface name consistently, leaving it empty when unavailable. DRA device names remain unchanged. `IsLACPBond` returns false without logging an error for an unavailable interface.

Related to #279.

External webhooks relying on the previous `name` value may require updates.

```release-note
Provider API behavior change: GetDeviceConfig, GetProfileConfig, and ReleaseProfileConfig receive the Linux interface name in DeviceIdentifiers.Name, or an empty string when unavailable, instead of the DRA device name. Action required for external webhooks that rely on the previous name value: update them to handle interface names and empty values.
```


gauravkghildiyal:
Thanks @Bowser1704 

I think the gap may have been created by https://github.com/kubernetes-sigs/dranet/pull/279 which switched DeviceIdentifiers.Name from the normalized PCI address (Or normalized interface name) to the interface name when fetching the device attributes but may have missed changing updateDeviceStore(). 

The rationale in that PR makes sense that we really don't need DeviceIdentifiers.Name to be either the normalized PCI address or the normalized interface name. So instead, it should rather be defined as the interface name itself. So would simply making sure `updateDeviceStore()` correctly assigns the value of the DeviceIdentifiers.Name field similar to `getProviderAttributes()` be sufficient?

Bowser1704:
@gauravkghildiyal nice catch.

Bowser1704:
Thanks @gauravkghildiyal, agreed. #279 already fixed `GetDeviceAttributes`; this change aligns the other three methods:

| Method | Previous `Name` | Updated `Name` |
|---|---|---|
| `GetDeviceAttributes` | Interface name, or empty | Unchanged |
| `GetDeviceConfig` | DRA device name | Interface name, or empty |
| `GetProfileConfig` | DRA device name | Interface name, or empty |
| `ReleaseProfileConfig` | DRA device name | Interface name, or empty |

The DRA device name can be PCI-derived or a normalized interface name. I checked how the affected methods use the identifiers:

| Provider | `GetDeviceConfig` | `GetProfileConfig` | `ReleaseProfileConfig` |
|---|---|---|---|
| Alibaba | Uses `Name` for local interface lookups | Uses `Name` for IPv6 configuration and allocation/routing settings | Does not use identifiers |
| Azure | Uses MAC for metadata and `Name` for the IPv6 gateway lookup | Not implemented | Not implemented |
| GCE | Does not use identifiers | Uses MAC, not `Name` | Does not use identifiers |
| AWS / OKE / CoreWeave | Returns nil; does not use identifiers | Not implemented | Not implemented |
| Webhook | Forwards identifiers | Forwards identifiers | Forwards identifiers |

In-tree consumers of `Name` expect an interface name. External webhooks relying on the previous DRA name may require updates; this is documented in the release note. I've adopted the shared helper approach, without adding a field or falling back to the DRA device name.

anson627:
Could you extend db_test.go to verify that GetDeviceConfig, GetProfileConfig, and ReleaseProfileConfig receive the interface name rather than the DRA device name? Covering a missing ifName and release after the device leaves inventory would also help protect the new behavior

Bowser1704:
@anson627 Thanks for the suggestion! Tests added.

PR #339, opened 2026-09-18
fix: make PrepareResourceClaim idempotent per device

### LINKED ISSUE #296: prepareResourceClaim idempotence: reuse already-stored device config for a claim

ngcxy:
#### What type of PR is this?

/kind bug

Previously, `prepareResourceClaim` unconditionally re-resolved device configurations on every call. Following the introduction of `LocalIPAM` (#241), retrying `PrepareResourceClaim` allocated new IP addresses on each retry, orphaning previously allocated IPs and risking address conflicts.

Additionally, the driver previously wrote an incomplete "early store" entry to `podConfigStore` to facilitate rollback via `UnprepareResourceClaims`. However, Kubelet retries `PrepareResourceClaim` directly without calling `UnprepareResourceClaims` on failure, and storing partial configurations under the same key prevented safe short-circuiting.

**What this PR does**
  - **`prepareDevice` & Idempotency**: Extracted per-device logic into `prepareDevice` and added an early `podConfigStore` check to reuse existing mature configurations directly.
  - **Deferred Rollback**: Removed the incomplete early store. Introduced a `deviceCommitted` flag with a `defer` hook to immediately release allocated profile if preparation fails before persistence, without relying on `UnprepareResourceClaims`.
  - **Tests**: Added coverage for reusing stored device configurations and verified profile rollback on partial failures.

Fixes #296 

- **DHCP & AddrList failure handling**: Previously, failures in `getDHCP` or `nlHandle.AddrList` appended to `errorList` without breaking out of the loop, which caused the driver to fall through and persist an incomplete `deviceCfg` (missing IPs/routes) into `podConfigStore`. In the refactored `prepareDevice`, these failures now return immediately, preventing incomplete configurations from persisting and ensuring any dynamic profile resources are cleanly rolled back.

```release-note
NONE
```


OguzPastirmaci:
Nice work on the profile release test, that's a clear improvement over the early store check.

One small suggestion: The idempotency case seeds the store directly, so it doesn't quite exercise the retry path from #296. A two-device claim where device B fails on the first call, followed by a second `prepareResourceClaim` asserting `GetProfileConfig` ran once for A and twice for B, would cover the real scenario end to end.

ngcxy:
> One small suggestion: The idempotency case seeds the store directly, so it doesn't quite exercise the retry path from #296. A two-device claim where device B fails on the first call, followed by a second `prepareResourceClaim` asserting `GetProfileConfig` ran once for A and twice for B, would cover the real scenario end to end.

Thank you for catching this. Agree - a test calling prepareResourceClaim twice and counting the GetProfileConfig call validates both the short-circuit and retry path. I have created the corresponding new test in `TestDynamicProfiles` in substitution for the old idempotency test case.

pkg/driver/dra_hooks.go
```diff
@@ -205,298 +203,317 @@ func (np *NetworkDriver) prepareResourceClaim(ctx context.Context, claim *resour
...
-					len(config.Requests) > 0 && !slices.Contains(config.Requests, requestName) {
-					continue
-				}
-				if errs := apis.ValidateRDMAOnlyConfig(&config.Opaque.Parameters); len(errs) > 0 {
-					errorList = append(errorList, errs...)
+	// deviceCommitted tracks whether the device config reached maturity and was persisted
+	// to the store. If the preparation fails in the middle, the deferred function will
+	// release the profile config to avoid leakage.
+	deviceCommitted := false
+	if netconf.Profile != "" {
```
gauravkghildiyal, line 275: nit: Is it correct that the real intent of the check `netconf.Profile != ""` is to gate invocation of `ReleaseProfileConfig()`. So we could still have a differ but within it do:

`if !deviceCommitted && netconf.Profile != ""`

(Functionality they are the same right now but it's useful for future clarity if we add more things into the differ which should not be gated by `netconf.Profile != ""`)

gauravkghildiyal (commented):
Thanks @ngcxy, looking good, just one small request.

pkg/driver/dra_hooks.go
ngcxy: Good point - this makes the defer function more general and extensible in the future. Moved the `netconf.Profile` into `defer()`. Thanks!

Issue #289, opened 2026-08-15
Add a provider neutral --cloud-provider-options flag


OguzPastirmaci:
Some provider settings cannot be auto detected. Example: #241 will add IPvlan subinterface support. On OKE, a planned follow-up calculates IPv4 child addresses for IPvlan interfaces on RDMA fabrics. The addresses come from a reserved range. That range must move when it conflicts with the cluster network plan, for example pods or services CIDR. A new CLI flag for each vendor setting is not practical.

This also matches the direction in #63. Providers are expected "to appropriately influence the configuration of the IPVLAN (things like which IP address to pick)" ([comment](https://github.com/kubernetes-sigs/dranet/issues/63#issuecomment-4513333652)). This flag is the configuration channel for that.

Proposal: one flag for provider-specific configuration.

```text
--cloud-provider-options=<provider>.<option>=<value>[,...]
```

Each key has a provider namespace, for example `oke.child-ipv4-cidr`. The binary parses the pairs into a map. Discovery passes the map to the active provider. Each provider defines and validates its own keys. The Helm chart gets a matching `cloudProviderOptions` value. The format follows `--feature-gates`.

Design:

- Options require an explicit `--cloud-provider-hint`. The namespace must match the hint. These checks read only flags, so a slow metadata service cannot cause a restart loop.
- Invalid pairs, unknown namespaces, and duplicate keys fail at parse time.
- An invalid option stops the driver. A transient metadata error keeps the current log-and-continue behavior.
- No keys exist yet. Any supplied option is rejected at startup with a clear error. This prevents silently ignored configuration. The first provider key arrives as its own PR.

I have a working implementation ready. If this approach sounds good, happy to open a PR.

Open question: how much of the mechanism should the first PR carry?

- Minimal: the flag, parsing, and the static checks. Any supplied option fails at startup. Discovery does not change. The accepted namespace list contains only providers that define options. The first provider key adds the discovery changes together with the code that uses them.
- Complete: discovery gets an options parameter and an error type that separates configuration failures from transient metadata failures. All provider namespaces parse. The first provider key then becomes a small change inside one provider.

The minimal shape carries no unused code. The complete shape puts the discovery review in this PR, so later provider PRs stay small. Feedback on which shape you prefer is welcome.

The mechanism does not depend on #241. The first OKE key configures the IPvlan work, so I can also hold the PR until #241 merges.

PR #340, opened 2026-09-18
cloudprovider: add OKE IPv4 RoCE IPvlan profile

OguzPastirmaci:
#### What type of PR is this?

/kind feature

On OKE, Oracle Cloud Agent (OCA) configures RDMA NIC addresses and per-NIC host routes. Moving a NIC into a pod removes host networking state that Dranet does not restore on return.

This PR adds the `oke-rdma` profile for recognized Ethernet RDMA NICs on IPv4 fabrics. The profile creates IPvlan children and keeps the physical NICs on the host.

The profile provides:

- A deterministic child address in `10.222.0.0/15`, derived from the primary VNIC and RDMA NIC index.
- An on-link route in table `100 + RDMA NIC index` and a source rule.
- The parent NIC's ARP settings, unless the claim overrides them.

N/A

The index comes from the `rdmaN` interface name. On nodes without those names, the provider derives the index from the OCA IPv4 address. Optional OCA configuration files supply the parent network and address mode.

The profile rejects explicit Passthrough, DHCP, unnumbered addressing, and claim-supplied addresses. Claims that supply routes, rules, or a VRF retain control of routing.

The change also adds provider tests and OKE documentation.

```release-note
Added the oke-rdma profile for OKE IPv4 RoCE NICs, providing IPvlan children with deterministic addresses and per-NIC routing.
```


Bowser1704:
@OguzPastirmaci Just curious about two things in the address derivation:

1. The child address is derived purely from `(hostPosition, nicIndex)`, so each RDMA NIC can only ever resolve to one IPvlan address. Does that mean this profile can't be combined with the sharing work in #63?

2. The child range is a hardcoded `10.222.0.0/15` while the parent range is overridable via `rdma_network.json`. Since the stride is the eth0 subnet size, a node subnet of /15 or larger leaves room for `rdma0` only (or nothing at all, past /15) — is that combination just not expected on OKE, or is failing there the intended behaviour?

OguzPastirmaci:
> @OguzPastirmaci Just curious about two things in the address derivation:
> 
> 1. The child address is derived purely from `(hostPosition, nicIndex)`, so each RDMA NIC can only ever resolve to one IPvlan address. Does that mean this profile can't be combined with the sharing work in [Support RDMA NIC sharing #63](https://github.com/kubernetes-sigs/dranet/issues/63)?
> 2. The child range is a hardcoded `10.222.0.0/15` while the parent range is overridable via `rdma_network.json`. Since the stride is the eth0 subnet size, a node subnet of /15 or larger leaves room for `rdma0` only (or nothing at all, past /15) — is that combination just not expected on OKE, or is failing there the intended behaviour?

@Bowser1704 thanks, good questions.

1. Yes, by design for this version, one pod per RDMA NIC. It does not block #63. Sharing needs a slot term in the address and a larger child range, with the slot reserved through `LocalIPAM` like Alibaba and GCE (assuming this is still the way you're planning to do it in Alibaba). The profile hooks for that exist and are unused here. We're working on a plan for NIC sharing, with the aim of following up once the #63 work lands.

2. A /15 or larger subnet cannot occur on OCI, VCN blocks are /16 to /30. The RDMA network is not part of the VCN, it's a separate network not routable inside the VCN. The real limit is a worker subnet wider than /19: the /15 child range fits 16 RDMA NICs on a /19, 8 on a /18, 4 on a /17, 2 on a /16. Above that, the claim for that NIC fails with a clear error (intended) and the other NICs keep working. The docs list this. Your question also led me to a bug on that path, fixed in the latest push. A cluster can also avoid the derived range today: with `--profile-provider=webhook`, a webhook (for example one backed by NVIDIA IPAM) resolves `oke-rdma` and supplies the address. I'd love to make the child range configurable next. It needs the provider options flag from #289, and I plan to open a PR for it. Would that flag be useful for the Alibaba provider too?

gauravkghildiyal:
@OguzPastirmaci does this need an LGTM from @dkennetzoracle?

/approve

dkennetzoracle:
> @OguzPastirmaci does this need an LGTM from @dkennetzoracle?
> 
> /approve

Reviewing now :)

pkg/cloudprovider/oke/oke.go
```diff
@@ -60,8 +69,50 @@ const (
...
+	// Oracle Cloud Agent names the RDMA fabric interfaces rdmaN.
+	rdmaNicPrefix = "rdma"
+
+	// okeRDMAProfile is advertised for every Ethernet RDMA NIC and resolved
+	// by this provider into an IPvlan child with a deterministic address.
+	okeRDMAProfile = "oke-rdma"
+	// okeRDMASBRTableBase plus the RDMA NIC index is the routing table of a child.
+	// TableIDForName hashes into 1000 to 1999 and can collide between two
+	// RDMA NICs of one pod; the RDMA NIC index cannot.
+	okeRDMASBRTableBase = 100
```
dkennetzoracle, line 82: nit: Should we use: `okeRDMASBRTableBase = apis.RouteTableOffset` instead? I think this is collision free by construction. Any issues with that?

pkg/apis/constants.go:33 documents these as DRANET-managed, and structurally unable to hit a reserved table.

dkennetzoracle: Would hit: oke_test.go:188, :189 (table 100) and :508, :509 (table 115), plus the doc sample at site/content/docs/user/oci-oke-rdma.md:86-87.

OguzPastirmaci: > nit: Should we use: `okeRDMASBRTableBase = apis.RouteTableOffset` instead? I think this is collision free by construction. Any issues with that?
> 
> pkg/apis/constants.go:33 documents these as DRANET-managed, and structurally unable to hit a reserved table.

1000 to 1015 is inside the range that `TableIDForName` hashes into, which is the subject of #344. A named VRF on another device of the pod could land on an RDMA NIC table: a VRF named `storage` defaults to 1014, which would be the table of `rdma14`. 100 to 115 is outside every table that Dranet generates. We can definitely move to the Dranet range once #344 is fixed.

Good point on the reserved tables. I will add a check that the RDMA NIC index is below 16, so the tables stay inside 100 to 115.

dkennetzoracle:
@gauravkghildiyal 

/lgtm

PR #341, opened 2026-09-18
feat: add acceptRA interface setting

OguzPastirmaci:
#### What type of PR is this?

/kind feature

The kernel resets `accept_ra` to the destination namespace default when an interface moves into a Pod, the same way it resets `arp_ignore` and `arp_announce`. A workload that depends on IPv6 router advertisements has no way to request the setting through the claim.

This change:

- Adds `interface.acceptRA` with values 0 to 2.
- Writes it through the sysctl helper that already applies the ARP settings, before addresses and link-up, on passthrough interfaces and on IPVLAN subinterfaces. The helper is renamed from ARP to sysctl to match.
- Rejects claim-supplied `acceptRA` for RDMA-only devices, which have no network interface.
- Rejects `acceptRA` together with an explicit `mtu` below 1280, because the kernel creates IPv6 settings only for an interface with an MTU of 1280 or more. An interface that keeps a smaller MTU from the host fails at setup with a sysctl setup error.

N/A

If merged validation fails after profile allocation, the driver calls `ReleaseProfileConfig` through the cleanup logic from #326.

#316 and this PR are coupled but independent: the second one to merge adds `acceptRA` to the explicit merge of #316 (one `overridePtr` line in `mergeInterfaceConfig`) and to the field lists in its tests. The coverage tests of #316 fail until that line exists.

Until #316 merges, the merge on `main` treats a pointed-to zero as empty and writes user values through the aliased pointer into the provider's stored configuration. That affects every pointer field today. None of the built-in providers sets `acceptRA` yet. The OKE provider will use it for IPv6 fabrics in a follow-up PR. A custom webhook can return it now, and until #316 merges a user `acceptRA: 0` loses to such a value and a user value is written into the stored webhook output for later claims on the same device.

```release-note
Added `interface.acceptRA` to set IPv6 router advertisement acceptance on a Pod interface.
```


pkg/driver/interface_sysctl.go
```diff
@@ -30,35 +30,38 @@ import (
...
 	if interfaceConfig.ARPIgnore != nil {
-		set("arp_ignore", *interfaceConfig.ARPIgnore)
+		set("ipv4", "arp_ignore", *interfaceConfig.ARPIgnore)
 	}
 	if interfaceConfig.ARPAnnounce != nil {
-		set("arp_announce", *interfaceConfig.ARPAnnounce)
+		set("ipv4", "arp_announce", *interfaceConfig.ARPAnnounce)
+	}
+	if interfaceConfig.AcceptRA != nil {
+		set("ipv6", "accept_ra", *interfaceConfig.AcceptRA)
```
dkennetzoracle: I think here we also want to check that disable_ipv6 is not set (or this check should be somewhere) so if AcceptRA > 0, but disable_ipv6 = 1, we should set an error in the errorlist.

dkennetzoracle: Right now, accept_ra is written unconditionally, but on a node booted with ipv6.disable=1, there is no filepath for the proc, so you get errors. I think we could do something like this which would make it consistent with `applyInterfaceForwarding` helper which already treats `os.ErrNotExist` on the IPv6 path as "IPv6 disabled, skip". ENOENT is also what you get for MTU < 1280 (no inet6_dev):

```go
      if interfaceConfig.AcceptRA != nil {
              name := fmt.Sprintf("net/ipv6/conf/%s/accept_ra", ifName)
              err := sysctlInterface.SetSysctl(name, int(*interfaceConfig.AcceptRA))
              switch {
              case err == nil:
              case errors.Is(err, os.ErrNotExist) && *interfaceConfig.AcceptRA == 0:
                      // No IPv6 on this interface (stack disabled or MTU below 1280), so
                      // nothing accepts router advertisements anyway.
                      klog.V(4).Infof("%s not found; IPv6 is not enabled on %s and acceptRA: 0 is already satisfied", name, ifName)
              case errors.Is(err, os.ErrNotExist):
                      errorList = append(errorList, fmt.Errorf("failed to set %s: IPv6 is not enabled on the interface (IPv6 disabled on the node or MTU below %d): %w", name, apis
              default:
                      errorList = append(errorList, fmt.Errorf("failed to set %s: %w", name, err))
              }
      }
```

pkg/driver/hostdevice.go
dkennetzoracle, line 36: I think we can fail fast in this function when mtu < 1280 because we know this ahead of time. Right now, we catch it after the the interface is moved into the pod.

dkennetzoracle, line 41: Something here like:
```go
      // The kernel creates no IPv6 settings below this MTU, so accept_ra cannot be
      // set. Reject it before the link is touched so the host device stays usable.
      if interfaceConfig.AcceptRA != nil && interfaceConfig.MTU == nil && hostDev.Attrs().MTU < apis.MinIPv6MTU {
              return nil, fmt.Errorf("acceptRA requires an MTU of at least %d, but %s has MTU %d and the claim sets no mtu", apis.MinIPv6MTU, hostIfName, hostDev.Attrs().MTU)
      }
```

dkennetzoracle:
@aojea - still reviewing, but I also noticed we may face some issues in merging that would be addressed by #316 or at least helped. How do we want to proceed there, since that is blocked by a doc. SLAAC will also likely run into this. I don't think it's a good use of time for us to fix code around the mergo behavior that should be replaced relatively soon. Thoughts?

pkg/driver/subinterfaces.go
dkennetzoracle, line 201: Probably similar here:
```go
if config.AcceptRA != nil && config.MTU == nil && parentLink.Attrs().MTU < apis.MinIPv6MTU {
              return nil, fmt.Errorf("acceptRA requires an MTU of at least %d, but parent interface %s has MTU %d and the claim sets no mtu", apis.MinIPv6MTU, hostIfName, parentLink.Attrs().MTU)
      }
```

pkg/apis/validation.go
```diff
@@ -402,7 +413,7 @@ func ValidateRDMAOnlyConfig(raw *runtime.RawExtension) []error {
 		config.Interface.GROMaxSize != nil || config.Interface.GSOIPv4MaxSize != nil ||
 		config.Interface.GROIPv4MaxSize != nil || config.Interface.DisableEBPFPrograms != nil ||
 		config.Interface.ARPIgnore != nil || config.Interface.ARPAnnounce != nil ||
-		config.Interface.IPVlan != nil {
+		config.Interface.AcceptRA != nil || config.Interface.IPVlan != nil {
```
dkennetzoracle: This list drifts every time a new field is added. Could we / should we do something like a `reflect.DeepEqual(config.Interface, InterfaceConfig{})` here?

dkennetzoracle (changes requested):
A few pending comments, and the major for me is the accept_ra written unconditionally. One more callout I noticed is that `applyInterfaceForwarding` is never called and duplicates the new sysctl helper added here. I think we could delete that one and give us one implementation to maintain:

```go
func hasInterfaceSysctlConfig(interfaceConfig apis.InterfaceConfig) bool {
      return interfaceConfig.ARPIgnore != nil || interfaceConfig.ARPAnnounce != nil ||
              interfaceConfig.AcceptRA != nil || interfaceConfig.Forwarding != nil
}

      if interfaceConfig.Forwarding != nil {
              v := int32(0)
              if *interfaceConfig.Forwarding {
                      v = 1
              }
              set("ipv4", "forwarding", v)
              set("ipv6", "forwarding", v)
      }
```
You'd need to ensure the set closure extends to tolerate os.ErrNotExist on the ipvj6 family for forwarding. Then add a `Forwarding` case to `interface_sysctl_test.go` and an e2e assertion on `/proc/sys/net/ipv4/conf/<if>/forwarding`

gauravkghildiyal:
Hey @dkennetzoracle and @OguzPastirmaci -- I'm trying to understand the relative priorities of the changes and I want to get you unblocked sooner. https://github.com/kubernetes-sigs/dranet/pull/316 is definitely something that needs a fix but would you say it's a strict blocker for some of your feature work like this PR?

OguzPastirmaci:
> Hey @dkennetzoracle and @OguzPastirmaci -- I'm trying to understand the relative priorities of the changes and I want to get you unblocked sooner. #316 is definitely something that needs a fix but would you say it's a strict blocker for some of your feature work like this PR?

@gauravkghildiyal thanks for checking. Not a strict blocker for this PR, the feature works without it, a claim with `acceptRA: 1` or `2` is applied as written.

That said, I think #316 will pay off soon. The next changes (SLAAC and the OKE IPv6 profile) lean on the merge behavior, and everything that lands before it is one more thing to port later. Since the merge path is shared by all providers, any provider returning pointer fields would benefit too.

dkennetzoracle:
> Hey @dkennetzoracle and @OguzPastirmaci -- I'm trying to understand the relative priorities of the changes and I want to get you unblocked sooner. #316 is definitely something that needs a fix but would you say it's a strict blocker for some of your feature work like this PR?

Thanks for checking in @gauravkghildiyal - I'll echo Oguz, it is not a strict blocker for the upcoming work but I do think we will get to a point soon-ish where it will become necessary to fix bugs / changes in the mergo merge code that are stopgaps until we get 316 in. So not blocked! Just something I think we should prioritize as a group as much as time allows (which we all know is limited 😄 )

pkg/driver/interface_sysctl.go
OguzPastirmaci: Agreed, will do. With the early MTU guards below, a low MTU is rejected before we get here, so the ENOENT branch should only cover a missing ipv6 sysctl for other reasons.

pkg/driver/hostdevice.go
OguzPastirmaci: Yep, will add it before `LinkSetDown`.

OguzPastirmaci: Taking this as written, including the rejection of `acceptRA: 0`. That matches the existing validator for an explicit `mtu`.

pkg/apis/validation.go
OguzPastirmaci: Good catch, it already has: `forwarding` and `vrf` are missing. One question on `DeepEqual`: it would also reject `{"interface":{"addresses":[]}}`, which the list accepts today. Do you see that as a problem, or is the stricter check fine? My first thought was to keep the current behavior and add a reflection based test that sets each `InterfaceConfig` field and expects the rejection, but I'm happy to go with `DeepEqual` if you'd rather have the simpler code.

dkennetzoracle: I don't have a straightforward answer for this. The following would preserve the behavior and seems a bit more maintainable. However, we'd need to do this check for every field which has an underlying slice.

The below code should work because `InterfaceConfig{}` initializes `Addresses` to nil, while decoding `"addresses":[]` creates a non-nil slice with length 0 so DeepEqual considers them different.

Creating an `iface` from the existing config as so would create a shallow copy, so each field is copied, but not slices, maps, and pointers which still reference the same underlying data. Setting `Addresses` to nil would only change the copied struct's slice field, but `config.Interface.Addresses` remains unchanged.

```go
iface := config.Interface
if Len(iface.Addresses) == 0 {
    iface.Addresses = nil
}

hasInterfaceConfig := !reflect.DeepEqual(iface, InterfaceConfig{})
...
```

IDK it just seems a bit simpler to me, but we'd want to make sure to add tests. We also don't have to change this, it is just preference.. Maybe wait for others to weigh in.

gauravkghildiyal:
Thanks for the details @OguzPastirmaci and @dkennetzoracle. I'll go through the doc this week and we can move it forward.

pkg/apis/validation.go
dkennetzoracle: Spoke with @OguzPastirmaci and we will address this logic in an isolated follow-up PR as a separate change

dkennetzoracle (approved):
/lgtm

